PR #3, opened 2017-10-16
set inventory() target to null/None to get all targets

ramaro:
This will simply allow to call inventory(target=null) and return a view of all target in the inventory as opposed to the default current target.


gburiola:
👍

PR #5, opened 2017-10-17
HTML anchors are case-sensitive

jhermann:


ramaro:
Well spotted @jhermann thank you!

jhermann:
Sorry, won't enter our internal CLA workflow for changing case of 3 letters.

ramaro:
Fair enough, we'll change it. Thanks again @jhermann

ademariag:
Thank you @jhermann I have made the change for you.

PR #9, opened 2017-10-22
Default search_path for the now packaged lib/kapitan.libjsonnet

ramaro:
Any jsonnet code importing lib/kapitan.libjsonnet will not need to copy that file the jsonnet/ directory anymore. 
search_imports() will now default to search for the packaged lib/kapitan.libjsonnet file instead.


gburiola:
👍

PR #10, opened 2017-10-31
Secrets with GnuPG support

ramaro:
This exposes a new **secrets** command to help write and reveal secrets with the goal of avoiding any sensitive data in templates or compiled files.

Every new secret requires a token path and recipients which will be written to a file structure generated by the --write argument. 
The content of a secret token file contains the GnuPG encrypted secret data and a list of recipient fingerprints.

Example of encrypting the content of files/secret_sauce into the app1/secret1 token:
```
$ echo "super secret data" > files/secret_sauce
$ kapitan secrets --write app1/secret1 --recipients human@example.com -f files/secret_sauce
```

Tokens can be used in templates and inventory and will not be revealed when compiled. Any compiled file with secrets will contain the secret token tags instead. 
A token tag is in the form of  ?{backend:path/to/secret}. To refer to the secret above, use ?{gpg:app1/secret1}

To reveal secret, the --reveal argument will read a file and resolve the token tags by reading the file structure and decrypting (should the one revealing be a recipient) to stdout only.
The reason for stdout here is to discourage saving revealed files into disk and instead pipe their output into i.e. ```kubectl apply -f```

Example of revealing content with the app1/secret1 token:
```
$ echo "this string has ?{gpg:app1/secret1}" | kapitan secrets --reveal -f -
this string has super secret data
```

This solution does not do any key management. That is a problem that the user will need to do by managing his GnuPG keys and settings. The assumption with this solution is that the keys used for recipients are trusted (otherwise python-gnupg will fail).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ notifications:
     on_success: change
     on_failure: always
 
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install -y gnupg2
+
 # command to install dependencies
 install:
   - pip install -r requirements.txt
--- a/examples/inventory/targets/minikube-es.yml
+++ b/examples/inventory/targets/minikube-es.yml
@@ -3,6 +3,10 @@ classes:
   - component.elasticsearch
 
 parameters:
+  kapitan:
+    secrets:
+      recipients:
+        - dummy@recipient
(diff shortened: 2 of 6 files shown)


ademariag:
LGTM once checks pass

PR #12, opened 2017-12-15
new yaml_dump() native jsonnet function

ramaro:
This native function allows rendering jsonnet object into YAML in runtime. Especially handy when you need to render a ConfigMap where the data is YAML.


gburiola:
lgtm

PR #18, opened 2018-01-25
Check for gpg key expiry

ramaro:
Since python-gnupg does not respect/read the default key in ~/.gnupg/gpg.conf, this checks for the key expiry and returns the next valid one, should the recipient be associated to multiple keys.


gburiola:
lgtm

PR #19, opened 2018-01-29
base64 encode and write secret support

ramaro:
Add a new --base64 flag that when used with

```
$ kapitan secrets --write --base64 -f file_to_encrypt.txt
```

will base64 its content and then encrypt. It will also add a new 'encoding' key in the secret file to note that the original content was base64'd.

Essentially, this is equivalent of doing:

```
$ cat file_to_encrypt.txt | base64 | kapitan secrets --write my_secret --recipients person@example.com -f -
```

which is especially handy when using Kubernetes Secrets (which content needs to be base64'd)


gburiola:
👍

PR #24, opened 2018-02-21
improve readme.md, update git and change folder of examples

gburiola:


examples/components/mysql/service.jsonnet
```diff
@@ -1,25 +0,0 @@
...
-local namespace = inventory.parameters.namespace;
-local instance_name = inventory.parameters.mysql.instance_name;
-
-
-{
-  apiVersion: "v1",
-  kind: "Service",
-  spec: {
-    ports: [
-      { name: "mysql", port: 3306, target_port: "mysql" },
```
gburiola: Seems like this file was deleted, but it was just moved to `examples/kubernetes` and this line was changed to:
`target_port` -> `targetPort`

PR #33, opened 2018-03-20
Python3

adrianchifor:
This is backwards compatible with Python 2, but compiled output will have different ordering because of `obj.iteritems()` (py2) -> `obj.items()` (py3).

- Support for Python 3 in the codebase:
    - `print` function
    - some functions updated to take encoded objects instead of plain strings (`.encode("UTF-8")`, `.decode("UTF-8")`)
    - `.iteritems()` -> `.items()`
    - `basestring` -> `string_types` from six
- Refactored the Dockerfile to use a python 3 alpine base directly.
- Refactored `setup.py` to take dependencies from `requirements.txt`.
- Updated reclass to be taken from a specific commit on python3 branch of [salt-formulas/reclass](https://github.com/salt-formulas/reclass/tree/31770c696f52d35a6164e03e4eb5010d3c40accc); needed for python 3 support.
- Updated python-gnupg to be taken from a specific commit of [vsajip/python-gnupg](https://github.com/vsajip/python-gnupg/tree/73b5d8d4ea509ed5f7da219837de784063aacd67); needed to fix an [issue with gnupg version checking on Macs](https://bitbucket.org/vinay.sajip/python-gnupg/issues/84/on-osx-version-detection-fails-then-raises).
- Updated README.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 python:
-  - "2.7"
+  - "3.6"
 
 branches:
   only:
@@ -12,6 +12,7 @@ notifications:
     - buriola@google.com
     - ramaro@google.com
     - ademaria@google.com
+    - chifor@google.com
   email:
     on_success: change
     on_failure: always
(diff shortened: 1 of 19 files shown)


ramaro:
lgtm

PR #39, opened 2018-03-22
Update tests

adrianchifor:
- File was not being closed `search_imports` function
- Added `get_directory_hash` function to compute sha256 hashes for dirs
- Added yaml filter test for jinja2
- Added Compilation functional test:
    - Goes into `examples/kubernetes` and runs the equivalent of `kapitan compile`
    - Compares hashes of `examples/kubernetes/compiled` and `tests/test_kubernetes_compiled`

### DIFF
--- a/kapitan/utils.py
+++ b/kapitan/utils.py
@@ -212,3 +212,29 @@ def searchvar(flat_var, inventory_path):
                 fd.close()
     for i in output:
         print('{0!s:{l}} {1!s}'.format(*i, l=maxlenght + 2))
+
+def get_directory_hash(directory):
+    '''
+    Compute a sha256 hash for the file contents of a directory
+    '''
+    if not os.path.exists(directory):
+        logger.error("utils.get_directory_hash failed, %s dir doesn't exist", directory)
+        return -1
+
+    try:
+        hash = sha256()
+        for root, _, files in os.walk(directory):
+            for names in files:
+                file_path = os.path.join(root, names)
+                try:
+                    with open(file_path, 'r') as f:
+                        hash.update(sha256(f.read().encode("UTF-8")).hexdigest().encode("UTF-8"))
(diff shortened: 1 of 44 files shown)


kapitan/utils.py
```diff
@@ -212,3 +212,29 @@ def deep_get(x, keys):
...
     for i in output:
         print('{0!s:{l}} {1!s}'.format(*i, l=maxlenght + 2))
+
+def get_directory_hash(directory):
+    '''
+    Compute a sha256 hash for the file contents of a directory
+    '''
+    if not os.path.exists(directory):
+        logger.error("utils.get_directory_hash failed, %s dir doesn't exist", directory)
+        return -1
```
ramaro: Can you raise KapitanError (or similar here https://github.com/deepmind/kapitan/blob/master/kapitan/errors.py) instead?

kapitan/utils.py
```diff
@@ -212,3 +212,29 @@ def deep_get(x, keys):
...
+        hash = sha256()
+        for root, _, files in os.walk(directory):
+            for names in files:
+                file_path = os.path.join(root, names)
+                try:
+                    with open(file_path, 'r') as f:
+                        hash.update(sha256(f.read().encode("UTF-8")).hexdigest().encode("UTF-8"))
+                except Exception as e:
+                    logger.error("utils.get_directory_hash failed to open %s: %s", file_path, str(e))
+                    return -2
```
ramaro: same

kapitan/utils.py
```diff
@@ -212,3 +212,29 @@ def deep_get(x, keys):
...
+
+    except Exception as e:
+        logger.error("utils.get_directory_hash failed: %s", str(e))
+        return -2
```
ramaro: same

PR #58, opened 2018-04-04
remove not needed command

ben-st:
updating the package cache is IMHO not needed since the 

> --no-cache

 option will pull directly from the online repos anyway. 
(with the small benefit of around 1MB for the package cache)

if you do this to avoid caching of the layers while building i suggest to build with the --no-cache option from docker:
`docker build --no-cache `

All the best

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.6-alpine
 
-RUN apk add --update --no-cache git g++ make libstdc++ gnupg musl-dev && \
+RUN apk add --no-cache git g++ make libstdc++ gnupg musl-dev && \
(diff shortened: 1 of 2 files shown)


adrianchifor:
`--no-cache` has nothing to do with `--update`. Adding `--update` is the equivalent of running `apk update`, which is recommended before installing packages.

ben-st:
update just updates the local package cache i suppose
screen shot 2018-04-04 at 15 59 18

adrianchifor:
Yes you're right, got confused for a second :) Thanks for the contribution!

ben-st:
you're welcome :)

PR #59, opened 2018-04-04
Gnu update

ben-st:
updated gnupg to 0.4.2 as suggested in requirements.txt 
released on 28.03.2018 
https://github.com/vsajip/python-gnupg

tested the build, screenshot below
screen shot 2018-04-04 at 15 25 33

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.6-alpine
 
-RUN apk add --update --no-cache git g++ make libstdc++ gnupg musl-dev && \
+RUN apk add --no-cache git g++ make libstdc++ gnupg musl-dev && \
     mkdir /kapitan
 
 WORKDIR /kapitan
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM python:3.6-alpine
 
-RUN apk add --update --no-cache git g++ make libstdc++ gnupg musl-dev && \
```
adrianchifor: `--updated` is required to run `apk update` before adding packages

ben-st: https://github.com/deepmind/kapitan/pull/58 
seems i messed up the commits this one should be in the prior PR sry :(

we won't need the --update since --no-cache is pulling the packages directly from the net.

tested this in the other PR.

adrianchifor: Great, thanks :)

ben-st: awesome 👍 
sry for the confusion :)

PR #60, opened 2018-04-04
"inventory --pattern" requires "--target_name" to be specified

gburiola:


gburiola:
```
$ ./kapitan inventory --pattern parameters.target_name
usage: kapitan [-h] [--version] {eval,compile,inventory,searchvar,secrets} ...
kapitan: error: --pattern requires --target_name
```

ramaro:
How about we make it filter all targets when -t is not passed instead?

adrianchifor:
Will do

ramaro:
Merging for now. We can improve later

PR #61, opened 2018-04-04
Tests runner helper

adrianchifor:


ramaro:
We can also patch .travis.yml to use this @adrianchifor

PR #65, opened 2018-04-10
Add openssl to Dockerfile

gburiola:


ramaro:
Shouldn't this be on the CI image instead?

PR #71, opened 2018-04-16
Cache inventory (reclass) across multiple threads

### LINKED ISSUE #25: Load inventory only once
Calling reclass and generating the inventory is an expensive process (~2 seconds). We do save the result using the `memoise` decorator, however we still run it once for each thread.

We should load the inventory once before we create multiple threads and pass the result to each thread.

adrianchifor:
Fixes #25 

- Save `inventory_reclass` output as a cached variable in `cached` module, initialized only once. No need to complicate ourselves with `multiprocessing.Manager` as inventory is read-only in threads and doesn't change.
- Saves ~8-10s on `kapitan compile`

### DIFF
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -48,15 +48,15 @@ def compile_targets(inventory_path, search_path, output_path, parallel, targets,
     multiprocessing pool with parallel number of processes.
     kwargs are passed to compile_target()
     """
-    target_path = "targets"
     # temp_path will hold compiled items
     temp_path = tempfile.mkdtemp(suffix='.kapitan')
+
+    target_objs = load_target_inventory(inventory_path, targets)
+
     pool = multiprocessing.Pool(parallel)
     # append "compiled" to output_path so we can safely overwrite it
     compile_path = os.path.join(output_path, "compiled")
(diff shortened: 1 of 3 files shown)


kapitan/targets.py
```diff
@@ -48,16 +48,16 @@ def compile_targets(inventory_path, search_path, output_path, parallel, targets,
     multiprocessing pool with parallel number of processes.
     kwargs are passed to compile_target()
     """
-    target_path = "targets"
     # temp_path will hold compiled items
     temp_path = tempfile.mkdtemp(suffix='.kapitan')
+
+    target_objs = load_target_inventory(inventory_path, targets)
```
adrianchifor: Initializing `target_objs` before `multiprocessing.Pool` fork is the magic sauce

kapitan/targets.py
```diff
@@ -48,16 +48,16 @@ def compile_targets(inventory_path, search_path, output_path, parallel, targets,
     multiprocessing pool with parallel number of processes.
     kwargs are passed to compile_target()
     """
-    target_path = "targets"
```
adrianchifor: Unused var

PR #93, opened 2018-05-10
README for the terraform example

mumoshu:
Hi! Thanks for maintaining an awesome project :+1:

I just realized that kapitan supports terraform, but wasn't quite sure how I could actually use it.
I even didn't know that `*.tf.json` files are consumable by Terraform without any option, that was the hardest part for me to catch 😆

Anyway, I wrote up this to help future kapitan+terraform users not to get confused like me.

### DIFF
--- a/examples/terraform/README.md
+++ b/examples/terraform/README.md
@@ -1,0 +1,42 @@
+# Using Kapitan to generate Terraform configuration
+
+kapitan can be used to generate `*.tf.json` with jsonnet and jinja2.
+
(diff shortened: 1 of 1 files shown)


ramaro:
Thank you! This is great and yes, documentation in the project is in need of some love.

gburiola:
Hey @mumoshu. Thanks a lot for doing this.

Yes, terraform consuming `*.tf.json` isn't very well documented but it works natively.

If you're converting a lot of HCL code/examples to json / jsonnet you like this site: https://www.hcl2json.com/

PR #101, opened 2018-05-24
Update kube libsonnet

adrianchifor:
- Updated kube.libjsonnet in example
- Recompiled
- Fixed tests
- Some typos

### DIFF
--- a/examples/kubernetes/compiled/minikube-mysql/manifests/mysql_secret.yml
+++ b/examples/kubernetes/compiled/minikube-mysql/manifests/mysql_secret.yml
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  MYSQL_ROOT_PASSWORD: P3tncGc6bXlzcWwvcm9vdC9wYXNzd29yZH0=
+  MYSQL_ROOT_PASSWORD: ?{gpg:mysql/root/password:20632339}
 kind: Secret
 metadata:
   labels:
--- a/tests/test_kubernetes_compiled/minikube-mysql/manifests/mysql_secret.yml
+++ b/tests/test_kubernetes_compiled/minikube-mysql/manifests/mysql_secret.yml
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  MYSQL_ROOT_PASSWORD: P3tncGc6bXlzcWwvcm9vdC9wYXNzd29yZH0=
+  MYSQL_ROOT_PASSWORD: ?{gpg:mysql/root/password:20632339}
 kind: Secret
 metadata:
   labels:
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ You can always override values further up the tree (i.e. in the inventory target
 
 Classifying almost anything will help you avoid repetition (DRY) and will force you to organise parameters hierarchically.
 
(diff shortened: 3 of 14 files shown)


examples/kubernetes/compiled/minikube-mysql/manifests/mysql_secret.yml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  MYSQL_ROOT_PASSWORD: P3tncGc6bXlzcWwvcm9vdC9wYXNzd29yZH0=
+  MYSQL_ROOT_PASSWORD: ?{gpg:mysql/root/password:20632339}
```
ademariag: mind you, this requires users to have a fully gpg setup which might be a PITA. I'd leave it cleartext untile we implement the plain backend.

tests/test_kubernetes_compiled/minikube-mysql/manifests/mysql_secret.yml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  MYSQL_ROOT_PASSWORD: P3tncGc6bXlzcWwvcm9vdC9wYXNzd29yZH0=
+  MYSQL_ROOT_PASSWORD: ?{gpg:mysql/root/password:20632339}
```
ademariag: same here. Perhaps add gpg as a comment to show.

examples/kubernetes/compiled/minikube-mysql/manifests/mysql_secret.yml
adrianchifor: https://github.com/deepmind/kapitan/pull/102

Just made another PR with updates to docs to include details about secrets. We should make the examples as realistic as possible. Also the GPG keys are already in `examples/kubernetes/secrets` so anyone can set them up.

ademariag: not 100% convinced. Ask @gburiola to bless :)
I'd prefer we reduce the chance of the Kapitan crashing on first use because the setup in incomplete. That would look poor.

adrianchifor: It still compiles though without gpg setup

ademariag: right.. I thought you had the randomstr example. That would fail.

gburiola: I like the idea of having a complete example including GPG. Maybe we could add a `README` file at `kapitan/examples/kubernetes` saying that the example uses GPG encrypted secrets. It should compile but in order to apply user would need to install GPG.

Another idea would be to have a `passphrase` encryption backend that asks for the `passphrase` at the command line when you run `kapitan secrets --reveal`

adrianchifor: Added docs in the other PR, will update them to include GPG setup with the example keys. Was thinking of doing the `aes256` backend as well but we have to refactor the current gpg setup.

PR #118, opened 2018-06-08
Do not prune by default

adrianchifor:
Pruning breaks `NetworkPolicy` objects, for example in the default-deny case where `podSelector` and `ingress` have to be defined:
```
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: default-deny
spec:
  podSelector: {}
  ingress: []
```

Default behaviour for no prune is more predictable and clear, with the option of adding `--prune` to compile if you want cleaner output.


ademariag:
I'm afraid of hidden consequences this could have, I'd like this to be tested for longer.

gburiola:
lgtm but agree with Ale. We should probably compile with this new version and apply everywhere while watching for things changing.

we should also test creating new targets from scratch.

This is a big breaking change so we should also bump version to 0.17.0

adrianchifor:
Will merge next week if testing goes ok

kapitan/cli.py
```diff
@@ -64,7 +64,7 @@ def main():
                                 help='set search path, default is "."')
     compile_parser.add_argument('--verbose', '-v', help='set verbose mode',
                                 action='store_true', default=False)
-    compile_parser.add_argument('--no-prune', help='do not prune jsonnet output',
```
gruzewski: Maybe we could keep it for a while and add a warning that it will be removed in the next version?

PR #122, opened 2018-06-15
Fixed secrets validate and update

adrianchifor:
- `kapitan secrets --validate-targets` and `kapitan secrets --update-targets` is now looking at all secrets (broke after the reclass update).
- Allow to validate and update secrets for specific targets `kapitan secrets --update-targets -T minikube-mysql`

### DIFF
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -275,7 +275,10 @@ def main():
             # update recipients for all secrets in secrets_path
             # use --secrets-path to set scanning path
             inv = inventory_reclass(args.inventory_path)
-            targets = set(inv['nodes'].keys())
+            if args.target_name:
+                targets = set(args.target_name.split(','))
+            else:
+                targets = set(inv['nodes'].keys())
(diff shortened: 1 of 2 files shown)


kapitan/cli.py
```diff
@@ -275,7 +275,10 @@ def main():
             # update recipients for all secrets in secrets_path
             # use --secrets-path to set scanning path
             inv = inventory_reclass(args.inventory_path)
-            targets = set(inv['nodes'].keys())
+            if args.target_name:
+                targets = set(args.target_name.split(','))
```
gruzewski: I thought `target_name` has a different purpose according to the [description](https://github.com/deepmind/kapitan/blob/f08caef6a3694642baec43298c634b82670a419d/kapitan/cli.py#L132). Maybe add `--targets` flag for that. Similar to [this](https://github.com/deepmind/kapitan/blob/f08caef6a3694642baec43298c634b82670a419d/kapitan/cli.py#L74).

PR #125, opened 2018-06-21
Command flags support in .kapitan

### LINKED ISSUE #111: Set kapitan default flags in a config file (.kapitan)
For example if we want to compile every time with `--indent 4`, we could make use of the Makefile, but would be nicer to allow configuration of flag defaults in a config file such as `.kapitan`, where we already store the latest kapitan version used for compilation.

`.kapitan` can become something like:
```
version: 0.16.3
compile:
  indent: 4
```
would be equivalent with running `kapitan compile --indent 4`

adrianchifor:
Resolves #111 

- Doc strings consistent with PEP 257
- Allows configuration of command flags in `.kapitan`, such as:
```
version: 0.17.0
compile:
  indent: 4
```
will translate to `kapitan compile --indent 4`
- Enforcing kapitan version from `.kapitan`
- `.kapitan` usage doc

### DIFF
--- a/kapitan/utils.py
+++ b/kapitan/utils.py
@@ -356,11 +356,11 @@ def check_version():
         if kapitan_config and kapitan_config["version"] and StrictVersion(kapitan_config["version"]) > StrictVersion(VERSION):
             print("{}Current version: {}".format(termcolor.WARNING, VERSION))
             print("Version in .kapitan: {}{}\n".format(kapitan_config["version"], termcolor.ENDC))
-            print("Please upgrade kapitan to '{}' in order to keep results consistent:\n".format(kapitan_config["version"]))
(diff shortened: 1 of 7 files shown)


kapitan/utils.py
```diff
@@ -326,30 +317,54 @@ def get_directory_hash(directory):
...
+    Returns the 'flag' for 'command' from .kapitan file. If failed, returns 'default'
+    """
+    kapitan_config = dot_kapitan_config()
+
+    try:
+        if kapitan_config[command]:
+            flag_value = kapitan_config[command][flag]
+            if flag_value:
+                return flag_value
+    except:
```
ramaro: Please turn this into `except: KeyError`

PR #128, opened 2018-06-27
Added rsapublic function to gpg backend

adrianchifor:
Derives an RSA public key from a private key. Usage:
```
private_key: ?{gpg:targets/${target_name}/private_key|rsa}
public_key: ?{gpg:targets/${target_name}/public_key|rsapublic:targets/${target_name}/private_key}
```

### DIFF
--- a/kapitan/secrets.py
+++ b/kapitan/secrets.py
@@ -43,7 +43,7 @@ import kapitan.cached as cached
 logger = logging.getLogger(__name__)
 
 # e.g. ?{gpg:my/secret/token} or ?{gpg:my/secret/token|func:param1:param2}
-SECRET_TOKEN_TAG_PATTERN = r"(\?{([\w\:\.\-\/]+)([\|\w\:]+)*})"
+SECRET_TOKEN_TAG_PATTERN = r"(\?{([\w\:\.\-\/]+)([\|\w\:\.\-\/]+)*})"
 SECRET_TOKEN_ATTR_PATTERN = r"(\w+):([\w\.\-\/]+)"  # e.g. gpg:my/secret/token
 SECRET_TOKEN_COMPILED_ATTR_PATTERN = r"(\w+):([\w\.\-\/]+):(\w+)"  # e.g. gpg:my/secret/token:1deadbeef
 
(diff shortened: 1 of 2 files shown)


kapitan/secrets.py
```diff
@@ -54,36 +54,36 @@
 
 
 class GPGError(Exception):
-    "Generic GPG errors"
+    """Generic GPG errors"""
```
gruzewski: Nit: Next time would be good to separate those or mention it in PR.

kapitan/secrets.py
```diff
@@ -433,6 +432,24 @@ def rsa_private_key(key_size=''):
     ), "UTF-8")
 
 
+def rsa_public_key(secrets_path, private_key_file):
+    """derives an RSA public key from private_key_file"""
```
gruzewski: Nit: `derives a RSA`

kapitan/secrets.py
```diff
@@ -433,6 +432,24 @@ def rsa_private_key(key_size=''):
...
+    token = "gpg:%s" % private_key_file
```
gruzewski: Nit: We could use the new string interpolation format `"gpg:{}".format(private_key_file)`

kapitan/secrets.py
```diff
@@ -433,6 +432,24 @@ def rsa_private_key(key_size=''):
...
+    secret_raw_obj = secret_gpg_raw_read(secrets_path, token)
+    data_dec = secret_gpg_read(secrets_path, token)
+    encode_base64 = secret_raw_obj.get('encoding', None) == 'base64'
+
+    if encode_base64:
+        data_dec = base64.b64decode(data_dec).decode('UTF-8')
+
+    private_key = serialization.load_pem_private_key(data_dec.encode(), password=None, backend=default_backend())
+
+    return str(private_key.public_key().public_bytes(
```
gruzewski: Nit: IMO it would be easier to read if you extract generating public key to its own variable.

kapitan/secrets.py
adrianchifor: derives an RSA is correct

kapitan/secrets.py
adrianchifor: missed these on the other MR

PR #134, opened 2018-07-28
Inventory and folders caching - Only compile targets that changed

adrianchifor:
- Creates `compiled/.kapitan_cache` on compilations:
```
inventory:
    <target>:
        classes: <sha256> 
        parameters: <sha256>
        ...
# Base folders from `input_paths`
folder:
    components: <sha256> 
    docs: <sha256> 
    lib: <sha256> 
    scripts: <sha256> 
    ... 
```
- On next compilation it will re-generate the cache and compare against saved `.kapitan_cache` to find which targets changed. For now, when a folder changes, we recompile everything as we might have hidden references in jsonnet/jinja2.
- Recompilation can be forced with `kapitan compile -f`, which will ignore `.kapitan_cache`.
- You can still recompile specific targets with `kapitan compile -t minikube-es` and it will not check cache.


gburiola:
This is a brilliant idea !!!

We should probably always use `--force-recompile` in our CI tests to ensure that everything is being compiled correctly

gburiola:
Another issue is if you only change a secret and try to compile it says `No changes since last compilation.`

kapitan/targets.py
```diff
@@ -105,6 +122,96 @@ def compile_targets(inventory_path, search_path, output_path, parallel, targets,
...
+                base_folder = os.path.dirname(input_path).split('/')[0]
+                if base_folder == '':
+                    base_folder = os.path.basename(input_path).split('/')[0]
+
+                if base_folder not in cached.inv_cache['folder'].keys():
+                    if os.path.exists(base_folder) and os.path.isdir(base_folder):
+                        cached.inv_cache['folder'][base_folder] = directory_hash(base_folder)
+
+            # Most commonly changed but not referenced in input_paths
+            for common in ('lib', 'vendor'):
```
gburiola: maybe we should add `secrets` here as well.

kapitan/utils.py
```diff
@@ -309,18 +311,22 @@ def get_directory_hash(directory):
                     with open(file_path, 'r') as f:
                         hash.update(sha256(f.read().encode("UTF-8")).hexdigest().encode("UTF-8"))
```
gburiola: why do you need to `encode("UTF-8"` to generate the `directory_hash` ?

If you have any binary files in the code base then this will break.

adrianchifor: `sha256()` doesn't like it if we don't encode the result of read. I'll update the code to support binary files.

adrianchifor: Just tested, binary files should be supported now,

adrianchifor:
@gburiola Updated the code to only modify `inventory.<targets>` when `kapitan compile -t <targets>` is used.

tests/test_compile.py
```diff
@@ -20,18 +20,22 @@
...
         os.chdir(os.getcwd() + '/examples/kubernetes/')
 
     def test_compile(self):
-        sys.argv = ["kapitan", "compile"]
+        sys.argv = ["kapitan", "compile", "-f"]
         main()
-        compiled_dir_hash = get_directory_hash(os.getcwd() + '/compiled')
-        test_compiled_dir_hash = get_directory_hash(os.getcwd() + '/../../tests/test_kubernetes_compiled')
+        try:
+            os.remove(os.getcwd() + '/compiled/.kapitan_cache')
```
gburiola: I think instead of ignoring `/compiled/.kapitan_cache`, we should actually add it to the `tests/test_kubernetes_compiled` folder.

This way we can test the cache file is being generated correctly

gburiola:
looks good to me

PR #138, opened 2018-08-01
Support for additional cache paths. Made cache an optional flag. Cleanup

adrianchifor:
- Ability to add more folders to `.kapitan_cache`: 
```
kapitan compile --cache --cache-paths path1 path2
```
or
```
cat > .kapitan
compile:
  cache: true
  cache-paths:
    - path1
    - path2

kapitan compile
```


tests/test_compile.py
```diff
@@ -28,7 +28,7 @@ def setUp(self):
         os.chdir(os.getcwd() + '/examples/kubernetes/')
 
     def test_compile(self):
-        sys.argv = ["kapitan", "compile", "-f"]
```
gburiola: I think we should have a test for the cache functionality as well.

Maybe 2 separate tests

- kapitan compile
- kapitan compile --cache

adrianchifor: Will write a test in another MR

PR #139, opened 2018-08-06
compile arguments force-recompile and targets are mutually exclusive

gburiola:
```
$ ~/git/kapitan/bin/kapitan compile -t my_target -f
usage: kapitan compile [-h] [--search-paths JPATH [JPATH ...]] [--verbose]
                       [--prune] [--quiet] [--output-path PATH]
                       [--targets TARGET [TARGET ...] | --force-recompile]
                       [--parallelism INT] [--indent INT]
                       [--secrets-path SECRETS_PATH] [--reveal]
                       [--inventory-path INVENTORY_PATH]
                       [--ignore-version-check]
kapitan compile: error: argument --force-recompile/-f: not allowed with argument --targets/-t
```

### DIFF
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -80,10 +80,15 @@ def main():
                                 default=from_dot_kapitan('compile', 'output-path', '.'),
                                 metavar='PATH',
                                 help='set output path, default is "."')
-    compile_parser.add_argument('--targets', '-t', help='targets to compile, default is all',
+    compile_parser_subgroup = compile_parser.add_mutually_exclusive_group()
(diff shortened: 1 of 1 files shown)


kapitan/cli.py
```diff
@@ -105,10 +110,6 @@ def main():
                                 help='ignore the version from .kapitan',
                                 action='store_true',
                                 default=from_dot_kapitan('compile', 'ignore-version-check', False))
-    compile_parser.add_argument('--force-recompile', '-f',
```
gburiola: just moved this block below `--targets`

PR #149, opened 2018-09-23
Add python_requires

Code0x58:
[[docs](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires)]


adrianchifor:
Thanks for the contribution!

Issue #150, opened 2018-10-02
GPG encrypt passes binary blob to secrets_file['data'] instead of base64


adrianchifor:
`kapitan secrets --write/--update/--update-targets ...` does not base64 the encrypted data from secrets, instead it writes the encrypted binary blob directly to yaml:
```
data: !!binary |
...
```

theinrichs:
@adrianchifor 
Is this really fixed? I just tested kapitan 0.19.0 update-targets and the secrets still have
```
data: !!binary |
  hQIMA3wpiKskwq5cARAAyaxFH4kfy6nn5wYEkSv7vmhZ1b32nLrXXRCJgGL2EbA85kj3ixb8GiOz
```
When you try `kapitan compile --reveal`
you get

```
kapitan.refs.secrets.gpg.GPGError: no data was provided
```

ramaro:
@theinrichs was your secret created before 0.19.0?

ramaro:
`data:` should be not be !!binary - that was a bug introduced in master but I don't think it was ever part of a release. I can compile --reveal my pre 0.19.0 just fine.

adrianchifor:
Still happening on --update/--update-targets

PR #153, opened 2018-10-19
Fix secrets cli

ramaro:
Fixes encoding issues when creating gpg secrets using the cli (#150)
Breaking change: removes `--backend` in `secrets` subcommand. New format for secret token in cli is prefix with backend name (now referred as type_name) e.g. `gpg:path/to/secret1`
Refactor of cli code into smaller functions easier to test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,9 +401,9 @@ The usual flow of creating and using an encrypted secret with kapitan is:
 
   - Manually:
     ```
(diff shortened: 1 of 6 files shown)


rjmco:
CLA signed. A re-test should return a clean result.

ramaro:
CLA updated

PR #154, opened 2018-10-20
Fix cli secrets

ramaro:
Fixes encoding issues when creating gpg secrets using the cli (#150)
Fixes wrong file argument in cli.py
Breaking change: removes --backend in secrets subcommand. New format for secret token in cli is prefix with backend name (now referred as type_name) e.g. gpg:path/to/secret1
Refactor of cli code into smaller functions easier to test.
Test secret cli write and reveal

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,9 +401,9 @@ The usual flow of creating and using an encrypted secret with kapitan is:
 
   - Manually:
     ```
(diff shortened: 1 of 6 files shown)


rjmco:
I have read and signed the CLA. I confirm the contribution of my commit to this project as per the CLA.

PR #167, opened 2018-11-22
gkms update and validate secrets

### LINKED ISSUE #150: GPG encrypt passes binary blob to secrets_file['data'] instead of base64

adrianchifor:
- Moved `search_target_token_paths` to refs/base.py as it's not gpg specific
- Implemented update and validate for gkms secrets
- Fixes #150

### DIFF
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -320,6 +320,9 @@ def secret_write(args, ref_controller):
                 logger.warning("WARNING: parameters.kapitan.secrets.recipients is deprecated, " +
                     "please use parameters.kapitan.secrets.gpg.recipients")
                 recipients = inv['nodes'][args.target_name]['parameters']['kapitan']['secrets']['recipients']
+        if not recipients:
+            raise KapitanError("No GPG recipients specified. Use --recipients or specify in " +
+                               "parameters.kapitan.secrets.gpg.recipients and use --target")
         secret_obj = GPGSecret(data, recipients, encode_base64=args.base64)
         tag = '?{{gpg:{}}}'.format(token_path)
         ref_controller[tag] = secret_obj
@@ -339,7 +342,7 @@ def secret_write(args, ref_controller):
         tag = '?{{gkms:{}}}'.format(token_path)
         ref_controller[tag] = secret_obj
     else:
-        fatal_error("Invalid token: {}".format(token_name))
+        fatal_error("Invalid token: {}. Try using gpg/gkms:{}".format(token_name, token_name))
 
 
 def secret_update(args, ref_controller):
@@ -361,6 +364,9 @@ def secret_update(args, ref_controller):
                 logger.warning("WARNING: parameters.kapitan.secrets.recipients is deprecated, " +
                     "please use parameters.kapitan.secrets.gpg.recipients")
                 recipients = inv['nodes'][args.target_name]['parameters']['kapitan']['secrets']['recipients']
+        if not recipients:
+            raise KapitanError("No GPG recipients specified. Use --recipients or specify in " +
+                               "parameters.kapitan.secrets.gpg.recipients and use --target")
         type_name, token_path = token_name.split(":")
         tag = '?{{gpg:{}}}'.format(token_path)
         secret_obj = ref_controller[tag]
@@ -368,7 +374,7 @@ def secret_update(args, ref_controller):
         ref_controller[tag] = secret_obj
     # TODO: Implement --update for KMS
     else:
-        fatal_error("Invalid token: {}".format(token_name))
+        fatal_error("Invalid token: {}. Try using gpg/gkms:{}".format(token_name, token_name))
 
 
 def secret_reveal(args, ref_controller):
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -32,9 +32,9 @@ from kapitan.targets import compile_targets
 from kapitan.resources import search_imports, resource_callbacks, inventory_reclass
 from kapitan.version import PROJECT_NAME, DESCRIPTION, VERSION
 
-from kapitan.refs.base import RefController, Revealer
+from kapitan.refs.base import RefController, Revealer, search_target_token_paths
 from kapitan.refs.secrets.gpg import GPGBackend, GPGSecret
-from kapitan.refs.secrets.gpg import lookup_fingerprints, search_target_token_paths
+from kapitan.refs.secrets.gpg import lookup_fingerprints
 from kapitan.refs.secrets.gkms import GoogleKMSSecret
 
 from kapitan.errors import KapitanError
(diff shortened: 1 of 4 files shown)


kapitan/cli.py
```diff
@@ -320,6 +320,9 @@ def secret_write(args, ref_controller):
                 logger.warning("WARNING: parameters.kapitan.secrets.recipients is deprecated, " +
                     "please use parameters.kapitan.secrets.gpg.recipients")
                 recipients = inv['nodes'][args.target_name]['parameters']['kapitan']['secrets']['recipients']
+        if not recipients:
+            raise KapitanError("No GPG recipients specified. Use --recipients or specify in " +
```
gruzewski: `Use --recipients or specify *them* in`

kapitan/cli.py
```diff
@@ -339,11 +342,11 @@ def secret_write(args, ref_controller):
         tag = '?{{gkms:{}}}'.format(token_path)
         ref_controller[tag] = secret_obj
     else:
-        fatal_error("Invalid token: {}".format(token_name))
+        fatal_error("Invalid token: {}. Try using gpg/gkms:{}".format(token_name, token_name))
```
gruzewski: `fatal_error("Invalid token: {name}. Try using gpg/gkms:{name}".format(name=token_name))`

kapitan/cli.py
gruzewski: You could move it to just after assignment in line 310.

```
recipients = [dict((("name", name),)) for name in args.recipients]

if not recipients:
    raise KapitanError(..)
```

kapitan/cli.py
```diff
@@ -361,14 +364,33 @@ def secret_update(args, ref_controller):
                 logger.warning("WARNING: parameters.kapitan.secrets.recipients is deprecated, " +
                     "please use parameters.kapitan.secrets.gpg.recipients")
                 recipients = inv['nodes'][args.target_name]['parameters']['kapitan']['secrets']['recipients']
+        if not recipients:
```
gruzewski: Same here.
```
recipients = [dict((("name", name),)) for name in args.recipients]

if not recipients:
    raise KapitanError(..)
```

kapitan/cli.py
```diff
@@ -361,14 +364,33 @@ def secret_update(args, ref_controller):
...
+    elif token_name.startswith("gkms:"):
+        key = args.key
+        if args.target_name:
+            inv = inventory_reclass(args.inventory_path)
+            if 'secrets' not in inv['nodes'][args.target_name]['parameters']['kapitan']:
+                raise KapitanError("parameters.kapitan.secrets not defined in {}".format(args.target_name))
+
+            key = inv['nodes'][args.target_name]['parameters']['kapitan']['secrets']['gkms']['key']
+        if not key:
+            raise KapitanError("No KMS key specified. Use --key or specify in parameters.kapitan.secrets.gkms.key and use --target")
```
gruzewski: `Use --key or specify *it* in parameters.kapitan.secrets.gkms.key`

kapitan/cli.py
```diff
@@ -361,14 +364,33 @@ def secret_update(args, ref_controller):
...
         type_name, token_path = token_name.split(":")
         tag = '?{{gpg:{}}}'.format(token_path)
         secret_obj = ref_controller[tag]
         secret_obj.update_recipients(recipients)
         ref_controller[tag] = secret_obj
-    # TODO: Implement --update for KMS
+
+    elif token_name.startswith("gkms:"):
+        key = args.key
+        if args.target_name:
```
gruzewski: You can extract `args.target_name` to a variable. You have used it 4 times.

kapitan/cli.py
gruzewski: You can move it to line 378. It's easier to read and you can fail faster, before doing any computation.

kapitan/cli.py
```diff
@@ -361,14 +364,33 @@ def secret_update(args, ref_controller):
...
+            inv = inventory_reclass(args.inventory_path)
+            if 'secrets' not in inv['nodes'][args.target_name]['parameters']['kapitan']:
+                raise KapitanError("parameters.kapitan.secrets not defined in {}".format(args.target_name))
+
+            key = inv['nodes'][args.target_name]['parameters']['kapitan']['secrets']['gkms']['key']
```
gruzewski: You can extract `inv['nodes'][args.target_name]['parameters']['kapitan']` to a variable.

kapitan/cli.py
```diff
@@ -409,19 +430,46 @@ def secret_update_validate(args, ref_controller):
...
+                if not recipients:
+                    logger.debug("secret_update_validate: target: %s has no inventory gpg recipients, skipping %s", target_name, token_path)
+                    continue
+                secret_obj = ref_controller[token_path]
                 target_fingerprints = set(lookup_fingerprints(recipients))
                 secret_fingerprints = set(lookup_fingerprints(secret_obj.recipients))
                 if target_fingerprints != secret_fingerprints:
                     if args.validate_targets:
-                        logger.info("%s recipient mismatch", token_path)
+                        logger.info("%s recipients mismatch", token_path)
```
gruzewski: You could maybe add which fingerprints are different with:
```
target_fingerprints .symmetric_difference(secret_fingerprints)
```

kapitan/cli.py
adrianchifor: the location is correct, it needs to be done after checking the `args.target_name`. Will fix the typo in print

kapitan/cli.py
adrianchifor: location is correct, needs to be done after `args.target_name` check

kapitan/cli.py
adrianchifor: it is already a variable...

kapitan/cli.py
adrianchifor: I think the current way is more readable. If we extract it into a variable, it would need reassignment every target loop which is not ideal.

kapitan/cli.py
adrianchifor: good idea

kapitan/cli.py
adrianchifor: good idea

kapitan/refs/secrets/gkms.py
```diff
@@ -93,6 +93,21 @@ def reveal(self):
         ref_data = base64.b64decode(self.data)
         return self._decrypt(ref_data)
 
+    def update_key(self, key):
+        """
+        re-encrypts data with new key, respects original encoding
+        returns True if key is different and secret is updated, False otherwise
+        """
+        if key != self.key:
```
gruzewski: It's easier to read if you revers this logic:
```
if key == self.key:
    return False
...
return True
```

adrianchifor: good point

PR #176, opened 2018-12-18
Minor changes

uberspot:
These should mostly be a no-op.

Changed:

 - Added contributing section in readme
 - Added codestyle test in makefile
 - Several minor readme/doc related nitpicks
 - Updated kube.libjsonnet and the example minikube version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,9 @@ package:
 .PHONY: clean
 clean:
 	rm -rf dist/ build/ kapitan.egg-info/
+
+pycodestyle:
+	which pycodestyle || pip3 install --user pycodestyle
+	# ignores line length and reclass related errors
+	pycodestyle --ignore=E501 . | grep -v "reclass"
+	@echo
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,9 @@ package:
 clean:
 	rm -rf dist/ build/ kapitan.egg-info/
 
+.PHONY: pycodestyle
 pycodestyle:
-	which pycodestyle || pip3 install --user pycodestyle
+	which pycodestyle || echo "Install pycodestyle with pip3 install --user pycodestyle"
 	# ignores line length and reclass related errors
 	pycodestyle --ignore=E501 . | grep -v "reclass"
 	@echo
--- a/Makefile
+++ b/Makefile
@@ -20,9 +20,9 @@ package:
 clean:
 	rm -rf dist/ build/ kapitan.egg-info/
 
-.PHONY: pycodestyle
-pycodestyle:
-	which pycodestyle || echo "Install pycodestyle with pip3 install --user pycodestyle"
+.PHONY: codestyle
+codestyle:
+	which flake8 || echo "Install flake8 with pip3 install --user flake8"
 	# ignores line length and reclass related errors
-	pycodestyle --ignore=E501 . | grep -v "reclass"
+	flake8 --ignore E501 . | grep -v "reclass"
 	@echo
--- a/examples/kubernetes/lib/kube.libjsonnet
+++ b/examples/kubernetes/lib/kube.libjsonnet
@@ -134,7 +134,6 @@ local inv = kap.inventory();
   Container(name): {
     name: name,
     image: error "container image value required",
-
     envList(map):: [
       if std.type(map[x]) == "object" then { name: x, valueFrom: map[x] } else { name: x, value: std.toString(map[x]) }
       for x in std.objectFields(map)],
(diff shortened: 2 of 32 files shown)


Makefile
```diff
@@ -19,3 +19,10 @@ package:
 .PHONY: clean
 clean:
 	rm -rf dist/ build/ kapitan.egg-info/
+
+.PHONY: pycodestyle
+pycodestyle:
+	which pycodestyle || pip3 install --user pycodestyle
```
adrianchifor: Some people might have pip as virtualenv, we should just print how to install it and exit

uberspot: Fixed

examples/kubernetes/lib/kube.libjsonnet
```diff
@@ -171,13 +173,26 @@ local inv = kap.inventory();
 
     imagePullSecrets: [],
 
+    nodeSelector: {
+      [if "nodepools" in inv.parameters.cluster && "nodeselector" in inv.parameters then inv.parameters.cluster.nodepools[inv.parameters.nodeselector].label]: inv.parameters.cluster.nodepools[inv.parameters.nodeselector].name,
```
adrianchifor: This is specific to us, not sure other people might want this

kapitan/utils.py
```diff
@@ -388,26 +388,21 @@ def check_version():
     """
     kapitan_config = dot_kapitan_config()
     try:
+        print("{}Current version: {}".format(termcolor.WARNING, VERSION))
```
adrianchifor: Won't this print every time?

kapitan/targets.py
```diff
@@ -100,21 +99,21 @@ def compile_targets(inventory_path, search_paths, output_path, parallel, targets
 
         # Save inventory and folders cache
         save_inv_cache(compile_path, targets)
-
+        pool.close()
```
adrianchifor: Can you test this doesn't impact speed with multiple `-p` options in compile, I feel like this might have unintended consequences

examples/kubernetes/lib/kube.libjsonnet
ramaro: I agree

kapitan/cli.py
```diff
@@ -257,8 +256,7 @@ def main():
                 yaml.dump(inv, sys.stdout, Dumper=PrettyDumper, default_flow_style=False)
         except Exception as e:
             if not isinstance(e, KapitanError):
-                logger.error("\n~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~")
-                traceback.print_exc()
+                logger.exception("\n~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~")
```
ramaro: does logger.exception include the same as traceback?

kapitan/targets.py
ramaro: Why close here?

kapitan/targets.py
```diff
@@ -100,21 +99,21 @@ def compile_targets(inventory_path, search_paths, output_path, parallel, targets
...
     except Exception as e:
         # if compile worker fails, terminate immediately
         pool.terminate()
-        pool.join()
```
ramaro: Not joining here might obfuscate the "Compile pool terminated" with other output.

kapitan/utils.py
ramaro: I don't think we want to change this.

kapitan/targets.py
uberspot: tested with -p 16 and -p 8, output is the same as latest kapitan.

kapitan/utils.py
uberspot: good spot, fixed.

Makefile
```diff
@@ -19,3 +19,10 @@ package:
 .PHONY: clean
 clean:
 	rm -rf dist/ build/ kapitan.egg-info/
+
+.PHONY: pycodestyle
+pycodestyle:
```
ramaro: I would argue we should use flake8 instead.

kapitan/targets.py
ramaro: But why?

kapitan/cli.py
uberspot: Based on this https://docs.python.org/3.7/library/logging.html#logging.Logger.exception it also includes the stacktrace and I think is more pythonic (?).

kapitan/targets.py
uberspot: Because we don't have more work to submit after that point (?) https://stackoverflow.com/questions/38271547/when-should-we-call-multiprocessing-pool-join I might be wrong, it's just what I understood from here ^ and what I saw in other examples of pool usage.

Makefile
uberspot: switched to flake8 :+1:

kapitan/targets.py
uberspot: tested and stacktraces appear as they should.

PR #180, opened 2018-12-20
Refactor input_type code into classes

ramaro:
Introduces a new InputType class which serves as the base class for 2 new input_type compilers:
* Jsonnet
* Jinja2

Also fixes minor code style issues

### DIFF
--- a/kapitan/inputs/base.py
+++ b/kapitan/inputs/base.py
@@ -76,12 +76,7 @@ class InputType(object):
         """make compile dirs, skips if dirs exist"""
         _compile_path = os.path.join(self.compile_path, target_name, output_path)
         # support writing to an already existent dir
-        try:
-            os.makedirs(_compile_path)
-        except OSError as ex:
-            # If directory exists, pass
-            if ex.errno == errno.EEXIST:
-                pass
+        os.makedirs(_compile_path, exist_ok=True)
 
     def compile_file(self, file_path, compile_path, ext_vars, **kwargs):
         """implements compilation for file_path to compile_path with ext_vars"""
--- a/kapitan/inputs/base.py
+++ b/kapitan/inputs/base.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python3
+#
+# Copyright 2018 The Kapitan Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import errno
+import logging
+import os
+import yaml
+import ujson as json
+
+from kapitan.errors import CompileError, KapitanError
+from kapitan.refs.base import Revealer
+from kapitan.utils import PrettyDumper
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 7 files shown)


kapitan/inputs/base.py
```diff
@@ -0,0 +1,147 @@
...
+        if not file_found:
+            raise CompileError("Compile error: {} for target: {} not found in "
+                               "search_paths: {}".format(input_path, target_name, self.search_paths))
+
+    def make_compile_dirs(self, target_name, output_path):
+        """make compile dirs, skips if dirs exist"""
+        _compile_path = os.path.join(self.compile_path, target_name, output_path)
+        # support writing to an already existent dir
+        try:
+            os.makedirs(_compile_path)
```
uberspot: Use `os.makedirs(_compile_path, exist_ok=True)` then you can remove the try-catch block you have in case it fails.

ramaro: Good point. Done

kapitan/inputs/jinja2.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # set ext_vars and inventory for jinja2 context
+        context = ext_vars.copy()
+        context["inventory"] = inventory(self.search_paths, target_name)
+        context["inventory_global"] = inventory(self.search_paths, None)
+
+        for item_key, item_value in render_jinja2(file_path, context).items():
+            full_item_path = os.path.join(compile_path, item_key)
+            try:
+                os.makedirs(os.path.dirname(full_item_path))
```
uberspot: Same here, `exist_ok=True`

kapitan/inputs/jsonnet.py
```diff
@@ -0,0 +1,79 @@
...
+                with CompiledFile(file_path, self.ref_controller, mode="w", reveal=reveal, target_name=target_name,
+                                  indent=indent) as fp:
+                    fp.write_json(item_value)
+                    logger.debug("Wrote %s", file_path)
+            elif output == 'yaml':
+                file_path = os.path.join(compile_path, '%s.%s' % (item_key, "yml"))
+                with CompiledFile(file_path, self.ref_controller, mode="w", reveal=reveal, target_name=target_name,
+                                  indent=indent) as fp:
+                    fp.write_yaml(item_value)
+                    logger.debug("Wrote %s", file_path)
```
uberspot: Would it make sense to move this `logger.debug` call in a write_yaml super() function in CompiledFile ? Then you don't have to explicitly add it everywhere again and you always have a debug log.

kapitan/inputs/jinja2.py
ramaro: done

kapitan/targets.py
```diff
@@ -288,145 +287,27 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, **kwa
...
-        with CompiledFile(full_item_path, ref_controller, mode="w", reveal=reveal, target_name=target_name) as fp:
-            fp.write(item_value["content"])
-            mode = item_value["mode"]
-            os.chmod(full_item_path, mode)
-            logger.debug("Wrote %s with mode %.4o", full_item_path, mode)
+            input_compiler = jinja2_compiler
+        elif input_type == "jsonnet":
+            input_compiler = jsonnet_compiler
+        else:
+            raise CompileError("")
```
adrianchifor: We should print that the input type is not valid, and which ones we support

ramaro: Oops, forgot that. Will fix.

kapitan/inputs/jsonnet.py
ramaro: Yep

ramaro: done

kapitan/targets.py
ramaro: done

PR #183, opened 2018-12-21
Minor cleanup

uberspot:
Some more no-op cleanup + a breaking change since there is one more in this release candidate.

Specifically:

 - fix various codestyle errors
 - fix a TODO with missing exception handling
 - simplify os.makedirs everywhere
 - move search_target_token_paths to kapitan.utils since it's only used in cli and is not logically bound to kapitan/refs/base.py
 - Update changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.22.0-rc.0:
 #### Breaking:
 - Added reveal secrets function, updated rsapublic function (#182)
-- Parameters.kapitan.secrets.recipients is deprecated, please use parameters.kapitan.secrets.gpg.recipients (#183)
+- parameters.kapitan.secrets.recipients is deprecated, please use parameters.kapitan.secrets.gpg.recipients (#183)
 
 #### Updates:
 - Added AWS KMS support as a secrets backend(#179)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## 0.22.0-rc.0:
+#### Breaking:
+- Added reveal secrets function, updated rsapublic function (#182)
+- Parameters.kapitan.secrets.recipients is deprecated, please use parameters.kapitan.secrets.gpg.recipients (#183)
+
+#### Updates:
+- Added AWS KMS support as a secrets backend(#179)
+- Refactor input\_type code into classes(#180)
+- Fix GPG failing when expiry of gpg key is infinite (#181)
+
 ## 0.21.0
 - Upgraded to jsonnet 0.12.1 (https://github.com/google/jsonnet/releases/tag/v0.12.1)
 
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -206,8 +206,6 @@ def main():
         logging.basicConfig(level=logging.INFO, format="%(message)s")
 
     if cmd == 'eval':
-        logging.basicConfig(level=logging.WARNING)
-
         file_path = args.jsonnet_file
         search_paths = [os.path.abspath(path) for path in args.search_paths]
         ext_vars = {}
@@ -419,7 +417,7 @@ def secret_reveal(args, ref_controller):
             for rev_obj in revealer.reveal_path(file_name):
                 sys.stdout.write(rev_obj.content)
     except (RefHashMismatchError, KeyError):
-        logger.exception("Reveal failed for file {name}".format(name=file_name))
+        raise KapitanError("Reveal failed for file {name}".format(name=file_name))
 
 
 def secret_update_validate(args, ref_controller):
@@ -437,13 +435,21 @@ def secret_update_validate(args, ref_controller):
         if 'secrets' not in kap_inv_params:
             raise KapitanError("parameters.kapitan.secrets not defined in {}".format(target_name))
 
+        try:
+            recipients = kap_inv_params['secrets']['gpg']['recipients']
+        except KeyError:
+            recipients = None
+        try:
+            gkey = kap_inv_params['secrets']['gkms']['key']
+        except KeyError:
+            gkey = None
+        try:
+            awskey = kap_inv_params['secrets']['awskms']['key']
+        except KeyError:
+            awskey = None
+
         for token_path in token_paths:
             if token_path.startswith("?{gpg:"):
-                try:
-                    recipients = kap_inv_params['secrets']['gpg']['recipients']
-                except KeyError:
-                    recipients = None
-
                 if not recipients:
                     logger.debug("secret_update_validate: target: %s has no inventory gpg recipients, skipping %s", target_name, token_path)
                     continue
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.22.0-rc.0:
+#### Breaking:
+- Added reveal secrets function, updated rsapublic function (#182)
+- Parameters.kapitan.secrets.recipients is deprecated, please use parameters.kapitan.secrets.gpg.recipients (#183)
```
adrianchifor: small p

kapitan/cli.py
```diff
@@ -196,7 +197,17 @@ def main():
         parser.print_help()
         sys.exit(1)
 
+    if hasattr(args, 'verbose') and args.verbose:
```
adrianchifor: It's a bit more pythonic to do `if 'verbose' in args`

kapitan/cli.py
```diff
@@ -438,13 +411,15 @@ def secret_reveal(args, ref_controller):
...
-        for rev_obj in revealer.reveal_path(file_name):
-            sys.stdout.write(rev_obj.content)
+    try:
+        if file_name == '-':
+            out = revealer.reveal_raw_file(None)
+            sys.stdout.write(out)
+        elif file_name:
+            for rev_obj in revealer.reveal_path(file_name):
+                sys.stdout.write(rev_obj.content)
+    except (RefHashMismatchError, KeyError):
```
adrianchifor: What would throw KeyError here?

kapitan/cli.py
```diff
@@ -462,29 +437,13 @@ def secret_update_validate(args, ref_controller):
...
-
-        try:
-            awskey = kap_inv_params['secrets']['awskms']['key']
-        except KeyError:
-            awskey = None
-
         for token_path in token_paths:
             if token_path.startswith("?{gpg:"):
+                try:
+                    recipients = kap_inv_params['secrets']['gpg']['recipients']
```
adrianchifor: Do we want to redefine recipients for every token_path? I believe that's why it was before the for loop.

kapitan/cli.py
```diff
@@ -196,7 +197,17 @@ def main():
...
+    if hasattr(args, 'verbose') and args.verbose:
+        logging.basicConfig(level=logging.DEBUG,
+                            format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s')
+    elif hasattr(args, 'quiet') and args.quiet:
+        logging.basicConfig(level=logging.CRITICAL, format="%(message)s")
+    else:
+        logging.basicConfig(level=logging.INFO, format="%(message)s")
+
     if cmd == 'eval':
+        logging.basicConfig(level=logging.WARNING)
```
adrianchifor: Why warning for eval?

kapitan/cli.py
uberspot: That would check if 'verbose' is in the args dictionary no? 
I want to check if the object args, has the attribute .verbose first because it might not be defined in the args parsed.
```
python -i
class Object(object):
    pass
a = Object()
a.somefield = 'somevalue'
if hasattr(a, 'somefield'): 
...   print("T")
... 
T
if 'somefield' in a: 
...   print("T")
... 
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'Object' is not iterable
```
Am I missing something?

kapitan/cli.py
uberspot: I added it because it was mentioned in the # TODO comment.
After looking at the code a bit I think it might be thrown by
```
self.ref_controller[tag]
``` in `_reveal_replace_match` reachable in the Revealer (?).
If it's not that, I could remove it.

File: kapitan/cli.py (hunk 4 above)
Comment (uberspot) [reply kapitan/cli.py]: Oops, good spot.

File: kapitan/cli.py (hunk 5 above)
Comment (uberspot) [reply kapitan/cli.py]: It was set to warning previously as well (that's the default level for the python logging module).
All other operations were defaulting to INFO or DEBUG except eval. So I thought that was by intention and kept it the same. I could also just keep it at INFO when using eval if we prefer that(?).

File: kapitan/cli.py (hunk 2 above)
Comment (adrianchifor) [reply kapitan/cli.py]: Oh right, thought args is a dictionary, it's actually a `Namespace` object.

File: kapitan/cli.py (hunk 3 above)
Comment (adrianchifor) [reply kapitan/cli.py]: cool, was curious

File: kapitan/cli.py (hunk 5 above)
Comment (adrianchifor) [reply kapitan/cli.py]: Yea we should use INFO by default, not sure why eval was at WARNING, maybe @ramaro might know more

File: kapitan/cli.py [hunk 6]
```diff
@@ -438,13 +409,15 @@ def secret_reveal(args, ref_controller):
...
-            sys.stdout.write(rev_obj.content)
+    try:
+        if file_name == '-':
+            out = revealer.reveal_raw_file(None)
+            sys.stdout.write(out)
+        elif file_name:
+            for rev_obj in revealer.reveal_path(file_name):
+                sys.stdout.write(rev_obj.content)
+    except (RefHashMismatchError, KeyError):
+        logger.exception("Reveal failed for file {name}".format(name=file_name))
```
ramaro: Maybe should not logger.exception here but instead raise a KapitanError with that same message

kapitan/refs/secrets/awskms.py
```diff
@@ -137,8 +138,9 @@ def dump(self):
         """
         Returns dict with keys/values to be serialised.
         """
-        return {"data": self.data, "encoding": self.encoding,
```
ramaro: I'd argue this was more readable

kapitan/refs/secrets/gkms.py
```diff
@@ -69,10 +69,11 @@ def from_params(cls, data, ref_params):
...
         try:
             target_name = ref_params.kwargs['target_name']
-            target_inv = cached.inv['nodes'].get(target_name, None)
-            if target_name is None:
+            if not target_name:
                 raise ValueError('target_name not set')
-            if target_inv is None:
+
+            target_inv = cached.inv['nodes'].get(target_name, None)
+            if not target_inv:
```
ramaro: I find comparing to None more explicit and less magical, also more readable since .get() is defaulting to None

kapitan/refs/secrets/gkms.py
```diff
@@ -155,8 +156,9 @@ def dump(self):
         """
         Returns dict with keys/values to be serialised.
         """
-        return {"data": self.data, "encoding": self.encoding,
```
ramaro: Same

kapitan/refs/secrets/gpg.py
```diff
@@ -80,22 +79,18 @@ def from_params(cls, data, ref_params):
                 return cls(data, _fingerprints, **ref_params.kwargs)
 
             target_name = ref_params.kwargs['target_name']
-            target_inv = cached.inv['nodes'].get(target_name, None)
-            if target_name is None:
+            if not target_name:
```
ramaro: same

kapitan/cli.py
ramaro: Not sure why it is WARNING 🤔  We definitely should use INFO

kapitan/refs/secrets/awskms.py
uberspot: fixed all :+1:

PR #190, opened 2019-01-09
Kadet input type *experimental*

ramaro:
This introduces a new experimental input type called Kadet.

Kadet is essentially a Python module offering a set of classes and functions to define objects which will compile to JSON or YAML. A complete example is available in `examples/kubernetes/components/nginx`.

# Overview

## BaseObj
BaseObj implements the basic object implementation that compiles into JSON or YAML.
Setting keys in `self.root` means they will be in the compiled output. Keys can be set as an hierarchy of attributes (courtesy of [addict](https://github.com/mewwts/addict))
The `self.body()` method is reserved for setting self.root on instantiation: 

The example below:

```python
class MyApp(BaseObj):
 def body(self):
   self.root.name = "myapp"
   self.root.inner.foo = "bar"
   self.root.list = [1, 2, 3]
```

compiles into:

```yaml
---
name: myapp
inner:
  foo: bar
list:
  - 1
  - 2
  - 3
```

The `self.new()` method can be used to define a basic constructor.
`self.need()` checks if a key is set and errors if it isn't (with an optional custom error message). 
`kwargs` that are passed onto a new instance of BaseObj are always accessible via `self.kwargs`
In this example, MyApp needs `name` and `foo` to be passed as kwargs.

```python
class MyApp(BaseObj):
 def new(self):
   self.need("name")
   self.need("foo", msg="please provide a value for foo")

 def body(self):
   self.root.name = self.kwargs.name
   self.root.inner.foo = self.kwargs.foo
   self.root.list = [1, 2, 3]

obj = MyApp(name="myapp", foo="bar")
```

## Setting a skeleton
Defining a large body with Python can be quite hard and repetitive to read and write.
The `self.update_root()` method allows importing a YAML/JSON file to set the skeleton of self.root.

MyApp's skeleton can be set instead like this:
```yaml
#skel.yml
---
name: myapp
inner:
  foo: bar
list:
  - 1
  - 2
  - 3
```

```python
class MyApp(BaseObj):
 def new(self):
   self.need("name")
   self.need("foo", msg="please provide a value for foo")
   self.update_root("path/to/skel.yml")
```

Extending a skeleton'd MyApp is possible just by implementing `self.body()`:

```python
class MyApp(BaseObj):
 def new(self):
   self.need("name")
   self.need("foo", msg="please provide a value for foo")
   self.update_root("path/to/skel.yml")

 def body(self):
   self.set_replicas()
   self.root.metadata.labels = {"app": "mylabel"}

def set_replicas(self):
   self.root.spec.replicas = 5
```

### Inheritance
Python inheritance will work as expected:

```python

class MyOtherApp(MyApp):
  def new(self):
    super().new()  # MyApp's new()
    self.need("size")

def body(self):
   super().body()  #  we want to extend MyApp's body
   self.root.size = self.kwargs.size
   del self.root.list  # get rid of "list"

obj = MyOtherApp(name="otherapp1", foo="bar2", size=3)
```
compiles to:

```yaml
---
name: otherapp1
inner:
  foo: bar2
replicas: 5
size: 3
```

# Components

A component in Kadet is a python module that must implement a `main()` function returning an instance of` BaseObj`. The inventory is also available via the `inventory()` function.

For example, a `tinyapp` component:

```python
# components/tinyapp/__init__.py
from kapitan.inputs.kadet import BaseOBj, inventory
inv = inventory() # returns inventory for target being compiled

class TinyApp(BaseObj):
  def body(self):
    self.root.foo = "bar"
    self.root.replicas = inv.parameters.tinyapp.replicas

def main():
  obj = BaseOb()
  obj.root.deployment = TinyApp() # will compile into deployment.yml
  return obj
```

An inventory class must be created for `tinyapp`: 

```yaml
# inventory/classes/components/tinyapp.yml

parameters:
  tinyapp:
    replicas: 1
  kapitan:
    compile:
    - output_path: manifests
      input_type: kadet
      output_type: yaml
      input_paths:
        - components/tinyapp
```
## Common components
A library in `--search-paths` (which now defaults to `.` and `lib/`) can also be a module that kadet components import. It is loaded using the `load_from_search_paths()`:

```python
kubelib = load_from_search_paths("kubelib") # lib/kubelib/__init__.py

def main():
  obj = BaseObj()
  obj.root.example_app_deployment = kubelib.Deployment(name="example-app")
  return obj
```


kapitan/inputs/kadet.py
```diff
@@ -0,0 +1,245 @@
...
+    loads and executes python module with module_name from search paths
+    returns module
+    """
+    for path in search_paths:
+        try:
+            _path = os.path.join(path, module_name)
+            mod, spec = module_from_path(_path, check_name=module_name)
+            spec.loader.exec_module(mod)
+            return mod
+        except ModuleNotFoundError:
```
uberspot: Why not `except (ModuleNotFoundError, FileNotFoundError): pass` here since you handle both anyway in the raise below? Tiny bit cleaner.

kapitan/inputs/kadet.py
```diff
@@ -0,0 +1,245 @@
...
+    @classmethod
+    def from_dict(cls, dict_value):
+        """
+        returns a BaseObj initialise with dict_value
+        """
+        bobj = cls()
+        bobj.root = Dict(dict_value)
+        return bobj
+
+    def update_root_yaml(self, file_path):
```
uberspot: Another nitpick but could this function and update_root_json be consolidated to one? The only difference is the yaml.safe_load or json.load line. Maybe with an extra argument, or infer the type from the filename?

tests/test_kadet.py
```diff
@@ -0,0 +1,162 @@
...
+    def new(self):
+        self.need('name', 'Need a name string')
+        self.need('size', 'Need a size int')
+
+    def body(self):
+        self.root.name = self.kwargs.name
+        self.root.size = self.kwargs.size
+        self.root.firstkey = 1
+        self.root.nested.firstkey = 2
+        self.root['tradicional_key'] = 3
```
uberspot: typo in traditional  ?

tests/test_kadet.py
```diff
@@ -0,0 +1,162 @@
...
+        self.root.withdict = {'A': 'dict'}
+        self.root.withbaseobj_init_as = BaseObj.from_dict({'init': 'as'})
+        bobj = BaseObj()
+        bobj.root.inside = "BaseObj"
+        self.root.withbaseobj = bobj
```
uberspot: Some of the vars here under self.root are oneword, some are oneWord and some are oneword_and_dashes. 
Would it be more readable if we followed one convention? Something like `self.root.i_am_inside` used below seems nice. :)

tests/test_kadet.py
```diff
@@ -0,0 +1,162 @@
...
+        self.assertEqual(output, desired_output)
+
+    def test_to_dict(self):
+        kobj = KadetTestObj(name='testObj', size=5)
+        output = kobj.to_dict()
+        desired_output = {
+            "name": "testObj",
+            "size": 5,
+            "firstkey": 1,
+            "tradicional_key": 3,
```
uberspot: Typo here and below.

uberspot:
I'd add the text you added in the PR to a separate Kadet.md and link to it from the main README.md. Or maybe move that and the entire `Modes of operation` section to a separate file(?). That could be a different PR, but it would make the Readme easier to digest.

kapitan/inputs/kadet.py
ramaro: good point. Fixed

Issue #70, opened 2018-04-11
Basic linter functionality


adrianchifor:
`kapitan lint` should check for things like:

- Variable definition in target/class but no usage or reference.
- Duplicate variables in the same target.
- Re-definition of the same variable with the same value between targets and classes.

uberspot:
Closing this since first stage is now done. Further linter checks can be added in other tickets :)

PR #193, opened 2019-01-15
 #70: Add some basic linter functionality

uberspot:
Partial work for #70:

Adding checks for 
- Classes unused in the inventory
- Orphan secrets files not used but in the secrets path

Also minor consolidation for the "list_all_paths" logic which was duplicated in 3-4 places and a better message for `kapitan compile` when it fails to load inventories in malformed yaml files.


README.md
```diff
@@ -570,6 +574,8 @@ parameters:
     address: https://localhost:8200
 ```
 
+Use `kapitan lint -h` to checkup on your inventory/secrets.
```
gburiola: why does it need `-h` after `lint` ?

uberspot: It doesn't, it's just pointing people to the lint --help to see all available options there. It felt weird copy pasting the options in the readme as well. I can add them if you think it's needed.

kapitan/lint.py
```diff
@@ -0,0 +1,103 @@
...
+                        secrets_paths.discard(secret_path)
+
+    status = len(secrets_paths) > 0
+    if status:
+        logger.info("No usage found for the following {} secrets files:\n{}".format(len(secrets_paths), pformat(secrets_paths)))
+
+    return status
+
+
+def lint_unused_classes(inventory_path):
```
gburiola: you need to have a special case for `init` classes.

For example, a class called `environment.dev` can refer to a file called `inventory/classes/environment/dev.yml` or `inventory/classes/environment/dev/init.yml`

More info here:
https://reclass.pantsfullofunix.net/operations.html

```
Classes files may reside in subdirectories, which act as namespaces. 
For instance, a class ssh.server will result in the class definition to be read from ssh/server.yml. 
Specifying just ssh will cause the class data to be read from ssh/init.yml or ssh.yml. 
Note, however, that only one of those two may be present.
```

README.md
gburiola: I just found the sentence a bit confusing.

it could be either:

```
Use `kapitan lint -h` to view all the lint options
```

or

```
Use `kapitan lint` to checkup on your inventory/secrets.
```

kapitan/lint.py
uberspot: Nice spot, I didn't know about this. :+1:

kapitan/lint.py
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python3.6
```
adrianchifor: python3

kapitan/lint.py
```diff
@@ -0,0 +1,113 @@
...
+import sys
+from pprint import pformat
+
+from kapitan.utils import list_all_paths
+from kapitan.errors import KapitanError
+
+logger = logging.getLogger(__name__)
+
+
+def start_lint(fail_on_warning, skip_class_checks, inventory_path, search_secrets, secrets_path, compiled_path):
```
ramaro: add docstring

kapitan/lint.py
```diff
@@ -0,0 +1,113 @@
...
+        status_secrets = lint_orphan_secrets(compiled_path, secrets_path)
+
+    status = status_secrets + status_class_checks
+    if fail_on_warning and status > 0:
+        sys.exit(1)
+
+    return status
+
+
+def lint_orphan_secrets(compiled_path, secrets_path):
```
ramaro: add docstring

kapitan/lint.py
```diff
@@ -0,0 +1,113 @@
...
+from kapitan.utils import list_all_paths
+from kapitan.errors import KapitanError
+
+logger = logging.getLogger(__name__)
+
+
+def start_lint(fail_on_warning, skip_class_checks, inventory_path, search_secrets, secrets_path, compiled_path):
+    if skip_class_checks and not search_secrets:
+        logger.info("Nothing to check. Remove --skip-class-checks or add --search-secrets to lint secrets")
+        sys.exit(0)
```
ramaro: shouldn't this be exit(1) instead?

kapitan/lint.py
```diff
@@ -0,0 +1,113 @@
...
+                        secrets_paths.discard(secret_path)
+
+    status = len(secrets_paths) > 0
+    if status:
+        logger.info("No usage found for the following {} secrets files:\n{}\n".format(len(secrets_paths), pformat(secrets_paths)))
+
+    return status
+
+
+def lint_unused_classes(inventory_path):
```
ramaro: add docstring

kapitan/utils.py
```diff
@@ -353,6 +351,12 @@ def get_entropy(s):
     return round(entropy, 2)
 
 
+def list_all_paths(folder):
```
ramaro: add docstring

kapitan/lint.py
```diff
@@ -0,0 +1,113 @@
...
+    logger.debug("Collected # of paths: {}".format(len(class_paths)))
+    logger.debug("Checking if all classes are declared in " + classes_dir)
+
+    for path in list_all_paths(inventory_path):
+        if os.path.isfile(path):
+            with open(path, "r") as compiled_file:
+                file_contents = compiled_file.read()
+                for class_path in list(class_paths):
+                    exists = class_path in file_contents
+                    if class_path.endswith(".init"):
```
ramaro: can you add a comment to clarify what's happening here?

kapitan/lint.py
```diff
@@ -0,0 +1,113 @@
...
+                        exists = (class_path[:-5] in file_contents) or (exists)
+
+                    if exists:
+                        class_paths.discard(class_path)
+
+    status = len(class_paths) > 0
```
ramaro: I find it a bit inconsistent that status is sometimes and int and other times a boolean

uberspot: Noted, I'll rename in `start_lint`.

PR #204, opened 2019-01-29
Fix failing docker build, add docker build to make test and update

uberspot:
update google client dependency


Dockerfile
```diff
@@ -7,7 +7,6 @@ COPY requirements.txt ./
 
 RUN apk add --no-cache --virtual build-dependencies g++ make musl-dev && \
     apk add --no-cache libstdc++ gnupg yaml-dev libffi-dev openssl-dev && \
-    pip install --upgrade --no-cache-dir pip && \
```
adrianchifor: Might be worth adding a note or todo to re-add this once pip fixed the assertion bug

uberspot: Added

Issue #217, opened 2019-02-26
Add `lib` to directory created with init command


caioaao:
**Describe the bug/feature**
The directory created with `init` doesn't contain `.kapitan` file and `lib` directory mentioned in 
the [typical folder structure part of the guide](https://github.com/deepmind/kapitan/tree/c70d20ced6b00053134252e8bf78f7e0645fd7cf#typical-folder-structure)

**To Reproduce**
Run `kapitan init --directory <directory-name>`

**Expected behavior**
Have a directory ready for kapitan development in `<directory-name>`. This means having `lib/kube.libjsonnet` and `lib/kapitan.libjsonnet` inside `<directory-name>`.

uberspot:
`lib/kapitan.libjsonnet` is already included in the kapitan installation so you don't have to include it. 
We avoided adding `lib/kube.libjsonnet` since that is kubernetes specific and kapitan can be used in other contexts as well so it would be 'noisy' to always include it in new directories. Also .kapitan is not needed in new directories unless you want to lock the version of kapitan to a specific release.
A better alternative might be to update the typical folder structure part of the guide to be more generic? 

cc @ramaro @adrianchifor  ?

caioaao:
Yeah, I misunderstood the usage of `lib/kapitan.libjsonnet` and `.kapitan` files. But about `kube.libjsonnet`, I do agree copying it to projects that won't be about kubernetes might be noisy, but the what are the alternatives? I don't think relying on copy/paste from the examples directory is a good idea or experience. Implementing a "dependency manager" for kapitan is not a good idea as well. This looks like a tool more related to jsonnet than kapitan, but while it doesn't exist, it could be part of the initial kapitan bundle. The init will always be noisy for some people, anyway, but should cover the most common use cases.

adrianchifor:
The `kube.libjsonnet` we have is a base example that needs improvement, but I believe [bitnami](https://github.com/bitnami-labs/kube-libsonnet) has a more up-to-date version. 

We are experimenting with [kadet](https://github.com/deepmind/kapitan/pull/190) right now to replace usages of jsonnet as we are slowly hitting limits with its implementation and complexity/readability. There's a plan to revamp the docs and examples in the following months but considering it's a 20% project for us, it might take a while :)

https://github.com/deepmind/kapitan/issues/132 might also be relevant and we now support https://github.com/jsonnet-bundler/jsonnet-bundler to include jsonnet libs from external sources.

caioaao:
[...]

PR #218, opened 2019-02-26
add lib dir on init

caioaao:
Fixes issue #217 

## Proposed Changes

- Add `kube.libjsonnet` to `lib` directory used for initializing a directory with the command `kapitan init`


uberspot:
Archiving for now since #217 answers why we didn't include it :+1:

Issue #241, opened 2019-03-15
Writing ref secrets from the command line fails


ramaro:
**Writing ref secrets from the command line fails**

Currently `ref:` secret types are not supported in the command line.

**To Reproduce**
```
$ kapitan secrets --write ref:my_password -f <(echo -n password123)

Invalid token: ref:my_password. Try using gpg/gkms/awskms:ref:my_password
```

**Expected behavior**
Create ref type secret

ramaro:
fixed in #242

PR #242, opened 2019-03-15
Add support for writing ref type secrets in command line

ramaro:
Fixes issue #241 

Adds support for creating ref type secrets via command line:
```
$ kapitan secrets --write ref:my_password -f <(echo -n password123)
```


uberspot:
It might be good to update the readme as well.

Issue #203, opened 2019-01-29
Warn user when mismatch between target mentioned in yml and the name of the target yml file


uberspot:
**Describe the bug/feature**
When the target mentioned in a yml file is different than the name of the target yml file itself you might get compilation errors like ` Jsonnet error: failed to compile ... `

**To Reproduce**

1. Go to a yaml under inventory/targets
2. In one of the target yml files: change the value of `parameters.target_name to something different than the name of the actual yaml file.
3. Try to compile.

**What happens**

Errors like ` Jsonnet error: failed to compile ... `, compilation fails without fully indicating why.
 
**Expected behavior**
An error/warning is printed to easily identify issue.

yoshi-1224:
Hi,

I went through the codebase and found that `load_target_inventory` function in `targets.py` is a good place to do the check, since we have `targets_list` variable and `inventory_path`. Can I proceed with this idea?

Also, should we throw `InventoryError` (defined in `errors.py`) for such a mismatch, or create another custom error class?

uberspot:
Feel free to proceed. :) I think InventoryError with a custom error message that explains the problem and which file caused it would be helpful enough.

yoshi-1224:
Thanks for the quick reply! I'll be making a PR shortly for this issue.

gburiola:
Instead of doing a check, why not getting rid of the need to define `target_name` inside the yaml file and dynamically creating the variable ?

Issue #147, opened 2018-09-11
Improve error messages for missing mandatory kapitan inventory keys


ramaro:
For an invalid target like the one below:

```
parameters:
  target_name: mynamespace
  namespace: ${target_name}
  kapitan:
    compile:
        - output_path: pre-deploy
          input_type: jsonnet
          output_type: yaml
          input_paths:
            - components/namespace/main.jsonnet
```

Compilation fails with a KeyError, not suggesting that `vars:` needs to be set:

```
Traceback (most recent call last):
  File "/kapitan/kapitan/targets.py", line 84, in compile_targets
    [p.get() for p in pool.imap_unordered(worker, target_objs) if p]
  File "/kapitan/kapitan/targets.py", line 84, in <listcomp>
    [p.get() for p in pool.imap_unordered(worker, target_objs) if p]
  File "/usr/local/lib/python3.6/multiprocessing/pool.py", line 735, in next
    raise value
KeyError: 'vars'
```

A valid target needs at least `vars:target`, as below:

```
parameters:
  target_name: mynamespace
  namespace: ${target_name}
  kapitan:
    vars:
      target: ${target_name}
    compile:
        - output_path: pre-deploy
          input_type: jsonnet
          output_type: yaml
          input_paths:
            - components/namespace/main.jsonnet
```

PR #243, opened 2019-03-16
Validate target name matches target yml file name

yoshi-1224:
Fixes issue #203, #147 

Before compilation, validates whether each target name has a corresponding target yml file in inventory. Target name is derived from `vars.target`. Should this parameter not be set, the following error is given:

```
Target missing: target "<target_filename>" is missing parameters.kapitan.vars.target
This parameter should be set to the target name
```

If the target name does not match the file name, the following error message is provided:
```
Target "<target_name>" is missing the corresponding yml file in <abs_inventory_path>/targets
Target name should match the name of the target yml file in inventory
```

### DIFF
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -360,10 +360,15 @@ def valid_target_obj(target_obj):
 
 
 def validate_matching_target_name(target_filename, target_obj, inventory_path):
-    """if target does not have a corresponding yaml file in *inventory_path*,
-    throws *InventoryError*
+    """Throws *InventoryError* if parameters.kapitan.vars.target is not set,
+    or target does not have a corresponding yaml file in *inventory_path*
     """
-    target_name = target_obj["vars"]["target"]
+    try:
+        target_name = target_obj["vars"]["target"]
+    except KeyError:
+        error_message = "Target missing: target \"{}\" is missing parameters.kapitan.vars.target\n" \
+                        "This parameter should be set to the target name"
+        raise InventoryError(error_message.format(target_filename))
 
     if target_filename != target_name:
         target_path = os.path.join(os.path.abspath(inventory_path), "targets")
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -269,7 +269,8 @@ def save_inv_cache(compile_path, targets):
 
 
 def load_target_inventory(inventory_path, targets):
-    """retuns a list of target objects from the inventory"""
+    """returns a list of target objects from the inventory.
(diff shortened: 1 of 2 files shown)


yoshi-1224:
I just signed the CLA too!

ramaro:
@yoshi-1224 thanks for this! Can you please add a test?

yoshi-1224:
I've implemented the test by temporarily renaming the target yaml file different to the target name to make sure the code captures the naming mismatch error. Have also created unit test for KeyError problem described in #147 

However, I don't actually know if we need `vars.target` parameter as this comment https://github.com/deepmind/kapitan/issues/203#issuecomment-473706413 points out, so if we end up deciding that we don't, I'll close this PR!

kapitan/targets.py
```diff
@@ -269,7 +269,8 @@ def save_inv_cache(compile_path, targets):
 
 
 def load_target_inventory(inventory_path, targets):
-    """retuns a list of target objects from the inventory"""
+    """returns a list of target objects from the inventory.
```
adrianchifor: Just nitpicking - to keep comments consistent, for one-liners we close the `"""` on the same line.

tests/test_compile.py
```diff
@@ -46,6 +49,41 @@ def test_compile_not_enough_args(self):
...
+                sys.argv = ["kapitan", "compile"]
+
+                try:
+                    main()
+                finally:
+                    # correct the filename again, even if assertion fails
+                    if os.path.exists(unmatched_filename):
+                        os.rename(src=unmatched_filename, dst=correct_filename)
+        error_message = "Target \"minikube-es\" is missing the corresponding yml file"
+        self.assertTrue(' '.join(cm.output).find(error_message) != -1)
```
uberspot: Maybe it's more future proof to check if the output contains an "InventoryError" and "is missing the corresponding yml file" so we avoid hardcoding more names in tests and have to manually change them in the future (?). The rest looks good. :)

kapitan/targets.py
yoshi-1224: @adrianchifor thanks for the review! Will be fixing it shortly

tests/test_compile.py
yoshi-1224: @uberspot Thanks for the review! 
I removed the target name from the string like you said to make it more generic. As for "InventoryError", it is not actually logged, and we don't seem to be able to capture the error thrown in stdout/error during unittest, so I will pass on it.

Issue #240, opened 2019-03-14
Use python fire for all cli commands


uberspot:
Some minor refactoring in cli.py to use https://github.com/google/python-fire will allow for much easier expansion in functionality later on with less code.

pranavgupta1234:
Hi @uberspot, I am planning to work on this issue. I was checking into this python-fire guide [https://github.com/google/python-fire/blob/master/docs/guide.md](url) and although there are different possible implementations, one of them can be, as we have 7 different subparsers, so we can make a function for each one of them with corresponding arguments included in it which can then be handled by fire. One other way can be to make class for each subparser. What are your thoughts on this ?

uberspot:
Sounds like a class for each subparser is an overkill. 
It might be simpler to follow their example like here:
``` if __name__ == '__main__':
  fire.Fire({
      'add': add,
      'multiply': multiply,
  }) ```
And import functions in cli.py that implement each part. 
Some more refactoring might be necessary in this process.

**Comment by pranavgupta1234** [comment]
We can shift each subparser code block inside our current if-else tree into separate functions along with necessary arguments and then just fire using a dict. Should I proceed with this implementation ?

**Comment by uberspot** [comment]
If you wish yes. We can continue comments on the PR. :+1:

**Comment by pranavgupta1234** [comment]
Hey @uberspot, I wanted to discuss on some issues before creating a PR. Like:
1) Fire uses arguments from function to create flags so we can use only one flag (either longhand or shorthand notation)
2) Features like, choices=('yaml', 'json') provided by argparse, such checks now have to be done manually inside function and appropriate error has to be raised.
3) Check for nargs also inside function
I am also using python fire for the first time and if I am missing some features please let me know. One example for lint subparser looks like:
```
def lint(fail_on_warning=from_dot_kapitan('lint', 'fail-on-warning', False), 
         skip_class_checks=from_dot_kapitan('lint', 'skip-class-checks', False),
         search_secrets=from_dot_kapitan('lint', 'search-secrets', False),
         secrets_path=from_dot_kapitan('lint', 'secrets-path', './secrets'),
         compiled_path=from_dot_kapitan('lint', 'compiled-path', './compiled'),
         inventory_path=from_dot_kapitan('lint', 'inventory-path', './inventory')):
    '''
    linter for inventory and secrets

    Args:
        fail_on_warning: bool
            exit with failure code if warnings exist, default is False
        skip_class_checks: bool
            skip checking for unused classes, default is False
[...]

PR #253, opened 2019-03-29
Integrate pythonfire

pranavgupta1234:
Fixes issue #240 

## Proposed Changes
  -Uses python fire for all cli commands
  -Some flags are updated due to constraints from python fire
  -tests updated according to new cli
For more info refer to issue #https://github.com/google/python-fire/issues/169

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,12 +7,12 @@ test:
 	@echo ----- Testing build of docker image -----
 	docker build . --no-cache -t kapitan
 	@echo ----- Testing run of docker image -----
-	docker run -ti --rm kapitan --help
+	docker run -ti --rm kapitan -- --help
 	docker run -ti --rm kapitan lint
 	@echo ----- Testing build of docker ci image -----
 	docker build . --no-cache -t kapitan-ci -f Dockerfile.ci
 	@echo ----- Testing run of docker ci image -----
-	docker run -ti --rm kapitan-ci kapitan --help
+	docker run -ti --rm kapitan-ci kapitan -- --help
 	docker run -ti --rm kapitan-ci kapitan lint
 
 .PHONY: test_coverage
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -24,6 +24,7 @@ import json
 import logging
 import os
 import sys
+import fire
 
 import yaml
 from kapitan.errors import KapitanError, RefHashMismatchError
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -7,12 +7,12 @@ test:
 	@echo ----- Testing build of docker image -----
 	docker build . --no-cache -t kapitan
 	@echo ----- Testing run of docker image -----
-	docker run -ti --rm kapitan --help
+	docker run -ti --rm kapitan -- --help
```
adrianchifor: Doesn't work

Makefile
```diff
@@ -7,12 +7,12 @@ test:
...
 	docker run -ti --rm kapitan lint
 	@echo ----- Testing build of docker ci image -----
 	docker build . --no-cache -t kapitan-ci -f Dockerfile.ci
 	@echo ----- Testing run of docker ci image -----
-	docker run -ti --rm kapitan-ci kapitan --help
+	docker run -ti --rm kapitan-ci kapitan -- --help
```
adrianchifor: Doesn't work either

kapitan/utils.py
```diff
@@ -267,7 +281,7 @@ def deep_get(dictionary, keys, previousKey=None):
     return value
 
 
-def searchvar(flat_var, inventory_path, pretty_print):
+def searchvars(flat_var, inventory_path, pretty_print):
```
adrianchifor: Why rename?

tests/test_compile.py
```diff
@@ -34,7 +34,7 @@ def setUp(self):
         os.chdir(os.getcwd() + '/examples/kubernetes/')
 
     def test_compile(self):
-        sys.argv = ["kapitan", "compile", "-c"]
+        sys.argv = ["kapitan", "compile", "--cache"]
```
adrianchifor: This doesn't change anything

tests/test_cli.py
```diff
@@ -277,9 +277,9 @@ def test_cli_searchvar(self):
 
     def test_cli_inventory(self):
         """
-        run $ kapitan inventory -t minikube-es -F -p cluster
+        run $ kapitan inventory --target-name minikube-es --flat --pattern cluster
```
adrianchifor: Why are you expanding the arguments?

tests/test_cli.py
```diff
@@ -85,7 +85,7 @@ def test_cli_secret_write_reveal_gpg(self):
 
     def test_cli_secret_base64_write_reveal_gpg(self):
         """
-        run $ kapitan secrets --write gpg:test_secretb64 --base64
+        run $ kapitan secrets --write gpg:test_secretb64 --b64e
```
adrianchifor: b64e is not a valid flag

kapitan/utils.py
```diff
@@ -52,6 +52,20 @@ def fatal_error(message):
     sys.exit(1)
 
 
+def parse_arg_delimiter(val, delimiter, is_a_path=False):
```
adrianchifor: Where are you using this function?

kapitan/cli.py
```diff
@@ -38,250 +39,268 @@
...
+
+        Args:
+            directory: str
+                set path, in which to generate the project skeleton, assumes directory already exists. default is "./"
+        '''
+        initialise_skeleton(directory)
+
+    def secrets(self, write=None, update=None, update_targets=from_dot_kapitan('secrets', 'update-targets', False), 
+                validate_targets=from_dot_kapitan('secrets', 'validate-targets', False),
+                b64e=from_dot_kapitan('secrets', 'base64', False),
```
adrianchifor: Do not remove or rename existing flags as it breaks backwards compatibility

kapitan/cli.py
```diff
@@ -38,250 +39,268 @@
...
-    elif hasattr(args, 'quiet') and args.quiet:
-        logging.basicConfig(level=logging.CRITICAL, format="%(message)s")
-    else:
         logging.basicConfig(level=logging.INFO, format="%(message)s")
+        
+        try:
+            cmd = sys.argv[1]
+        except IndexError:
+            print("Usage Options:")
+            print("kapitan -- --help")
```
uberspot: What do the "--" do?

pranavgupta1234: We must add an isolated -- argument in order to have arguments treated as Flags rather than be consumed by Python Fire. All arguments to a Fire CLI after the final standalone -- argument are treated as Flags.
Please check this https://github.com/google/python-fire/blob/master/docs/using-cli.md

kapitan/cli.py
pranavgupta1234: Sorry for this. Due to conflict of name of library and flag I changed the flag. It is fixed as I have now used an alias for library while importing.

kapitan/utils.py
pranavgupta1234: In order to use fire I have to create a wrapper around the original imported searchvar. I will use an alias for it while importing to remove name conflict.

tests/test_compile.py
pranavgupta1234: python-fire renders cli on basis of argument names supplied in function signature and so earlier only --cache used to work (as it was supplied in function argument) but later (refer to https://github.com/google/python-fire/issues/108 and https://github.com/google/python-fire/pull/141 ) the shorthand notation ability is indeed added into fire. But the released version on pypi lacks that ability which is soon going to be updated so we can have the shorthand test cases updated later. Also note that capital flags(like -F) are still not allowed and Fire raises error and tell the user in case one uses shorthand notation for two flags whose first letter is same and in that case one has to explicitly write the whole name to avoid conflict.

tests/test_cli.py
pranavgupta1234: Just for the same reason mentioned above. I have checked in the current version they only support shorthand notation for bool types but to maintain uniformity I updated all flags.

kapitan/cli.py
adrianchifor: This is not a viable solution as it significantly changes the way we use the CLI

tests/test_cli.py
adrianchifor: We can't remove the short notation as we and other companies mostly use that.

kapitan/cli.py
pranavgupta1234: I agree, but logically fire needs some way to distinguish between two types of flags, because ``` $ python test.py --help``` could mean you are passing argument to some function(say) fired through python-fire.

tests/test_compile.py
adrianchifor: Ok, can you open an issue on python-fire for them to update the release and we change things after.

tests/test_cli.py
```diff
@@ -61,7 +61,7 @@ def test_cli_secret_write_reveal_gpg(self):
             fp.write(test_secret_content)
 
         sys.argv = ["kapitan", "secrets", "--write", "gpg:test_secret",
-                    "-f", test_secret_file,
+                    "--file", test_secret_file,
```
adrianchifor: We can't remove the test for the short notation as most people rely on this. Let's add an additional test that covers the long notation. If tests break, that's a side effect of the python-fire limitations or the implementation.

tests/test_cli.py
pranavgupta1234: I agree as well and shorthand notation is indeed mostly used.
Once the python-fire's new version is released then even with shorthand notation searchvar, inventory, init and eval will have no problem as each of them have unique first letter in their arguments.
compile, lint and secrets have some conflicts which would be pointed out by fire if someone tries that.

adrianchifor:
Can you resolve the conflict and comments/conversations so it's easier to review. Where are we with pythonfire fixing the things we outlined?

pranavgupta1234:
Some of the desired features are already present and one important feature(required by us as well) which utilizes helptext is under development in python-fire(due to which pypi is not updated yet, @dbieber is trying to include this feature before release). It will probably be completed by end of month and pypi will be up with new release. The code will require some modifications (will do that :) ) which will eventually resolve the current conflicts.
Issue for new release of fire https://github.com/google/python-fire/issues/173

pranavgupta1234:
The newer version of fire has been released. Will try to do the necessary fixes in upcoming couple of days.

pranavgupta1234:
@adrianchifor @uberspot closing this if favour of https://github.com/deepmind/kapitan/pull/377

PR #263, opened 2019-04-05
Fixes #259: Add parseYaml function to jsonnet

### LINKED ISSUE #259: parseYaml function in jsonnet
Kubecfg has a parseYaml function in kubecfg.libsonnet (implemented through a native callback, so I can't just import kubefcg.libsonnet in my kapitan jsonnet files). It would be great to get the same functionality in kapitan.libsonnet.

harsh-98:
Fixes issue #259 

## Proposed Changes

  -
  -
  -


kapitan/resources.py
```diff
@@ -99,6 +102,19 @@ def jinja2_render_file(search_paths, name, ctx):
 
     raise IOError("jinja2 failed to render, could not find file: {}".format(_full_path))
 
+def yaml_to_json(search_paths, name):
+    """return content of file in json"""
+    for path in search_paths:
```
uberspot: We probably don't need to search multiple paths. Just one file path would be fine for this function in jsonnet.

kapitan/resources.py
```diff
@@ -99,6 +102,19 @@ def jinja2_render_file(search_paths, name, ctx):
...
+        _full_path = os.path.join(path, name)
+        logger.debug("yaml_to_json trying file %s", _full_path)
+        if os.path.exists(_full_path):
```
uberspot: It would also be a good idea to check if the file ends with yml or yaml here to avoid unnecessary exceptions.

kapitan/resources.py
```diff
@@ -99,6 +102,19 @@ def jinja2_render_file(search_paths, name, ctx):
...
+            print(_full_path)
+            logger.debug("yaml_to_json found file at %s", _full_path)
+            try:
+                with open(_full_path) as f:
+                    return json.dumps(yaml.safe_load(f.read()))
```
uberspot: This will basically return the first result and immediately exit the for loop. So multiple search paths won't be meaningful here.

uberspot:
Hello :) Can you also update the README with the new function and add a test for it in tests?

kapitan/resources.py
harsh-98: For testing it, I have to create a sample yaml file. Where should I create it?

harsh-98:
For testing it, I have to create a sample yaml file. Where should I create it? Also readme doesn't contain information about other native functions, should I write about them too?

kapitan/resources.py
ramaro: search_paths is necessary to keep it consistent with the other functions

kapitan/resources.py
ramaro: I disagree, this seems correct as it needs to search through because that's how jsonnet search paths work. We do the same on other file based functions - you can't guarantee the file will be found in the first path.

kapitan/resources.py
```diff
@@ -99,6 +102,19 @@ def jinja2_render_file(search_paths, name, ctx):
...
+            except Exception as e:
+                raise CompileError("Parse yaml failed to parse {}: {}".format(_full_path, e))
```
ramaro: The default should be to raise IOError() like in https://github.com/deepmind/kapitan/blob/master/kapitan/resources.py#L100

uberspot:
You could put the sample yaml file under a new folder in `tests/test_resources/` :)

kapitan/resources.py
uberspot: Got it, sounds good then. :)

kapitan/resources.py
harsh-98: The error raised should be compile only as whether exist or not is handled in `if` condition. And in `try except` block only parsing is done.

harsh-98:
@uberspot Are any other changes needed?

README.md
```diff
@@ -378,6 +378,16 @@ This will compile all targets to `compiled` folder.
 #### Using the inventory in jsonnet
 
 Accessing the inventory from jsonnet compile types requires you to import `jsonnet/kapitan.libjsonnet`, which includes the native_callback functions glueing reclass to jsonnet (amongst others).
+Available native_callack functions are:
```
uberspot: callback*

README.md
```diff
@@ -378,6 +378,16 @@ This will compile all targets to `compiled` folder.
...
+```
+yaml_to_json - returns json as string of the yaml file specified
```
uberspot: `returns a json string of the specified yaml file`

README.md
```diff
@@ -378,6 +378,16 @@ This will compile all targets to `compiled` folder.
...
+yaml_dump - returns yaml as string from json string
```
uberspot: `returns a string yaml from a json string`

tests/test_jsonnet.py
```diff
@@ -27,6 +29,15 @@ def test_yaml_dump(self):
         yaml = yaml_dump("{\"key\":\"value\"}")
         self.assertEqual(yaml, "key: value\n")
 
+    def test_parse_yaml(self):
+        """
+            search_paths is also provided as partial
```
uberspot: What exactly is this comment explaining? Can you simplify/reword it?

tests/test_resources/test_parse_yaml.yaml
```diff
@@ -0,0 +1,2 @@
+test:
+  parse: yaml
```
uberspot: For consistency, can you also add a yaml array here with some items? To test yaml syntax a bit more strictly.

kapitan/resources.py
uberspot: IOErrors could come up because of a faulty hardware as well or a bunch of other reasons other than the file missing. 
We should still check for this explicitly. It's a bit more defensive as a coding strategy but it helps debug things in the long term.

tests/test_jsonnet.py
harsh-98: It is explaining why yaml_to_json takes two arguments whereas parse_yaml takes one. So, we also have to pass search_paths. simplifying it.

uberspot: Maybe update the codeblock to
```
""" 
   This tests the yaml_to_json function.
   It converts the yaml file in test_resources/ to a json file in the current_pwd and checks if the output is as expected.
"""
```  
Generic codeblocks that clearly explain what a function does in 1-2 sentences help other developers immensely when reading lots of code.

harsh-98:
@uberspot done requested changes.

tests/test_jsonnet.py
```diff
@@ -27,6 +29,15 @@ def test_yaml_dump(self):
...
         self.assertEqual(yaml, "key: value\n")
 
+    def test_parse_yaml(self):
+        """
+            This tests the yaml_to_json function.
+            It converts the yaml file in test_resources/ to a json string
+        """
+        current_pwd = os.path.dirname(__file__)
+        json = yaml_to_json([current_pwd], "test_resources/test_parse_yaml.yaml")
+        self.assertEqual(json, "{\"test\": {\"key\": \"value\", \"array\": [\"ele1\", \"ele2\"]}}")
```
uberspot: Can you also put ` "{\"test\": {\"key\": \"value\", \"array\": [\"ele1\", \"ele2\"]}}"` in an expected_json variable and then pass it to assertEqual? It makes the code self-explanatory and is a good practice.

kapitan/resources.py
```diff
@@ -99,6 +101,20 @@ def jinja2_render_file(search_paths, name, ctx):
...
+        logger.debug("yaml_to_json trying file %s", _full_path)
+        if os.path.exists(_full_path) and ( name.endswith(".yml") or name.endswith(".yaml") ):
+            logger.debug("yaml_to_json found file at %s", _full_path)
+            try:
+                with open(_full_path) as f:
+                    return json.dumps(yaml.safe_load(f.read()))
+            except Exception as e:
+                raise CompileError("Parse yaml failed to parse {}: {}".format(_full_path, e))
+
+    raise IOError("yaml failed to render, could not find file: {}".format(_full_path))
```
uberspot: change to `could not find any yaml input file in: {}` ? So it's clearer.

README.md
```diff
@@ -378,6 +378,16 @@ This will compile all targets to `compiled` folder.
 #### Using the inventory in jsonnet
 
 Accessing the inventory from jsonnet compile types requires you to import `jsonnet/kapitan.libjsonnet`, which includes the native_callback functions glueing reclass to jsonnet (amongst others).
+Available native_callback functions are:
```
adrianchifor: New line here

kapitan/resources.py
```diff
@@ -55,9 +55,11 @@ def resource_callbacks(search_paths):
                           partial(inventory, search_paths)),
             "file_read": (("name",),
                           partial(read_file, search_paths)),
-            "yaml_dump": (("obj",), yaml_dump),
             "sha256_string": (("obj",), sha256_string),
             "gzip_b64": (("obj",), gzip_b64),
+            "yaml_dump": (("obj",), yaml_dump),
+            "yaml_to_json": (("obj",),
```
adrianchifor: `"obj",` should be `"name",`

kapitan/resources.py
```diff
@@ -99,6 +101,20 @@ def jinja2_render_file(search_paths, name, ctx):
 
     raise IOError("jinja2 failed to render, could not find file: {}".format(_full_path))
 
+def yaml_to_json(search_paths, name):
```
ramaro: Rename to `yaml_load`?

kapitan/resources.py
```diff
@@ -55,9 +55,11 @@ def resource_callbacks(search_paths):
                           partial(inventory, search_paths)),
             "file_read": (("name",),
                           partial(read_file, search_paths)),
-            "yaml_dump": (("obj",), yaml_dump),
             "sha256_string": (("obj",), sha256_string),
             "gzip_b64": (("obj",), gzip_b64),
+            "yaml_dump": (("obj",), yaml_dump),
+            "yaml_to_json": (("name",),
```
ramaro: Rename to `yaml_load`?

kapitan/lib/kapitan.libjsonnet
```diff
@@ -1,5 +1,6 @@
 {
   jinja2_template(name, ctx):: std.native("jinja2_render_file")(name, std.toString(ctx)),
+  parse_yaml(name):: std.native("yaml_to_json")(name),
```
ramaro: Rename to `yaml_load`?

kapitan/resources.py
harsh-98: `yaml_to_json` tells us exactly  functionality of this function. Though yaml_load matches other functions in naming style.

uberspot: it's explicitly used for jsonnet anyway so this yaml_load makes sense and matches the style of other functions. :)

ramaro: Let's keep the naming style :)

uberspot:
:+1: Nice!

Issue #215, opened 2019-02-25
Expand on kapitan init


uberspot:
Expand on the work done https://github.com/deepmind/kapitan/pull/213 
and add functionality to:

 - generate parameterizable target files
 - generate parameterizable class files
 - generate parameterizable components (jinja2, kadet, jsonnet)

vaibhavk:
parameterizable here refers to the command-line parameters support?

uberspot:
I think the concept here was to try and make `kapitan init` generate example target/class/component files based on some parameters that a user inputs. 
Something like `kapitan init --target-name <a new target name>` and it would generate the correctly named file for the user. And the generation would handle the `kapitan.compile` parameters, depending if the user wants jinja2, jsonnet or kadet.
This would probably simplify kapitan usage even more.

vaibhavk:
This makes more sense to me now, thanks

MatteoVoges:
Closing this as the PR #265 got closed too. You could just reopen the PR

PR #265, opened 2019-04-06
Init and target-name and compile-input

harsh-98:
Feature: for issue #215 
If target-name is provided then under `inventory/targets`, target yaml files with provided target-names are created else my_target.yml is copied.
If compile-input is provided then `my_component.yml` file only contains compile_obj with matching input_type. And only those resources are copied.  
Command:
`kapitan init --targets dev staging prod --compile-inputs jinja2`

Content of my_components.yml:

Content of staging.yml:

Currently working on writing test for this.


kapitan/inputs/base.py
```diff
@@ -19,6 +19,7 @@
 import yaml
 import json
 
+from shutil import copy
```
adrianchifor: Why are you importing this?

kapitan/cli.py
```diff
@@ -219,6 +219,13 @@ def main():
     init_parser.add_argument('--directory',
                              default=from_dot_kapitan('init', 'directory', '.'),
                              help='set path, in which to generate the project skeleton, assumes directory already exists. default is "./"')
+    init_parser.add_argument('--verbose', '-v', help='set verbose mode',
+                                action='store_true',
+                                default=from_dot_kapitan('init', 'verbose', False))
+    init_parser.add_argument('--target-name', help='provide a target-name ex: name1,name2',
```
adrianchifor: To keep behaviour consistent with other commands, turn this into `--targets/-t` and take an array instead of comma-separated.

kapitan/cli.py
```diff
@@ -219,6 +219,13 @@ def main():
...
+                                default=from_dot_kapitan('init', 'verbose', ""))
```
adrianchifor: `from_dot_kapitan('init', 'targets', []))`

kapitan/cli.py
```diff
@@ -219,6 +219,13 @@ def main():
...
+    init_parser.add_argument('--compile-input', help='provide a target-name ex: jinja2,jsonnet,kadet',
+                                default=from_dot_kapitan('init', 'verbose', ""))
```
adrianchifor: `from_dot_kapitan('init', 'compile-inputs', []))`

kapitan/cli.py
```diff
@@ -219,6 +219,13 @@ def main():
     init_parser.add_argument('--directory',
                              default=from_dot_kapitan('init', 'directory', '.'),
                              help='set path, in which to generate the project skeleton, assumes directory already exists. default is "./"')
+    init_parser.add_argument('--verbose', '-v', help='set verbose mode',
+                                action='store_true',
+                                default=from_dot_kapitan('init', 'verbose', False))
+    init_parser.add_argument('--target-name', help='provide a target-name ex: name1,name2',
+                                default=from_dot_kapitan('init', 'verbose', ""))
+    init_parser.add_argument('--compile-input', help='provide a target-name ex: jinja2,jsonnet,kadet',
```
adrianchifor: Same as above, turn this into `--compile-inputs` and take an array instead of comma-separated.

kapitan/initialiser.py
```diff
@@ -20,27 +20,151 @@
...
-    logger.info("Populated {} with:".format(directory))
-    for dirName, subdirList, fileList in os.walk(directory):
-        logger.info('{}'.format(dirName))
-        for fname in fileList:
-            logger.info('\t {}'.format(fname))
-        # Remove the first entry in the list of sub-directories
-        # if there are any sub-directories present
-        if len(subdirList) > 0:
-            del subdirList[0]
+class initialise_skeleton(object):
```
adrianchifor: Camel case for classes

vaibhavk:
I have been working on this issue, should I stop working on it now?

kapitan/cli.py
harsh-98: It is later converter to array in the initialiser object __init__ function.

kapitan/cli.py
harsh-98: Same as target-name, later converted to array in init function.

kapitan/cli.py
adrianchifor: Yes I've noticed, but the correct behaviour would be to accept it as an array in the first place, see `compile --targets` arg

kapitan/initialiser.py
harsh-98: Yup, will do it . Earlier it was written like this. So, i used it.

adrianchifor: Yes it was a function, not a class :)

kapitan/cli.py
harsh-98: Ok. Understand, ignore my comment on target-name and compile-input.

harsh-98: Ok. Understand, ignore my comment on target-name and compile-input.

harsh-98:
@adrianchifor 
@vaibhavk Pushed a PR on the same issue this PR is on. What should we do?

harsh-98:
@adrianchifor @uberspot currently I and vaibhavk are working on the same issue. It doesn't make sense. What can be done about that? It needs urgent attention.

adrianchifor:
@harsh-98 This is not urgent, relax. It's very common in open-source for multiple people to work on the same issues, it helps raise different points and implementations. In this particular case, I see both implementations are very similar so I commented on the other PR https://github.com/deepmind/kapitan/pull/269#issuecomment-480908495. Chat with @vaibhavk and try to combine your ideas :)

vaibhavk:
@harsh-98 I'll be working with you on this issue from now.
You could have commented on issue itself before making PR in the first place since I was working on it.

kapitan/initialiser.py
```diff
@@ -20,27 +20,150 @@
...
-            del subdirList[0]
+class Initialiser(object):
+    def __init__(self,directory,targets,compile_inputs):
+        """ Initialises a directory with a recommended skeleton structure
+        Args:
+            directory (string): path which to initialise, directory is assumed to exist
+        """
+        self.target_template_name = 'my_target.yml'
+        self.component_template_name = 'my_component.yml'
+        self.target_dir = "inventory/targets"
```
vaibhavk: @harsh-98 wouldn't this cause issue with windows system?

harsh-98: Python has a hack, https://docs.python.org/3/library/os.html#os.altsep. It internally, changes forwardslash to backward slash when used in windows. Ref: [here](https://medium.com/@ageitgey/python-3-quick-tip-the-easy-way-to-deal-with-file-paths-on-windows-mac-and-linux-11a072b58d5f).

harsh-98:
@uberspot review plz.

kapitan/initialiser.py
```diff
@@ -20,27 +20,150 @@
...
+
+
+
+    def dump_yaml(self,obj,path):
+        with open(path, 'w') as outfile:
+            yaml.dump(obj, outfile, default_flow_style=False)
+
+    def list_new_directory(self):
+        """
+            list all files in new location
```
uberspot: in self.directory* And rename list_new_directory to just `list_directory` if that's the actual function.

kapitan/initialiser.py
```diff
@@ -20,27 +20,150 @@
...
+                self.copy_path_list += compile_obj['input_paths']
+                compile_objs.append(compile_obj)
+        component_file['parameters']['kapitan']['compile'] = compile_objs
+
+        path = self.get_new_component_path()
+        self.dump_yaml(component_file, path)
+
+    def copy_compile_components(self):
+        """
+            copy for files present in copy_path_list to new location
```
uberspot: This comment needs rewording as well. Also the function is only used in one place. Does it need to be a function?

kapitan/initialiser.py
```diff
@@ -20,27 +20,150 @@
...
+        self.targets = targets
+        self.compile_inputs = compile_inputs
+        current_pwd = os.path.dirname(__file__)
+        self.templates_directory = os.path.join(current_pwd, 'inputs/templates')
+
+        self.copy_path_list = []
+
+
+
+    # copy_tree(templates_directory, directory)
```
uberspot: Remove stale comments from code.

kapitan/initialiser.py
```diff
@@ -20,27 +20,150 @@
...
-            logger.info('\t {}'.format(fname))
-        # Remove the first entry in the list of sub-directories
-        # if there are any sub-directories present
-        if len(subdirList) > 0:
-            del subdirList[0]
+class Initialiser(object):
+    def __init__(self,directory,targets,compile_inputs):
+        """ Initialises a directory with a recommended skeleton structure
+        Args:
+            directory (string): path which to initialise, directory is assumed to exist
```
uberspot: The other args aren't explained here as well.

kapitan/initialiser.py
```diff
@@ -20,27 +20,150 @@
...
+    def dump_yaml(self,obj,path):
+        with open(path, 'w') as outfile:
+            yaml.dump(obj, outfile, default_flow_style=False)
+
+    def list_new_directory(self):
+        """
+            list all files in new location
+        """
+        logger.info("Populated {} with:".format(self.directory))
+        for dirName, _, fileList in os.walk(self.directory):
```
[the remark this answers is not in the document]
uberspot: It lists both directories and files. Then you just `os.path.isfile` and use only those instead. I think I already did that in some other part of kapitan codebase.

uberspot:
> When these arguments are passed, some files are deleted, added and their content also changes. For testing this, I tried to use dir_hash but it failed as tmpdir has different permission attributes.
> So, should I read each file and compare the content?

Why not just fix the permissions of your init output before using `dir_hash` in the test itself?

harsh-98:
@ramaro review plz.

kapitan/cli.py
```diff
@@ -219,6 +219,15 @@ def main():
     init_parser.add_argument('--directory',
                              default=from_dot_kapitan('init', 'directory', '.'),
                              help='set path, in which to generate the project skeleton, assumes directory already exists. default is "./"')
+    init_parser.add_argument('--verbose', '-v', help='set verbose mode',
+                                action='store_true',
+                                default=from_dot_kapitan('init', 'verbose', False))
+    init_parser.add_argument('--targets', '-t', help='specify target names to generate,  e.g. name1 name2',
+                                type=str, nargs='+',
+                                default=from_dot_kapitan('init', 'verbose', []))
+    init_parser.add_argument('--compile-inputs', '-i', help='specify type of templates to generate , e.g. jinja2 jsonnet kadet',
```
ramaro: why not name this `--input-types` to keep thing consistent?

kapitan/initialiser.py
```diff
@@ -20,27 +20,143 @@
...
+    def __init__(self,directory,targets,compile_inputs):
+        """ Initialises a directory with a recommended skeleton structure
+        Args:
+            directory (string): path which to initialise, directory is assumed to exist
+            targets (list): target files to be created under targets/
+            compile_inputs (list): input type of templates to be generated
+        """
+        self.target_template_name = 'my_target.yml'
+        self.component_template_name = 'my_component.yml'
+        self.target_dir = "inventory/targets"
```
ramaro: the inventory path should not be hardcoded. It should take the value of `--inventory-path`.

adrianchifor:
Can you resolve the comments/conversations so we know what's left to review?

harsh-98:
@adrianchifor All the comments and conversation has been resolved.

harsh-98:
I have rebased this PR with master branch.

Issue #151, opened 2018-10-17
Simpler Example


manhof:
It may be useful to do a simpler example for kaptian, something like a helloworld/hostname example that builds for multiple namespaces with a cluster.

uberspot:
Just for reference, there are some examples also available here https://github.com/ramaro/kapitan-examples/ in case it helps anybody.

harsh-98:
@uberspot I have use https://github.com/ramaro/kapitan-examples/. I would like to work on this.

PR #266, opened 2019-04-06
chore(kapitan/examples): Add easy nginx example

harsh-98:
Fixes issue #151


examples/nginx/components/app/nginx.jsonnet
```diff
@@ -0,0 +1,54 @@
+local kap = import "lib/kapitan.libjsonnet";
+local inv = kap.inventory();
+
+local myContainers = [
```
adrianchifor: You can use `kube.libjsonnet` from the k8s example for `Container`, `Deployment` and `Service`.

examples/nginx/components/app/nginx.jsonnet
```diff
@@ -0,0 +1,54 @@
...
+                labels: { app: "my-app" },
+            },
+            spec: {
+                containers: myContainers,
+            },
+        },
+    },
+};
+
+local svc = {
```
adrianchifor: Can you also split `Deployment` and `Service` into separate files

examples/nginx/components/app/main.jsonnet
```diff
@@ -0,0 +1,18 @@
...
+        labels: {
+            name: inv.parameters.namespace,
+        },
+        name: inv.parameters.namespace,
+        namespace: inv.parameters.namespace,
+    },
+};
+
+{
+    namespace: n,
```
adrianchifor: Why do you need both `main.jsonnet` and `nginx.jsonnet` ?

harsh-98:
@adrianchifor Now, there is only one jsonnet file. Since, it has only 30 lines of code, so no need for split. Thanks to kube.libjsonnet suggested by you. Kjsonnet is awesome.  :+1:

uberspot:
Could this instead be added to examples/kubernetes? 
A lot of things are overlapping with that example, so it would be nicer to instead **enhance** the kubernetes example we already have with a new component (nginx) and 2 new targets (prod-nginx, dev-nginx) for example (?). That way it also get's tested with the existing tests that build examples/kubernetes as well. This ensures our example is always up to date and actually functional.  

Remember that when you update the examples/kubernetes you need to update the test files for it as well.

harsh-98:
@uberspot where are prod-nginx, dev-nginx files. I am not able to find them.

uberspot:
> @uberspot where are prod-nginx, dev-nginx files. I am not able to find them.

Those are the files you'd add with this PR.

harsh-98:
@ubrspot there is already a component named nginx, it is in kadet. So, I can change existing one to nginx-kadet and add new one as nginx-jsonnet.

harsh-98:
@adrianchifor should I squash all commits or make a new commit? As examples/nginx folder will be deleted.

uberspot:
Squashing is cleaner. :+1:

harsh-98:
@uberspot done the changes

harsh-98:
@uberspot review this, plz.

harsh-98:
@uberspot regarding #303 users are looking for simple examples, can you review?

examples/nginx/components/app/nginx.jsonnet
uberspot: This is still left todo. Just move the `local svc = {... }` to separate files and import them in nginx.jsonnet with `local deployment = import "deployment.jsonnet";` and then the 
```
{
    "app-service": svc,
    "app-deployment": deployment,
}
``` 
stays the same :)

harsh-98: service has field which requires `deployment.spec.template`, so if I move svc to other file it will create circular dependency instead i can move deployment to other file and import it in main.jsonnet

adrianchifor: It's not going to create a circular dependency. Just split deployment and service into separate files, and import in main. You can also import the deployment in the service and reference it, jsonnet can figure it out.

harsh-98:
@adrianchifor you can also review this PR. I have done all the asked changes.

harsh-98:
@uberspot Can u look at this PR? I think it is ready to be merged.

Issue #237, opened 2019-03-08
Allow defining custom jinja2 filters in templates


ramaro:
**Allow defining custom jinja2 filters in templates**

We now have more jinja2 filters (thanks to @adrianchifor work on #234). However I'd like support for providing a custom python file with more filters, e.g.:

```
$ kapitan compile --jinja2-filters=lib/jinja2_filters.py
```
The default behaviour should be for Kapitan not to error should the default jinja2 filters file not exist. 

Should it exist, Kapitan will first import those filters into jinja2's internal env settings.

(A better default file name other than one in the example should be discussed)

pranavgupta1234:
Hey @ramaro , I am planning to work on this issue. The execution plan can be to use [importlib](https://docs.python.org/2/library/importlib.html#importlib.import_module) to dynamically import the custom filter module specified by user and then define one more function "load_jinja2_filters_from_file" to import the functions and suppy them in env. We can extract the name for function but then the exposed name of filter will be same as the original name.

PR #267, opened 2019-04-06
add support for specifying custom jinja2 filters

pranavgupta1234:
Fixes issue #237 

## Proposed Changes
  - Updates CLI to allow --filter-paths
  - adds a function load_jinja2_filters_from_file to load filters dynamically(one or more)
  - updates jinja rendering functions inside utils.py to add filter_paths
  - test added

Usage is :
```
$ kapitan compile --filter-paths=some/custom/filter.py
```
can pass multiple as well
```
$ kapitan compile --filter-paths some/filter/one.py some/filter/two.py
```


examples/kubernetes/lib/custom_filter.py
```diff
@@ -0,0 +1,4 @@
+import base64
```
uberspot: can we rename the file and function to `custom_jinja2_filter` ? so that it's clearer for people new to kapitan. :)

examples/kubernetes/lib/custom_filter.py
```diff
@@ -0,0 +1,4 @@
...
+
+def custom_filter(string):
+    return base64.b64encode(string.encode("UTF-8")).decode("UTF-8")
```
uberspot: Adding 1-2 comment lines explaining what this example filter does, would be a good practice.

kapitan/cli.py
```diff
@@ -72,6 +72,10 @@ def main():
                                 default=from_dot_kapitan('compile', 'search-paths', ['.', 'lib']),
                                 metavar='JPATH',
                                 help='set search paths, default is ["."]')
+    compile_parser.add_argument('--filter-paths', '-J2F', type=str, nargs='+',
```
uberspot: Can we rename the flag to `--jinja2-filter-paths` since these are jinja2 specific?

kapitan/inputs/jinja2_filters.py
```diff
@@ -47,6 +51,19 @@ def load_jinja2_filters(env):
     env.filters['shuffle'] = randomize_list
 
 
+def load_jinja2_filters_from_file(env, filter_paths):
+    for path in filter_paths:
```
uberspot: Add comment block explaining what the function does and what output it returns.

kapitan/inputs/jinja2_filters.py
```diff
@@ -47,6 +51,19 @@ def load_jinja2_filters(env):
...
+        try:
+            custom_filter_spec = util.spec_from_file_location(path.split('/')[-1].split('.')[0], path)
```
uberspot: Move `path.split('/')[-1].split('.')[0]` to a separate line and assign to a var to make this more readable for other people.

kapitan/inputs/jinja2_filters.py
```diff
@@ -47,6 +51,19 @@ def load_jinja2_filters(env):
...
+            custom_filter_module = util.module_from_spec(custom_filter_spec)
+            custom_filter_spec.loader.exec_module(custom_filter_module)
+            for f in dir(custom_filter_module):
```
uberspot: Try to avoid using single letter variables unless the meaning is implied. It's more intuitive/self-explanatory when you read `file` rather than `f`. :)

kapitan/inputs/jinja2_filters.py
```diff
@@ -47,6 +51,19 @@ def load_jinja2_filters(env):
...
+                if isinstance(getattr(custom_filter_module, f), types.FunctionType):
```
uberspot: Move `getattr(custom_filter_module, f)` to a separate line to improve readability.

kapitan/inputs/jinja2_filters.py
```diff
@@ -47,6 +51,19 @@ def load_jinja2_filters(env):
     env.filters['shuffle'] = randomize_list
 
 
+def load_jinja2_filters_from_file(env, filter_paths):
```
uberspot: In this function, you also need to check whether filter_paths are actual paths/directories or files with os.path.isdir/isfile etc. If you get a `filter_paths` that points to an empty directory, maybe the function should at least warn the user without stopping execution? To prevent mistakes and make debugging easier.

tests/test_jinja2.py
```diff
@@ -158,3 +159,16 @@ def test_inventory_global_context(self):
...
+class Jinja2ExternalFilterTest(unittest.TestCase):
+    def test_custom_filter_jinja2(self):
+        with tempfile.NamedTemporaryFile() as f:
+            f.write("{{ inventory.parameters.cluster.name | custom_filter }}".encode("UTF-8"))
+            cluster_name = "minikube"
+            target_name = "minikube-es"
+            inv = inventory(["examples/kubernetes"], target_name)
+            context = {"inventory": inv}
+            f.seek(0)
+            self.assertEqual(render_jinja2_file(f.name, context, ["./examples/kubernetes/lib/custom_filter.py"]), base64_encode(cluster_name))
```
uberspot: Move the `render_jinja2_file(f.name, context, ["./examples/kubernetes/lib/custom_filter.py"]),` and `base64_encode(cluster_name)` to separate lines and add to variables like `expected_output` and `actual_output`. 
Try keeping your code in <100 characters of width. It improved readability for people using smaller screens. :+1:

kapitan/inputs/jinja2_filters.py
harsh-98: os.path.basename(path).split('.')[0]

kapitan/inputs/jinja2_filters.py
pranavgupta1234: In case if user provides path to a directory then should we walk in it and load the all the modules present there ?

uberspot: We can start by taking into account only modules in the provided directory. This is easy to expand later on as well if needed. It should be a one line change + test change.  :+1:

uberspot:
Just a small note, to also update the README with this new functionality. 1-2 sentences should be enough.

kapitan/inputs/jinja2_filters.py
pranavgupta1234: As nargs='+' so he may provide mixture of file path and directories. I am thinking to separate the logic and flatten all file paths first and then load them.

pranavgupta1234: Hey @uberspot , I have refactored the logic, have a look.

kapitan/inputs/jinja2_filters.py
```diff
@@ -51,17 +51,40 @@ def load_jinja2_filters(env):
...
-                    env.filters[f] = getattr(custom_filter_module, f)
-        except Exception as e:
-            raise IOError("jinja2 failed to render, could not find filter at {}: {}".format(path, e))
+def load_module_from_path(env, path):
+    """
+    loads a python module from provided path and adds it to jinja2 environment
+    filter name is same as that of function
+    """
+    try:
+        custom_filter_spec = util.spec_from_file_location(os.path.basename(path).split('.')[0],path)
```
uberspot: Move the `os.path.basename(path).split('.')[0]` to a line above it to improve readability.

kapitan/inputs/jinja2_filters.py
```diff
@@ -51,17 +51,40 @@ def load_jinja2_filters(env):
...
+    iterates over all paths provided in jinja2_filter_paths list (files and directories) 
+    and calls load_module_from_path for each path
+    """
+    for path in jinja2_filter_paths:
+        if os.path.isfile(path) and path.endswith(".py"):
+            load_module_from_path(env, path)
+        elif os.path.isdir(path):
+            for filename in os.listdir(path):
+                subpath = os.path.join(path, filename)
+                if os.path.isfile(subpath) and subpath.endswith(".py"):
```
uberspot: If you restructure your code a bit, you can avoid repeating the lines:
```  if os.path.isfile(subpath) and subpath.endswith(".py"):
                    load_module_from_path(env, subpath) ``` 
and just do a check for paths before that and add all the subpaths in one list if it's a directory.  
Then for all of them iterate and load module.

Comment (pranavgupta1234) [reply kapitan/inputs/jinja2_filters.py]: Hey @uberspot ,  I have updated the code. Moreover I also removed checking by isfile() because it returns False for files which do not exist. But I guess we want to throw error in case the name or path of file is wrong rather than failing at that stage when it would require those filters. What are  your thoughts on this ?

Comment (uberspot) [reply kapitan/inputs/jinja2_filters.py]: If the file doesn't exist and the flag is pointing to an individual file instead of a dir, then it should fail with an exception (and you can do that in an if in the start of the function). If a dir is provided it should try and load all modules (aka anything that isfile and endswith ".py") but it shouldn't throw exceptions only if there aren't any files in the directory at all. You need an extra check in your code for that.

Comment (pranavgupta1234) [reply kapitan/inputs/jinja2_filters.py]: What is the expected behaviour in case user passes two directories, one empty and other containing valid modules ?

**Comment by ramaro** [comment]
This is good, but it deviates from the behaviour defined in the issue.
A default value for a filters filename should not be empty and be set to the suggested value in the issue.
The behaviour needs to be: if the default filters filename does not exist, do not error. If a different filename is set and the file does not exist, then error.
This is important to keep the compilation simple by default.

**Comment by pranavgupta1234** [comment]
@ramaro , should we allow directories then ? Also are we packaging some default filter file ?

**Comment by ramaro** [comment]
@pranavgupta1234 I don't think supporting directories is necessary. The user can extend their filters code to whatever is needed. Let's keep it to a minimum to start with.

**Comment by ramaro** [comment]
And no, we don't need to package a default filter file.

**Comment by pranavgupta1234** [comment]
So should I modify it to support only a single file, which as mentioned is jinja2_filters.py and in case this file does not exist then do not raise any error but if any different filename is provided which does not exist then raise error ?

**Comment by ramaro** [comment]
@pranavgupta1234 Yes, that's it.

**Comment by pranavgupta1234** [comment]
Just one more doubt that the default location is supposed to be lib/jinja2_filters.py, I mean inside lib folder.

**Comment by ramaro** [comment]
That's fine, it doesn't need to exist by default :)

**Comment by pranavgupta1234** [comment]
Hey @ramaro , now by default kapitan will load custom filters from lib/jinja2_filters.py if present, and user doesn't even have to specify it.
In case he specifies default path then it also tries to load, and no exception in this case.
If ill file path, then exception will be thrown.
Have a look.

File: kapitan/inputs/jinja2_filters.py (hunk 11 above)
Comment (uberspot) [reply kapitan/inputs/jinja2_filters.py]: A warning for the empty dir (to warn the user about it but not stop execution) and the valid modules loaded from the other dir. At least that's what I would expect.

File: kapitan/utils.py [hunk 12]
```diff
@@ -111,7 +111,7 @@ def sha256_string(string):
     return sha256(string.encode("UTF-8")).hexdigest()
 
 
-def render_jinja2_file(name, context, jinja2_filter_paths=[]):
+def render_jinja2_file(name, context, jinja2_filters=os.path.join('lib', 'jinja2_filters.py')):
```
uberspot: Can you put `os.path.join('lib', 'jinja2_filters.py')` in a readonly var in `kapitan/inputs/jinja2.py`  and then reuse it everywhere instead of defining a hardcoded value in 2 places? Otherwise changing this in the future might be a headache.

pranavgupta1234: It works fine in cli.py but I think it is creating cyclic import problem between utils and jinja2 modules.

pranavgupta1234: @uberspot can you reproduce this behaviour ?

uberspot: If the cyclic dependency doesn't occure in jinja2_filters.py then you can put it there. It's mostly to avoid duplication in obscure places. And add a comment too detailing what that path is exactly. :+1:

pranavgupta1234: Updated !

kapitan/inputs/jinja2_filters.py
```diff
@@ -21,13 +21,21 @@
...
 from six import string_types
 from random import Random, shuffle
+from importlib import util
 
 from kapitan.errors import CompileError
 from kapitan import utils
 
+logger = logging.getLogger(__name__)
+
+"""default path from where user defined custom filters are read"""
```
ramaro: this should be a comment, not a string

pranavgupta1234: sorry, fixed. :)

PR #270, opened 2019-04-10
Fixes #260: Add support for url imports in jsonnet

### LINKED ISSUE #260: Remote URLs in jsonnet imports
kubecfg has implemented the ability to do import and importstr on URLs, making it possible to import libraries and JSON/YAML (if combined with #259) from remote sources.
This would make it possible to import an upstream YAML and output a modified version, giving Kustomize a run for it's money with the power that jsonnet brings to the table. See something like [kustomize-libsonnet](https://github.com/anguslees/kustomize-libsonnet) for examples of how this could be used (only currently possible with kubecfg).

harsh-98:
Fixes issue #260 

## Proposed Changes

-  [x] import URL's content in jsonnet
-  [x] add check for IOError when no jsonnet file is found
-  [x] write test case for jsonnet's url import feature

### DIFF
--- a/kapitan/resources.py
+++ b/kapitan/resources.py
@@ -26,6 +26,7 @@ import gzip
 import base64
 import reclass
 import reclass.core
+import requests
 from reclass.errors import ReclassException, NotFoundError
 import yaml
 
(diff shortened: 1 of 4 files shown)


ramaro:
I'm not totally this is a good approach as it introduces external dependencies which will make a compilation non deterministic.

Ideally we'd support this via dependency management.

Issue #191, opened 2019-01-15
Create kubelib in python for Kadet


uberspot:

michiboo:
Hi I would like to work on this issues! Could you point to an example that are already done or explain it in more detail?

vaibhavk:
@ademariag can I work on this?

ademariag:
Hi, I'm not involved with this process. Please ask @uberspot

adrianchifor:
@vaibhavk @michiboo See https://github.com/deepmind/kapitan/pull/190 for more details. We can start off with creating the equivalent of `kube.libjsonnet` in kadet (`kube.py`). 

@uberspot Can we split this issue into 2 for k8s and terraform libs. For grafana there's already https://github.com/weaveworks/grafanalib.

uberspot:
Updated description and created #271

vaibhavk:
@adrianchifor while working on it, I'm not able to implement [Endpoint](https://github.com/deepmind/kapitan/blob/d9f2d82d8d286eb2792377321e6a40e3befa897d/examples/kubernetes/lib/kube.libjsonnet#L58), it has hidden functions inside it, Ip and Port, to implement this possible solutions I can think of are a single functions with kwargs, or a class based implementation, any suggestions?

uberspot:
A class in the same file might be preferred but it's up to you as long as code is readable and doesn't repeat too much

PR #273, opened 2019-04-19
[WIP] implement kube.py(kube.libjsonnet) for Kadet

vaibhavk:
Fixes issue #191 

## Proposed Changes

- [ ] Implement kube.libjsonnet for Kadet
- [ ] Tests

### DIFF
--- a/examples/kubernetes/lib/kube.py
+++ b/examples/kubernetes/lib/kube.py
@@ -1,0 +1,79 @@
+from addict import Dict
+from kapitan.inputs import kadet
+
+inv = kadet.inventory()
+
+# Returns array of [key, value] pairs from given object.  Does not include hidden fields.
+def objectItems(o):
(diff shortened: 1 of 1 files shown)


vaibhavk:
@adrianchifor can you review the class based function? I doubt it'll give the same functionality as [kube.libjsonnet's Endpoints](https://github.com/deepmind/kapitan/blob/d9f2d82d8d286eb2792377321e6a40e3befa897d/examples/kubernetes/lib/kube.libjsonnet#L58)

ramaro:
@vaibhavk  I don't think it's worth or even the right approach to directly translate/port kube.libjsonnet into python. We're looking for a pythonic abstraction here!

vaibhavk:
@ramaro can I ask for a simple example of abstraction example, so I can continue with that approach?

Issue #131, opened 2018-07-25
Secret files containing multiple values / allowing GPG encrypted target files even?


frittentheke:
When many secrets are used, gpg will be running / invoked over and over again to decrypt each an every file containing just a little string.
It is certainly inefficient to run GPG for every such little string and it clutters the classes files to add gpg and the secret path for each time.

1) What about allowing referencing sub-variables in secret files like

```
  mysql:
    storage: 10G
    storage_class: standard
    image: mysql:latest
    users:
      root:
         password: ${gpg:allmysecretsinonefile:mysql:password}

```

-- This would allow decrypting and then caching a whole file containing multiple values once and allow for much faster compiles when many secrets are used.

2) Much easier / less error-prone:

Allow using whole gpg encrypted target files which are simply decrypted prior to rendering the templates.
this could be as easy as adding ".gpg" to your target yaml file by convention. Inside the classes those variables should be addressable just like any other variable, making the fact that this is actually sourced from an encrypted file transparent.

yoshi-1224:
Hi @uberspot! Could you share with us what the current situation for this issue is, i.e. is this enhancement going to be implemented, in the middle of implementation etc.?

If I understood correctly, I personally prefer the first idea, because it more clearly separates the secrets from the rest, and is more intuitive as is an extension of what we already have. It also seems more backwards compatible.

uberspot:
Agreed, the first idea sounds more elegant as well. As far as I know no one is looking into this yet. :)

yoshi-1224:
Ok! Let me draft up an implementation for this

yoshi-1224:
Hi @uberspot 

Sorry for a late start on this. After looking into this, I have a few things to confirm
1. we currently have the structure `${<ref_type>:<file_path>:<hash_digest>|<func_str>}` where hash_digest and func_str is optional. So we need a delimiter other than `:` to express sub-variables in a file (otherwise it will be considered hash digest). Do you have any suggestions as to which character to use?
2. my idea for implementation is that whenever we see tokens with sub-variables, we save it into RefController in a dictionary (preserving the hierarchical structure). For example, if we have `file1:var1:x`, `file1:var2:z` and `file1:var1:y` we will get a structure like:
```
file1
- var1:
  x
  y
- var2:
  z
```
Then, RefController will expose a `flush` method which, at the end of per-target compilation, will be called to encrypt and write the secret file with all the variables inside. This way, it should not introduce change to the way secret files without sub-variables are currently saved.

[...]

PR #282, opened 2019-04-29
Implement secret sub-variables

yoshi-1224:
Fixes issue #131 

This PR is just for the feedback on the implementation. No tests are added yet either (temporarily commented out the assertion for compile hash equivalence).

## Proposed Changes
  - notation: `?{<ref_type>:<file_path>@<yaml_path>}` (only implemented for gpg Ref so far). 
  - Users are expected to run `kapitan secrets --write ...` where the secret file is a yaml file.

## Implementation idea
  - set `yaml_path` attribute to Ref (sub-class) instances. Use this to check if the secret is a sub-variable, and if so, after decryption in reveal function, index into the yaml_path and return the value
  - decryption is cached using `lru_cache`. But it is inefficient because the parameter is the data itself, rather than ref file path. We should ideally use ref file path as the parameter since it is a just short string.
  - hash of the secrets after `kapitan compile` is the hash of the entire file content. e.g. if we have `?{gpg:file1@var1}` and `?{gpg:file1@var2}` then the hash for the two secrets are identical. This should be secure still based on collision resistance.
  - during `kapitan compile`, it doesn't check if the sub-vars are actually present in the encrypted yaml file. Should we implement this as well?


uberspot:
> * during `kapitan compile`, it doesn't check if the sub-vars are actually present in the encrypted yaml file. Should we implement this as well?

I think yes, we need to make sure that's covered as well otherwise it will fail silently when deployed instead of during compiling and will make debugging more painful for users. 

>decryption is cached using lru_cache. But it is inefficient because the parameter is the data itself, rather than ref file path. We should ideally use ref file path as the parameter since it is a just short string.
:+1: Makes sense as an approach.

examples/kubernetes/secrets/subvars.yml
```diff
@@ -0,0 +1,5 @@
+var1:
```
uberspot: I assume this file is here just to demonstrate how the yml looks like? We shouldn't need this extra file in the final user flow right?

kapitan/refs/secrets/gpg.py
```diff
@@ -143,11 +153,24 @@ def _encrypt(self, data, fingerprints, encode_base64):
...
 
+    @classmethod
+    @lru_cache(maxsize=256)
+    def decrypt_with_cache(cls, data):
+        return gpg_obj().decrypt(data, **GPG_KWARGS)
+
     def _decrypt(self, data):
         """decrypt data"""
-        dec = gpg_obj().decrypt(data, **GPG_KWARGS)
+        dec = self.decrypt_with_cache(data)
```
uberspot: Maybe it's better to do the caching in the place that calls `_decrypt` already?

examples/kubernetes/secrets/subvars.yml
yoshi-1224: Yes, this is used for demonstration and testing only. The user would store this somewhere else.

kapitan/refs/secrets/gpg.py
yoshi-1224: You’re right. I was afraid caching takes place per instance method but it does so as a class. Will correct this

examples/kubernetes/secrets/subvars.yml
uberspot: Ok, it would also be good to add updated examples for these so that people don't have to figure out multiline string syntax etc to write secrets.

ramaro:
> * during `kapitan compile`, it doesn't check if the sub-vars are actually present in the encrypted yaml file. Should we implement this as well?

@yoshi-1224 how is this going to work? We can't decrypt on compile - I'm not sure implementing this will be possible.

uberspot:
What about storing the structure of the subvars in the secret itself(like we store the encoding)? Just the keys of the yaml, decrypted names same as they would be in individual files. You could store that as a base64'ed json. It is a bit divergent again from the simpler secret concept we had but it's an alternative if decrypting during compile proves too challenging.

ramaro:
@yoshi-1224 @uberspot I'm not totally sure this should be implemented for each ref type. Ideally this should be abstracted away from gpg/akms/gkms and offered as given feature, regardless of the type. It doesn't make sense to me for each of these to have to implement it again and again.

How about exploring your ideas instead in the `Revealer` (methods `reveal_obj()` and `_reveal_replace_match()`) and `RevealedObj` classes?

ramaro:
> What about storing the structure of the subvars in the secret itself(like we store the encoding)? Just the keys of the yaml, decrypted names same as they would be in individual files. You could store that as a base64'ed json. It is a bit divergent again from the simpler secret concept we had but it's an alternative if decrypting during compile proves too challenging.

I don't know if it will be necessary to add extra complexity for this feature. I like to think that subvars are simply a _reveal filter_ feature for yaml files encrypted into secrets. The workflow to "validate" these on compile would have to be by passing the `--reveal` flag on compile - it should fail if revealing subvars errors out. Compile per se should not validate subvars (unless we think of a less contrived way).

yoshi-1224:
@ramaro @uberspot 
Thanks for the feedback!
> How about exploring your ideas instead in the Revealer (methods reveal_obj() and _reveal_replace_match()) and RevealedObj classes?

Thanks for pointing this out. I believe this is a much better idea. I just implemented it in the most recent commit (sorry the rebasing really messed things up. We should continue this discussion on a new PR). Please have a look at `refs/base.py`.

Regarding the validation, it necessitates decryption which may not be desirable. Also, as @ramaro pointed out, reveal will automatically fail if the sub-variables do not exist so I agree it is not strictly necessary to validate.

yoshi-1224:
hi @ramaro @uberspot
The CLA issue resolved as I force-pushed, so let me continue the discussion here. I believe this new approach allows us to make caching & decryption generic, so please provide some reviews!

TODO:
- write test for subvars
- update examples

kapitan/refs/base.py
```diff
@@ -233,10 +237,38 @@ def _reveal_dir(self, dirname):
     def _reveal_replace_match(self, match_obj):
         """returns decrypted value from tag in match_obj"""
         tag, _, _ = match_obj.groups()
-        ref = self.ref_controller[tag]
-        logger.debug("Ref: revealing: %s", tag)
+        m = re.search(REF_TOKEN_SUBVAR_PATTERN, tag)
+        if m is None:
+            # no subvar pattern is found
```
uberspot: It would be good to log a debug message here for info.

kapitan/refs/base.py
```diff
@@ -233,10 +237,38 @@ def _reveal_dir(self, dirname):
...
+            return self._reveal_tag_without_subvar(tag)
+
+        else:
+            subvar_path = m.groups()
+            if len(subvar_path) != 1:
+                assert False
```
uberspot: Same here and in the branch below

tests/test_compile.py
```diff
@@ -39,7 +39,7 @@ def test_compile(self):
         os.remove('./compiled/.kapitan_cache')
         compiled_dir_hash = directory_hash(os.getcwd() + '/compiled')
         test_compiled_dir_hash = directory_hash(os.getcwd() + '/../../tests/test_kubernetes_compiled')
-        self.assertEqual(compiled_dir_hash, test_compiled_dir_hash)
+        # self.assertEqual(compiled_dir_hash, test_compiled_dir_hash)
```
uberspot: We should fix this test as well before merging. :)

examples/kubernetes/secrets/subvars.yml
ramaro: I think a comment here will suffice

ramaro:
@yoshi-1224 I think this is good to go! Let's just add the comment in subvar.yml and we can merge.

examples/kubernetes/secrets/subvars.yml
uberspot: Added comment, resolving this :)

uberspot:
Merged, nice work! :) I'll do a small rc release to keep this as a tested feature

Issue #272, opened 2019-04-11
Update inventory validation schema to incorporate constraints on output_type


pranavgupta1234:
Currently the schema for validation of inventory structure does not incorporate constraint on output_type, instead they are handled at compilation phase. I think if we incorporate this at schema level then the error can be shown much before and this can make Kapitan more responsive and fast.

I have the modified schema ready for this which will put constraint on output_type to be only yaml or json(in case input_type is jsonnet or kadet) and allow all output_type for jinja2(it doesn't matter if one specifies it in current implemetation as well).

By looking, it seems like current schema is defined on abstraction for validation of basic structure only and not for any specific type of key that's why I am not sure about making this change or not.

pranavgupta1234:
After giving a bit more thought about this issue I think just before invoking worker threads for doing compilation, inside load_target_inventory(...) where we are currently doing two step validation : 
1) valid_target_obj : for validation of basic compile structure
2) validate_matching_target_name
Now we can add one more step which includes validation of our currently supported types or just modify the schema of inside valid_target_obj to put the necessary constraint.

PR #285, opened 2019-05-04
updated validation schema for compile

pranavgupta1234:
Fixes issue https://github.com/deepmind/kapitan/issues/272 
This is just a sample implementation draft for discussion.
## Proposed Changes
  -Updates jsonchema for validation of output_type based on input type before invoking compilation threads.

### DIFF
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -349,6 +349,25 @@ def valid_target_obj(target_obj):
                     },
                     "required": ["input_type", "input_paths", "output_path"],
                     "minItems": 1,
+                    "oneOf": [
+                        {
+                            "properties": {
+                                "input_type": {
+                                    "enum": ["jsonnet", "kadet"] 
+                                },
+                                "output_type" : {
+                                    "enum": ["yaml", "json"]
+                                }
+                            },
+                        },
+                        {
+                            "properties": {
+                                "input_type": {
+                                    "enum": ["jinja2"]
+                                } 
+                            }
+                        }
+                    ],
                 }
             },
         },
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -356,7 +356,7 @@ def valid_target_obj(target_obj):
                                     "enum": ["jsonnet", "kadet"] 
                                 },
                                 "output_type" : {
-                                    "enum": ["yaml", "json"]
+                                    "enum": ["yaml", "json", "plain"]
                                 }
                             },
                         },


pranavgupta1234:
Any updates @ramaro @adrianchifor @gburiola ?

kapitan/targets.py
```diff
@@ -349,6 +349,25 @@ def valid_target_obj(target_obj):
...
+                            },
+                        },
+                        {
+                            "properties": {
+                                "input_type": {
+                                    "enum": ["jinja2"]
+                                } 
+                            }
+                        }
+                    ],
```
gburiola: thanks @pranavgupta1234 

I find it a bit more legible if we add the `enum` fields inside `compile.items.properties`. what do you think ?

Something like this:

```
                "items": {
                    "type": "object",
                    "properties": {
                        "input_paths": {"type": "array"},
                        "input_type": {
                            "type": "string",
                            "enum": ["jsonnet", "kadet", "jinja2"]
                        },
                        "output_path": {"type": "string"},
                        "output_type": {
                            "type": "string",
                            "enum": ["yaml", "json"]
                        },
                    },
                    "required": ["input_type", "input_paths", "output_path"],
                    "minItems": 1,
```

kapitan/targets.py
```diff
@@ -349,6 +349,25 @@ def valid_target_obj(target_obj):
...
                     "required": ["input_type", "input_paths", "output_path"],
                     "minItems": 1,
+                    "oneOf": [
+                        {
+                            "properties": {
+                                "input_type": {
+                                    "enum": ["jsonnet", "kadet"] 
+                                },
+                                "output_type" : {
+                                    "enum": ["yaml", "json"]
```
ramaro: we also now support "plain" output_type for all input_types, so please make sure to include that as well please.

kapitan/targets.py
pranavgupta1234: This is much cleaner and I also think the conditional logic is not required right now. Just a small reminder that the above jsonschema will not pass any arbitrary output file format with jinja2, i.e if I supply output format as xml with jinja2 it will fail while we can ignore this output format and let it pass which also happens currently and this property will break. Let me know your thoughts on this.

pranavgupta1234: The jsonschema I have originally proposed ignores any output_type for jinja2 and accepts json and yaml (will add plain too) for jsonnet and kadet.

pranavgupta1234: any updates @gburiola ?

kapitan/targets.py
pranavgupta1234: Added.

PR #296, opened 2019-05-17
Update resources.py

saikirankv:
reclass is not considering the parameters passed in reclass-config.yml file

Fixes issue #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/kapitan/resources.py
+++ b/kapitan/resources.py
@@ -239,7 +239,7 @@ def inventory_reclass(inventory_path):
                                           reclass_config['classes_uri'],
                                           reclass_config['compose_node_name'])
             class_mappings = reclass_config.get('class_mappings')  # this defaults to None (disabled)
-            _reclass = reclass.core.Core(storage, class_mappings, reclass.settings.Settings())
+            _reclass = reclass.core.Core(storage, class_mappings, reclass.settings.Settings(reclass_config))
 
             cached.inv = _reclass.inventory()
         except ReclassException as e:


saikirankv:
I have signed the CLA

kapitan/resources.py
```diff
@@ -239,7 +239,7 @@ def inventory_reclass(inventory_path):
                                           reclass_config['classes_uri'],
                                           reclass_config['compose_node_name'])
             class_mappings = reclass_config.get('class_mappings')  # this defaults to None (disabled)
-            _reclass = reclass.core.Core(storage, class_mappings, reclass.settings.Settings())
+            _reclass = reclass.core.Core(storage, class_mappings, reclass.settings.Settings(reclass_config))
```
gburiola: hey @saikirankv.

I'm not sure if this is actually needed. 

We keep our reclass config inside `<inventory_path>/reclass-config.yml` and it works fine for us.

Where do you keep your file ?

Please also send the first 5 lines of running `kapitan inventory --verbose`
As you can see below, it does load the file correctly

```
$ kapitan inventory --verbose | head -3
2019-05-17 16:54:43,074 kapitan.resources DEBUG    Using reclass inventory config at: ./inventory/reclass-config.yml
__reclass__:
  timestamp: Fri May 17 16:54:43 2019
```

saikirankv: @gburiola 
`2019-05-22 14:04:49,998 kapitan.resources DEBUG    Using reclass inventory config at: ./inventory/reclass-config.yml`
without this change reclass is consider the default value, its not considering the value present in the config file. I have debugged the issue and came up with this change.

**without these changes**
setting the strict value to False
image

Printing the value in setting.py
image

as you can see in terminal it is considering the default value

image

saikirankv: upon making the above code change,I can see the value is passing to the settings.py file.

Issue #295, opened 2019-05-17
Docker test failing due to urllib version


yoshi-1224:
**Describe the bug/feature**
When I run `make test`, the following error occurs.

```bash
----- Testing run of docker ci image -----
docker run -ti --rm kapitan-ci kapitan --help
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 583, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (urllib3 1.25.2 (/usr/lib/python3.6/site-packages), Requirement.parse('urllib3<1.25,>=1.20; python_version >= "3.4"'), {'botocore'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/kapitan", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3241, in <module>
    @_call_aside
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3225, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3254, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 585, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 598, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 791, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (urllib3 1.25.2 (/usr/lib/python3.6/site-packages), Requirement.parse('urllib3<1.25,>=1.20; python_version >= "3.4"'), {'botocore'})
Makefile:5: recipe for target 'test' failed
make: *** [test] Error 1
```

**To Reproduce**
Run `make test`

[...]

PR #300, opened 2019-05-22
Fix requests version to circumvent urllib3 version conflict

yoshi-1224:
Fixes issue #295 

requests module update from 2.21.0 to 2.22.0 is breaking the test. For the time being, it makes sense to fix it at 2.21.0.


uberspot:
Nice :+1: We should eventually just use boto3 instead of boto.

Issue #91, opened 2018-05-09
Dependency management


adrianchifor:
We could of course use plain git submodules, but what if we only need only one file or folder from an external git repo, http url or even helm chart/templates ?

I would like to introduce the concept of modular dependencies in kapitan, and it would look something like this:

```
parameters:
  kapitan:
    compile:
    ...

    dependencies:
    - type: git
      output_path: lib
      source: git@github.com:grafana/grafonnet-lib.git
      subdir: grafonnet
      version: 0.1
    - type: http
      output_path: lib
      source: https://raw.githubusercontent.com/bitnami-labs/kube-libsonnet/master/kube.libsonnet
      version: 0.1
```

In this scenario, `kapitan compile` will pull `github.com/grafana/grafonnet-lib/grafonnet/*` and `github.com/bitnami-labs/kube-libsonnet/master/kube.libsonnet` into `lib` at the root. Dependencies will be cached and only re-pulled when the **version** changes, keeps things simple and we don't need to implement an extra kapitan subcommand.

Users have full control of what the **source** can be (jsonnet, libsonnet, jinja2 templates, k8s manifests etc.) and where it's placed inside the project (**output_path**), can be per target, class or global.

In terms of modularity, each **type** can be defined as a separate python module with the same name, as in `kapitan/kapitan/dependencies/git.py` will control the functionality of `type: git`; in the example above, **subdir** is a git module specific variable, we can also add branch, release etc.

@ademariag @gburiola @ramaro What do you think?

sbarzowski:
FYI there's an effort to create a standard dependency management for jsonnet, called jsonnet-bundler. I think the authors have thought it out pretty well. I see you want to do deal with much more than just jsonnet, so I'm not sure how relevant it is.

repo: https://github.com/jsonnet-bundler/jsonnet-bundler
doc: https://docs.google.com/document/d/1czRScSvvOiAJaIjwf3CogOULgQxhY9MkiBKOQI1yR14/edit?usp=sharing

adrianchifor:
Yes we've seen that, we can easily create a dependency module specifically for jsonnet (`type: jsonnet`) that uses the standard. The idea is still very fresh, once it's a bit more mature we will take a look :)

adrianchifor:
@sbarzowski Forgot to mention, we merged #133 a while ago, which allows the usage of jsonnet-bundler to pull external dependencies and reference them inside kapitan.

caioaao:
[...]

PR #304, opened 2019-05-27
External dependency manager

yoshi-1224:
Fixes issue #91 

This PR is to track the progress for GSoC as well as to get feedback on the implementation!

#### Proposed Timeline
Git type: by May 31
HTTP[s] type: by June 7


yoshi-1224:
@uberspot @adrianchifor 
Where should I store the cache for dependencies? It shouldn't be in the `compiled` directory as it seems we remove the directory when we run `kapitan compile`. Could it be simply in `.`?

yoshi-1224:
@ramaro Thanks for the reply. I was just thinking the first time user compiles targets with external dependencies (say http file), it will download and save the actual dependency file in the "cache" directory. Next time the user runs compile command, it won't download the same file again but just copy the file from the "cache" directory to the compile/target directory.

uberspot:
Definitely make the cache directory configurable via a flag or the kapitan .config file. For the default location itself you could use whatever system directory is set for cache. $XDG_CACHE_HOME in most distros for example points to the default ~/.cache directory. https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html if that doesn't exist in the system maybe .cache/ in the root kapitan directory is an ok default? Otherwise i think ramaro is also suggesting putting them under components/ ? I'm not sure if that'd cause any other conflicts though.

yoshi-1224:
Hi @uberspot
For the time being I will implement the ability for the user to specify "cache" directory. 
As for the default location, components/ may make sense considering the fact that we will be allowing the user to actually use the downloaded dependencies as inputs to kapitan (e.g. helm inputs).

ramaro:
Why can't the dependency output_path be "cache" here? We can simply download it straight to the output_path and cache won't be necessary. Fetching updates would overwrite with a new version. Am I missing something?

uberspot:
Won't that imply that it will have to download the same dependency multiple times for multiple targets, as the output_path is 'linked' semantically for each target? I might be missing something too, though.

ramaro:
Why make it target specific? The intended use for dependencies will be
jsonnet/kadet/helm components, right?

yoshi-1224:
@ramaro @uberspot 
So the output_path is most likely under `./components` directory (although user can specify), into which we directly save the fetched files. And then depending on the file type (say jsonnet) the user can further specify them in each of the target inventory as `kapitan.compile` items, just as they would do for other files, and compile into target directory?

ramaro:
> @ramaro @uberspot
> So the output_path is most likely under `./components` directory (although user can specify), into which we directly save the fetched files. And then depending on the file type (say jsonnet) the user can further specify them in each of the target inventory as `kapitan.compile` items, just as they would do for other files, and compile into target directory?

Yes! :+1:

kapitan/cli.py
```diff
@@ -276,7 +280,7 @@ def _search_imports(cwd, imp):
         compile_targets(args.inventory_path, search_paths, args.output_path,
                         args.parallelism, args.targets, ref_controller,
                         prune=(args.prune), indent=args.indent, reveal=args.reveal,
-                        cache=args.cache, cache_paths=args.cache_paths,
+                        cache=args.cache, cache_paths=args.cache_paths, dependency_cache_path=args.dependency_cache_path, force_fetch=args.fetch,
```
yoshi-1224: @uberspot @ramaro  caching is still done for git repositories since we have subdir and/or ref attributes that may make the actual directories different among different targets. For example, if we have a git dependency with subdir, we will still want to retain the entire repository for other targets to use other subdirs in the repository. We could of course make this caching temporary and fetch the repository every time the subdir or ref changes. Let me know what you think.

kapitan/targets.py
```diff
@@ -70,11 +74,22 @@ def compile_targets(inventory_path, search_paths, output_path, parallel, targets
...
+
+        fetch = partial(fetch_dependencies, cache_path=kwargs.get('dependency_cache_path', None))
+        target_dependencies_map = get_dependencies_per_target(target_objs)
+        if kwargs.get('force_fetch', False):
+            # force fetch
+            remove_saved_dependencies(target_dependencies_map, kwargs.get('dependency_cache_path', None))
+
+        for target_name, items in target_dependencies_map.items():
+            [p.get() for p in pool.imap_unordered(fetch, items) if p]
+            logger.info("Dependency fetching complete for target {}".format(target_name))
```
yoshi-1224: Fetching is done concurrently among dependencies, target by target. I think it makes sense because two different targets could contain the same (inherited) dependencies, in which case we could be fetching the same thing at the same time unless we first take care of such duplicates. If we do them target-by-target, then the second one onward could simply check if the file already exists inside fetch() function and avoid fetching twice.

uberspot: Have you tested this with `kapitan compile -p 4` for example? With 4 parallel threads? I'm curious how it would behave then when multiple targets re-use a dependency. :)

kapitan/targets.py
```diff
@@ -324,6 +340,46 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, **kwa
...
+        try:
+            dependencies = target_obj["dependencies"]
+            target_name = target_obj["vars"]["target"]
+            mapping[target_name] = []
+            for item in dependencies:
+                dependency_type = item["type"]
+                source_uri = item["source"]
+                output_path = item["output_path"]
+                if dependency_type == "git":
+                    del item["output_path"]
```
uberspot: Is this deleted because for git dependencies you still cache? 
If so, might be good to add a small comment here for it. :+1:

kapitan/targets.py
```diff
@@ -324,6 +340,46 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, **kwa
...
+                    mapping[target_name].append(Git(source_uri, output_path, **item))
+                elif dependency_type in ("http", "https"):
+                    mapping[target_name].append(Http(source_uri, output_path))
+                else:
+                    logger.error("type {} is not supported as dependency".format(dependency_type))
+                    assert False
+
+        except KeyError:
```
uberspot: Add logger.debug() messages in all exceptions to help during debugging :)

kapitan/targets.py
yoshi-1224: @uberspot when multiple targets use the same dependency, then since fetching is done target-by-target sequentially, the set of dependencies for the first target will be fetched and then if there are any duplicates for the remaining targets, then the fetch would be skipped. This is done by checking whether the file at the output_path exists for http types, and for git types, whether the repository exists in cache.

yoshi-1224: The potential issue is when a target uses the same dependency (source) for some reason. Then there could be situations where for example git clone is done simultaneously for the same repo, and one clone process finishes saving to output_path, and the other clone process throws error because the output_path already exists.

kapitan/targets.py
yoshi-1224: @uberspot it is deleted because we have a parameter called output_path in the constructor, while also having output_path as the key in the item dict. When I pass in **item as kwargs then it would throw error for duplicate arguments for output_path. I could just make subdir and ref parameters explicit in the constructor (defaulting to none) as well. Let me know what you think.

kapitan/targets.py
yoshi-1224: Got it! Also, by right this error will not happen because of the json validation schema that limits type to one of http, https or git. Let me know if assert should at all be used.

kapitan/targets.py
uberspot: This is ok, just add a code comment to make it clearer for future reviewers too. :)

kapitan/targets.py
uberspot: You could just thrown an exception instead of `assert False` and in that exception use a good descriptive message like the one you have in `logger.error("type {} is not supported as dependency".format(dependency_type))`

kapitan/targets.py
uberspot: In that case ^ we don't want to error out and stop the compilation. The user can't do anything to solve this ^. So kapitan should just ideally ignore it and assume the dependency is already downloaded. And add logger.debug() messages to make it easy to detect if that happens when debugging. :+1:

kapitan/cli.py
uberspot: sgtm, it's much clearer now as a process.

kapitan/targets.py
yoshi-1224: @uberspot Makes sense! I'll catch the exception and log that message.

kapitan/cli.py
ramaro: Wouldn't fetching the whole repo and dealing with whatever subdir a target might need externally be simpler? That way the "cache" is the downloaded repo itself.
I propose we do something like:
 * The first time a repo is needed for a target (e.g. files no found), download it entirely
 * If another target needs a different subdir of that repo, the user should override it in the inventory for that target.

Or am I failing to understand the need for a dedicated cache dir here?

yoshi-1224: @ramaro thanks for the comment.
I'm just not sure where to save the entire repo in the way that is common across targets such that it can be reused as cache across targets. Say we have a target that has a git dependency with subdir specified, I believe the `output_path` should contain the subdir extracted from the repo. Then where should the entire cloned repo be saved? Or following your approach,

> The first time a repo is needed for a target (e.g. files no found), download it entirely

So in this case, the entire repo is first saved to the `output_path` as given in this target.

> If another target needs a different subdir of that repo, the user should override it in the inventory for that target.

For this target, the `output_path` is different because it wants the subdir to be at `output_path`, which is fine. But then how should we know where the `output_path` to the entire repo is during compilation of this target in order to reuse it as cache?

I could totally be misunderstanding your approach. Please let me know if that's the case!

kapitan/targets.py
ramaro: This makes me think that perhaps compilation shouldn't be downloading dependencies and maybe we should introduce a new kapitan subcommand for deps? (just a thought)

kapitan/targets.py
ramaro: Yep, please add comment! It was also not obvious to me :)

kapitan/targets.py
ramaro: Please raise KapitanError/CompileError or even create a new error from errors.py

kapitan/dependency_manager/git.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+from distutils.dir_util import copy_tree, remove_tree
+
+from kapitan.dependency_manager.base import Dependency
+from kapitan.errors import GitSubdirNotFoundError
+
+logger = logging.getLogger(__name__)
+
+
+class Git(Dependency):
```
ramaro: Please can you add docstrings for all methods

kapitan/dependency_manager/http.py
```diff
@@ -0,0 +1,44 @@
+import logging
+import os
+import requests
+
+from kapitan.dependency_manager.base import Dependency
+
+logger = logging.getLogger(__name__)
+
+
+class Http(Dependency):
```
ramaro: Please can you add docstrings for all methods

tests/test_dependency_manager.py
```diff
@@ -0,0 +1,83 @@
...
+import tempfile
+
+from kapitan.cached import reset_cache
+from kapitan.cli import main
+from kapitan.dependency_manager.base import Dependency
+
+from kapitan.dependency_manager.git import Git
+
+
+class GitDependencyTest(unittest.TestCase):
```
ramaro: Please can you add docstrings for all methods

kapitan/dependency_manager/base.py
```diff
@@ -0,0 +1,30 @@
+import os
+from abc import abstractmethod
+
+
+class Dependency(object):
```
ramaro: Please add docstrings

requirements.txt
```diff
@@ -11,5 +11,7 @@ boto3==1.9.138
 requests==2.21.0
 addict==2.2.1
 yamllint>=1.15.0
+rfc3987==1.3.8
```
ramaro: where is this used?

kapitan/targets.py
```diff
@@ -324,6 +340,46 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, **kwa
...
+            continue
+    return mapping
+
+
+def remove_saved_dependencies(dependency_mapping, cache_path=None):
+    if cache_path is not None:
+        Dependency.set_cache_path(cache_path)
```
ramaro: Is this going to work correctly when compiling in parallel? I'm worried about race conditions

kapitan/targets.py
```diff
@@ -324,6 +340,46 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, **kwa
...
+    for items in dependency_mapping.values():
+        for item in items:
+            item.clear_cache()
+
+
+def fetch_dependencies(item, cache_path=None):
+    if cache_path is not None:
+        Dependency.set_cache_path(cache_path)
```
ramaro: same question ^

kapitan/targets.py
yoshi-1224: @ramaro Possibly we can do both? Allow the user to specify flag for `compile` command that does fetching before compile. Since fetching is not per-target, I'm actually starting to see it as a different process to compile as well.

kapitan/targets.py
yoshi-1224: @ramaro @uberspot 
I'm assuming the user wouldn't specify the same dependency source twice for each target as it doesn't make sense to save the same file in two different output_paths. In that case there wouldn't be any race condition.
Maybe we need to redesign the code for fetching? Since fetched resources are not per-target, it doesn't make sense to follow how `ref`s are written (i.e. Ref object/class for each secret). It is probably easier to loop through the `dependencies` parameter across targets, remove duplicate sources to make the sources unique, fetch and save to tmp, and then copy those into their respective output_path(s) if no file at those path(s) exists. In my opinion, having separate Git/Http instances makes the fetching process rather convoluted.

kapitan/targets.py
ramaro: Isn't that the point of `compile --fetch`?

yoshi-1224: @ramaro ok I see. I originally thought `--fetch` was to force the fresh dependency fetch (to update the existing files when the external resources are updated), where fetching was done automatically prior to compile.

uberspot: That's what i thought as well. It would be simpler for users if fetching was seamless, at least for the first time (?)

requirements.txt
yoshi-1224: @ramaro 
This is used to validate the source against valid rfc protocol in json schema.

yoshi-1224:
@uberspot @ramaro 
I've rewritten the dependency manager without classes since they make it harder to deal with duplicate dependencies. Now, we go through the dependencies parameters across targets, and create a mapping from source_uri to dependency items (items consists of source_uri, output_path etc.). If a file or dir at output_path exists, then that dependency item will be ignored (i.e. won't be fetched). This in a way acts like cache.

We then fetch each of the unique sources into temp dir, and then copy them to their respective output_paths. This step is done in parallel. No race condition should arise either since sources are unique.

One improvement to be made now is to make sure the filename in the temp dir (i.e. before we copy to output_path) is unique. For example, if we have foo.com/file1.txt and bar.com/file1.txt there will be issues since we only store the files as file1.txt. This could be done by not only using the basename(source_uri) but also the domain and path in the stored file (e.g. hash them). 

Please let me know what you think. Thanks!

uberspot:
Nice work. :) Prepending the hash to the filename would work as a solution for now. We can always re-iterate that in another PR.

yoshi-1224:
@uberspot Is there some permission issue, possibly something to do with public key on Travis that prevents kapitan from running `git clone` command successfully? `make test` works fine locally, so I'm guessing this is Travis-specific issue. Maybe the cause is as described [here](https://awolski.com/sudden-permission-denied-publickey-error-in-travis/)?

kapitan/targets.py
uberspot: @yoshi-1224  Should this be marked as Resolved since it was revised in a newer commit in this PR?

uberspot:
> @uberspot Is there some permission issue, possibly something to do with public key on Travis that prevents kapitan from running `git clone` command successfully? `make test` works fine locally, so I'm guessing this is Travis-specific issue. Maybe the cause is as described [here](https://awolski.com/sudden-permission-denied-publickey-error-in-travis/)?

If you look at `scripts/inc_version.sh` we have the env var `GH_TOKEN` defined which is used for authentication to Github. There's no other git keys by default on travis. So I think in your tests you might need to adjust your remote repos to be of the format `https://${GH_TOKEN}@github.com/deepmind/kapitan.git` IF GH_TOKEN exists in the environment (that way local testing doesn't brake). 
That is very specific to kapitan tests so I'd keep the change in those tests only if possible.

yoshi-1224:
Hi @uberspot, 
Thanks for the tip! In the test, I tried to change the git source url as follows:

```python
if 'GH_TOKEN' in os.environ: # for remote tests only
    git_source = 'https://{}@github.com/deepmind/kapitan.git'.format(os.environ.get('GH_TOKEN'))
```
However, this statement did not run, meaning 'GH_TOKEN' is not available as env var? 

Anyway, I changed the git url from ssh to https (i.e. https://github.com/deepmind/kapitan.git) and seems to also work on Travis (and locally). So I guess we can use this for now. Do you think it's necessary to test both ssh and https git sources though?

uberspot:
> Hi @uberspot,
> Thanks for the tip! In the test, I tried to change the git source url as follows:
> 
> ```python
> if 'GH_TOKEN' in os.environ: # for remote tests only
>     git_source = 'https://{}@github.com/deepmind/kapitan.git'.format(os.environ.get('GH_TOKEN'))
> ```
> 
> However, this statement did not run, meaning 'GH_TOKEN' is not available as env var?
> 
> Anyway, I changed the git url from ssh to https (i.e. https://github.com/deepmind/kapitan.git) and seems to also work on Travis (and locally). So I guess we can use this for now. Do you think it's necessary to test both ssh and https git sources though?

I think it's because that env var isn't available to non-trusted builds https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings 
``` 
Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.
``` 
You can still test if this works, by creating a new API Token in your personal github account (make sure to make it read-only, assuming you don't have any private repos because that will be public in the code) and then hardcoding that token in your tests to see if it works while you are in the branch. Then we just remove that and use GH_TOKEN once we verify a token works.

Alternatively we'll have to test this once merged in master which I'd rather avoid.

uberspot:
We can keep both tests for https and git and my gut tells me that GH_TOKEN will be available once we merge on travis. I don't mind having another PR to review the test if it fails worst case. I'll be available to merge/fix if that happens.

yoshi-1224:
@uberspot Thanks for the help. Great! We could for now assume that if https repo source urls work then so would git urls, as that's on GitPython library to handle. I'll keep that if statement so that when we merge into master, git source urls would get tested as well.

kapitan/targets.py
yoshi-1224: Should `fetch` now be a separate command as well, with `--fetch` a flag for `compile`?

kapitan/targets.py
```diff
@@ -70,11 +71,14 @@ def compile_targets(inventory_path, search_paths, output_path, parallel, targets
 
         # append "compiled" to output_path so we can safely overwrite it
         compile_path = os.path.join(output_path, "compiled")
-        worker = partial(compile_target, search_paths=search_paths, compile_path=temp_path, ref_controller=ref_controller,
+        worker = partial(compile_target, search_paths=search_paths, compile_path=temp_path,ref_controller=ref_controller,
```
ramaro: missing space after `temp_path,`

kapitan/cli.py
```diff
@@ -276,7 +278,7 @@ def _search_imports(cwd, imp):
         compile_targets(args.inventory_path, search_paths, args.output_path,
                         args.parallelism, args.targets, ref_controller,
                         prune=(args.prune), indent=args.indent, reveal=args.reveal,
-                        cache=args.cache, cache_paths=args.cache_paths,
+                        cache=args.cache, cache_paths=args.cache_paths, force_fetch=args.fetch,
```
ramaro: where is `force_fetch` implemented?

yoshi-1224: @ramaro Thanks for the review! It's not yet implemented, because I was unsure whether `--fetch` flag indicates force-fetch or simply fetch-then-compile.

yoshi-1224: @ramaro I'm gonna assume `--fetch` as force fetch. I don't think there is much use in compile without fetch because that will fail always.

ramaro: @yoshi-1224 I think it's best to stick to fetch-then-compile for now. In the future we might even think about a `$ kapitan deps --fetch` option for force-fetch only. Thanks!

yoshi-1224: @ramaro I think we posted comments at the same moment!
Ok so is it correct that without `--fetch` flag, `compile` will just run without fetching the dependencies?

ramaro: That's it :+1:

Dockerfile
```diff
@@ -7,6 +7,7 @@ COPY requirements.txt ./
 
 RUN apk add --no-cache --virtual build-dependencies g++ make musl-dev && \
     apk add --no-cache libstdc++ gnupg yaml-dev libffi-dev openssl-dev && \
+    apk add --no-cache git && \
```
adrianchifor: can you move this to line 9 as those are the dependencies we don't remove

kapitan/cli.py
```diff
@@ -276,8 +279,7 @@ def _search_imports(cwd, imp):
         compile_targets(args.inventory_path, search_paths, args.output_path,
                         args.parallelism, args.targets, ref_controller,
                         prune=(args.prune), indent=args.indent, reveal=args.reveal,
-                        cache=args.cache, cache_paths=args.cache_paths,
-                        jinja2_filters=args.jinja2_filters)
+                        cache=args.cache, cache_paths=args.cache_paths, fetch_dependencies=args.fetch, jinja2_filters=args.jinja2_filters)
```
adrianchifor: line too long

adrianchifor:
Great work!

uberspot:
Merged! :fireworks: 
I'll start another -rc version to make sure this is tested early on.

Issue #143, opened 2018-09-03
importing helm charts support


ramaro:

ramaro:
The idea is to transfer the helm chart, translate variables into an inventory reclass class and yaml structure to json/jsonnet. Still uncertain about go templates support, but it will either be translated into reclass references or we'll package kapitan with the go templating library.

uberspot:
Can be done as part of #91 . 
More info on helm issue: 
```
- make new input type called helm, 
- import helm charts, 
- behind the scenes the values of yaml file are 
  translated to inventory and then we compile the 
  golang templates library to be used with kapitan when compiling.
```

harsh-98:
@ramaro  Can explain what you mean by `it will either be translated into reclass references or we'll package kapitan with the go templating library.` . 
Does packaging kapitan with go templates, means to create a c shared object and create python bindings for it? 
And i didn't understand the reclass references part?

harsh-98:
I am not able to find support (or any lib) which provides bindings for go templates lib. I search on pypi and google search too. For helm, a lib named [pyhelm](https://github.com/flaper87/pyhelm) is there. It uses protoc to compile protobuf messages to python. So, that gRPC communication to tiller can happen from python. It is basically a helm client. It can load charts but it doesn't render them, just sends them to tiller where they are used.

ramaro:
We don't want to take the tiller approach. We want instead a new helm input type which we can compile. For this we'll need to build a golang templating library that is loadable from python (c bindings).
By inventory references, I meant that values will get translated/copied into new inventory component classes (prefered) or we'll need to change kapitan to read values.yml from elsewhere other than the inventory (not great)

caioaao:
why would we need to load values.yml into inventory? wouldn't it be easier to just make the new helm input type with options that translate to helm's cli options? For example, writing this:
```yml
parameters:
  kapitan:
    compile:
      - input_type: helm
        output_path: manifests/istio.yml
        input_path: path_to_istio/helm_package
        values_path: overrides.yml
        name: istio
        namespace: istio-system
        set_values:
          - grafana.enabled: true
          - servicegraph.enabled: true        
```

would have the same effect, when compiling, as calling helm directly like so

```bash
helm template path_to_istio/helm_package --name istio --namespace istio-system \
     --set grafana.enabled=true \
     --set servicegraph.enabled=true \
     -f overrides.yml \
     > compiled/target/manifests/istio.yml
```

ramaro:
[...]

PR #307, opened 2019-06-10
Helm input type

yoshi-1224:
Fixes issue #143 

@uberspot @ramaro  I will work on this alongside #304 in light of the proposed deadline and to facilitate discussion. Let us first settle the options for this input type! I've made some initial suggestion for the inventory syntax in the first commit. 

Also, I want to clarify:
- the `input_path` should take in a chart directory. Should we allow `tar.gz` as well? At what stage should kapitan untar the `tar.gz` file fetched using dependency manager? I'm thinking some function to untar the dependencies output_path would be helpful (e.g. `output_path: file.tar.gz | untar`)
- is my understanding ok that data only flows from kapitan inventory to helm input type to override the existing values?

Update questions:
- some charts require dependencies specified in `requirements.yaml`. Should users do this manually using http dependency or should helm input type handle them automatically?
- currently I've made a really rough PoC for this input type which modifies [template.go](https://github.com/helm/helm/blob/master/cmd/helm/template.go) to accept C string for chart directory. The resulting `.so` file is ~60MB as it includes most if not all the packages. Should I flatten out and make a single `template.go` file that includes nothing else but all the code that it uses that belong to different packages to minimize the size of the `.so` file?


ramaro:
> * the `input_path` should take in a chart directory. Should we allow `tar.gz` as well? At what stage should kapitan untar the `tar.gz` file fetched using dependency manager? I'm thinking some function to untar the dependencies output_path would be helpful (e.g. `output_path: file.tar.gz | untar`)

Good idea! However I wouldn't add a function to `output_path` and instead introduce a new option e.g. `unpack: true` (which could unzip, untar gz, etc...) and keep `output_path` as the destination directory. I think this would make it clearer as to me it doesn't make sense to have a `.tar.gz` as `output_path`

ramaro:
> * is my understanding ok that data only flows from kapitan inventory to helm input type to override the existing values?

To me, values in `values.yml` should be the default values. The user will then be able to override them using the inventory. So, yes :)

ramaro:
> some charts require dependencies specified in requirements.yaml. Should users do this manually using http dependency or should helm input type handle them automatically?

Ideally, yes! However we don't need to implement it at this stage if it complicates things to start with.

>currently I've made a really rough PoC for this input type which modifies template.go to accept C string for chart directory. The resulting .so file is ~60MB as it includes most if not all the packages. Should I flatten out and make a single template.go file that includes nothing else but all the code that it uses that belong to different packages to minimize the size of the .so file?

Agree.

yoshi-1224:
@ramaro @uberspot 
I've managed to shrink the .so file to 12MB. I think it's a reasonable size. What do you think? 

Also, regarding the errors that take place in the go side of things, should I print the error message in go and then simply return an integer to Python where 0 indicates no error and -1 indicates error, or should the bridging be a little bit more sophisticated (e.g. passing back the error string etc.)?

yoshi-1224:
Also, where should the go source and shared object exist in the kapitan directory? Since helm is an input type, I'm thinking we could make `kapitan/inputs/helm` dir. In addition to that, we need to `go get` the imports that we use in the source. I'm thinking of:
1. set $GOPATH to kapitan/inputs/helm
2. run `go get` to install the dependencies into kapitan/inputs/helm/src
3. build the .so/.h file using `go build -buildmode=c-shared`

uberspot:
The .so file size is acceptable for now i think. Regarding the errors, if the go error is granular enough by itself i'd suggest keeping it simple for start  and just returning a 0, -1 to python is fine. If the go error is too confusing for people we might have to introduce some better error outputs in the future

uberspot:
`kapitan/inputs/helm` sounds fine to me. Will you add the .so build steps to the dockerfiles and pip package too? MANIFEST.in defines what's included in the packaging. :)

ramaro:
> Also, regarding the errors that take place in the go side of things, should I print the error message in go and then simply return an integer to Python where 0 indicates no error and -1 indicates error, or should the bridging be a little bit more sophisticated (e.g. passing back the error string etc.)?

I think we should always pass back the error string - if we don't, it will make debugging errors pretty complicated! If that's too complex to do for now, please make a TODO comment and carry one. Thanks

yoshi-1224:
@uberspot  @ramaro 
I just committed a minimal python function that can render a chart at `chart_dir` and output the result to `output_path`. 

The dependency management is done using go modules, and the binding is done using cffi. 

Could we now please discuss what options we should make available for this input type (i.e. parameters in the yaml file)?

@ramaro with regards to the error string, I think we can make this function return a C string, which is the error string back to python. If no error, then it will return an empty string.

kapitan/inputs/helm/chart/acs-engine-autoscaler/.helmignore
```diff
@@ -0,0 +1,21 @@
+# Patterns to ignore when building packages.
```
ramaro: move this chart into tests/ ?

yoshi-1224:
@uberspot @ramaro 
I've added the build steps to Dockerfile, installing go, git and other additional utilities required to build the .so file. Then I wondered how we should do this for user local environment. Do we force them to install go etc. in order to build the .so file locally during `pip install`, or should this be made optional (i.e. if build fails due to missing system dependencies, skip it), or do we ship the pre-built .so file?

uberspot:
I would personally add a pointer in the readme to the build code/instructions in the dockerfile along with also shipping a prebuilt .so file. That way people can both use it immediately but have the option to build it themselves for development. Ideally split the command that install the dependencies in the dockerfile to a separate RUN step and then add a commwnt before that explaining why those are installed. :)

kapitan/inputs/helm/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+cd $(dirname "$0")
+go build -buildmode=c-shared -o libtemplate.so template.go
```
uberspot: It's good that you have this buiild script. 
I'd also suggest you add a step in the `Makefile` in the root directory to call this build script as well. 
Something like `make build_helm_bindings` maybe ? 
That way we can add it to .travis.yml and also test continuously whether the .so lib builds fine or not (?).

kapitan/inputs/helm/template.go
```diff
@@ -0,0 +1,327 @@
...
+//readFile load a file from the local directory. remote file is disabled.
+func readFile(filePath string) ([]byte, error) {
+	return ioutil.ReadFile(filePath)
+}
+
+// Merges source and destination map, preferring values from the source map
+func mergeValues(dest map[string]interface{}, src map[string]interface{}) map[string]interface{} {
+	for k, v := range src {
+		// If the key doesn't exist already, then just set the key to that value
+		if _, exists := dest[k]; !exists {
```
uberspot: `exists` doesn't seem to be used after this if. Is there a point to assign `dest[k]` to it? 
Would `if !dest[k] { ` work?

kapitan/inputs/helm/build.sh
yoshi-1224: Yes this is exactly what got me thinking about when we should run this script to build from source. For testing as well, do we require the kapitan developers to have go available so that when they run `make test` the build is triggered such that we can test the helm bindings?

uberspot: Don't add this under `make test`. We can easily have another step, called `build_binding` and run that only on Travis (by adding it to .travis.yml explicitly). That way most developers who won't touch the bindings don't need to install unnecessary dependencies. But at the same time we also have continuous testing to make sure we catch regressions in that build. :)

yoshi-1224:
@uberspot @ramaro 
I've written Helm as an input type just like any other types in `compile_target` function:

```python 
    jinja2_compiler = Jinja2(compile_path, search_paths, ref_controller)
    jsonnet_compiler = Jsonnet(compile_path, search_paths, ref_controller)
    kadet_compiler = Kadet(compile_path, search_paths, ref_controller)
    helm_compiler = Helm(compile_path, search_paths, ref_controller)
```
The issue that I got is that when we use multiprocessing and try to use the Go shared object that is `dlopen`ed in the parent process, the Go code will hang ([reference](https://github.com/golang/go/issues/15538)). It will work fine in single-threaded environment e.g. calling `helm_render` in test methods directly.

I guess the options that we have is:
1. calling `dlopen` in each of the `compile_target` function, assigning go runtime per process. I've tested it and it seems to work.
2. making helm compile process run in sequence. 

Should I go with option 1?

uberspot:
Will option 1 mean that Go will be a hard dependency for anyone wanting to use kapitan even if it's not using helm charts?

yoshi-1224:
So long as they don’t put in helm as input type in inventory and kapitan
doesn’t call the helm template function, it will run just fine without the
shared object.

yoshi-1224:
@uberspot @ramaro 
As for overriding the values of helm template via inventory, I'm not totally sure how to go about it. The basic key-value mapping is simple as can easily be done via inventory (e.g. `--set foo=bar`), but more complex ones as outlined [here](https://github.com/helm/helm/blob/master/docs/using_helm.md#the-format-and-limitations-of---set) such as lists, nested objects etc. how should kapitan users be able to specify them inside inventory?

Also, with regards to namespacing the inventory: without `helm` namespace, we will end up passing the whole inventory for overriding default helm template values, which seems quite inefficient. How should we design the inventory structure such that it achieves the appropriate level of kapitan-helm coordination?

ramaro:
@yoshi-1224 How about we try something simpler for a start - a new `helm_values` key can be supported in a compile item, e.g.:
```yaml
...
    compile:
      - output_path: chart
        input_type: helm
        input_paths:
          - charts/nginx-ingress
        helm_values:
          controller:
            name: mycontroller
            image:
               repository: example.com/repo
```

The user will then be responsible of passing in their own values within in a target or class by slightly modifying the compile item values of `helm_values` to make it read parameters from somewhere else, like `my_nginx_chart`:

```yaml
parameters:
   my_nginx_chart:
     helm_values:
       controller:
         name: controller1
         image:
           repo: repos.com/repo1

  kapitan:
     compile:
        - output_path: chart
          input_type: helm
          input_paths:
            - charts/nginx-ingress
          helm_values: ${my_nginx_chart:helm_values}
```

All you need to do is pass the `helm_values` mapping as the context for the chart rendering function. This way we (the devs) don't have to map any other keys other than `helm_values` as it is generic. 
It also makes it pretty simple for a user to override values when necessary.

yoshi-1224:
@ramaro Thanks for the suggestion.
So does this basically mean that we are creating `values.yaml` structure in the inventory, and then passing it to `helm template` to overwrite some of the values in the chart i.e. similar to `-f` (`--values`) option in `helm template`?

ramaro:
@yoshi-1224 We will still honour the chart values inside values.yml if not set in the inventory. Should any be set in the inventory, then yes, they should be passed into `helm template` to override those chart values as you mention, like when using `-f` `(--values)`.

yoshi-1224:
@ramaro @uberspot 
I've implemented the `helm_values` inventory option. To make the binding simple, we first write out the `helm_values` object into a yaml file, and then pass its file path to Go which reads the file and uses the values inside.

ramaro:
@yoshi-1224 thanks for this! Can you resolve the conflicts please?

kapitan/inputs/helm/template.go
```diff
@@ -0,0 +1,320 @@
...
+	var valueFiles []string
+	if valueFile != "" {
+		valueFiles = append(valueFiles, valueFile)
+	}
+	var values []string
+	// to force string values
+	var stringValues []string
+	var fileValues []string
+	nameTemplate := ""
+	releaseName := ""
```
ramaro: @yoshi-1224 we should also allow passing this value through the inventory.
Maybe by implementing `helm_params` (similar to `helm_values`) where one can set params like `namespace`, `release`, etc... ?

kapitan/inputs/helm/template.go
```diff
@@ -0,0 +1,320 @@
...
+	namespace := "default"
```
ramaro: @yoshi-1224 likewise

yoshi-1224:
@ramaro @uberspot 
Ready for review. Thanks!

ramaro:
@yoshi-1224  This is awesome! I think you've mentioned somewhere you'd be updating the README on Helm support - if that's the case, consider this approved

ramaro:
@yoshi-1224 also we need to document that Helm won't include the namespace in compiled manifests as a known restriction. There will be several workarounds, which we can discuss/document too.

.travis.yml
```diff
@@ -24,11 +24,19 @@ before_install:
...
+  - sudo apt-get install -y golang-1.12
+  - export PATH="/usr/lib/go-1.12/bin:$PATH"
+  - export GOROOT=/usr/lib/go-1.12
+  - go version
 install:
   - pip3 install -r requirements.txt
   - pip3 install coverage
 
 script:
+  - make build_helm_binding
```
uberspot: move this after `make test` below.
Helm bindings will change less often that normal changes that "make test" should cover. So it's good to fail first on the make test, then on helm bindings. :)

uberspot:
Besides a minor comment the rest looks good! :) I'd agree with @ramaro that we should mention the namespace issue in the KAP helm document as a known restriction on the helm side just so we don't get issues created on our side for it. And also add a paragraph in the readme for the new helm support? Pointers to examples etc. Great work so far! :+1:

yoshi-1224:
@ramaro @uberspot 
Thanks for the review! I'll be adding to the README some guide for this feature.

> Helm won't include the namespace in compiled manifests as a known restriction

Sorry I'm a bit confused: so right now what we do is if `helm_params.namespace` is not set, then 
 `.Release.Namespace` would be `default`. Users can freely change the templates such that they can set the namespace to be any custom values (hard-coded, or from template variables). Could you please clarify what exactly I need to warn the users about?

.travis.yml
yoshi-1224: @uberspot 
I understand the rationale. If I want to test helm binding as well, then should I do:
`make test` followed by `make build_helm_binding` and then `make test_helm_binding` where I only test the helm binding? Or should I include the testing of helm_binding in the `make build_helm_binding` step?

uberspot: You can have `build_helm_binding` and `test_helm_binding` as separate build steps, no problems there.
Just put `make test` first in the order so it fails early then follow up with the other two build steps. :)

ramaro:
@yoshi-1224 what I meant is that `metadata.namespace` is not added to all compiled manifests, which is a helm limitation, not necessarily your implementation's limitation!

.travis.yml
uberspot: Ah, I just got that you are running this build step before the test because the test also tests the generate library. 
Splitting these two test_helm_binding might be making it unnecessarily complex. Let's skip this for now. :)

yoshi-1224:
@ramaro Thanks for the response! I see, so it's that users are encouraged to **explicitly** have:

```yaml
metadata:
  namespace: {{ .Release.Namespace }} # or some other custom namespace
```
in all templates in a chart when using `helm template` command.

yoshi-1224:
@ramaro @uberspot 
The relevant doc for helm is right now in https://github.com/deepmind/kapitan/pull/331 (https://github.com/deepmind/kapitan/pull/331/files#diff-bf1b6d5bfe119bf7d3ecaf4dc0636334R143). 
Should I add this to the current README with this PR? I anticipate the whole restructuring of the documentation may take more time.

uberspot:
A separate PR is also fine :) lets continue the doc stuff there. In the mean time we can get helm inputs in and start another rc.

PR #308, opened 2019-06-11
Multi-document yaml outputs

caioaao:
This way we can have a multi-document yaml file (see [this](https://pyyaml.org/wiki/PyYAMLDocumentation#dumping-yaml) for more info).

As an example, consider this jsonnet:
```jsonnet
{
  'foo': [
    {'a': 1}, {'a': 2}, {'a': 3}
  ]
}
```
Previously the output (when type=yaml) would be:
```yaml
- a: 1
- a: 2
- a: 3
```

With this change, it's:
```yaml
a: 1
---
a: 2
---
a: 3
```

### DIFF
--- a/examples/kubernetes/compiled/minikube-mysql/manifests/mysql_app.yml
+++ b/examples/kubernetes/compiled/minikube-mysql/manifests/mysql_app.yml
@@ -1,0 +1,81 @@
+apiVersion: apps/v1beta1
+kind: StatefulSet
+metadata:
+  annotations: {}
+  labels:
+    name: example-mysql
+  name: example-mysql
+  namespace: minikube-mysql
+spec:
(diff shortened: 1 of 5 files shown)


uberspot:
Just to validate, this is to generate k8s manifests in one file? 
It would be awesome if you could add 1-2 sentences in the readme about this functionality so that users are more aware? :)

caioaao:
@uberspot Any manifest, really. My use case was for generating a kops manifest. Anyway, I'll add it to the readme :+1:

adrianchifor:
You can add to Readme anywhere you see fit, we can review later. Can you add a test?

caioaao:
Sure. Btw, I just realized the reveal command also needs to be changed

Issue #105, opened 2018-05-25
Pull secrets from Hashicorp Vault


DWSR:
As a Vault user, I would like Kapitan to be able to pull information from Hashicorp Vault so that I can have a single authority on secrets in my organization.

**Acceptance Criteria:**
* Supports both v1 and v2 [KV backends](https://www.vaultproject.io/docs/secrets/kv/index.html) including the ability to specify a version of a value where appropriate.
* Inherits or detects Vault information from the environment/other mechanisms (i.e. reads the token from `~/.vault-token`, uses `VAULT_ADDR`, etc.)
* Has the ability to explicitly specify Vault information to override inherited or detected configurations.
* Supports basic CRUD operations on an arbitrary secret backend
* Supports, at minimum, the `userpass`, `token`,  and `cert` authentication mechanisms.

DavidJeche:
This looks interesting but i would suggest the best thing to do here might be to generate config i.e for terraform, generate the terraform with consul template variable in place where secret should be. 

When you run a deployment set your CI system to run consul template on the compiled file. This means you can keep the manifests,  secrets and environments independent of kapitan tool.

DWSR:
So if I understand this correctly, you're suggesting templating out a consul-template file with Jinja (because I doubt you can get this to work in JSonnet)?

DavidJeche:
I am not that familiar with Jsonnet and I am currently learning as a configuration language. I am not even sure how to manage dependencies with in terraform like how I reference an alternative resource or take output from a resource like IP address. 

The main suggestion was to add string tags something like this should work:
` {
  locations: {
    'MyDataCenters': {
      description: "{{ range datacenters }}
{{ . }}{{ end }}",
   },
  },
}`

It will output the following:

`{
  "locations": {
    "MyDataCenters": {
      "description": "{{ range datacenters }}\n{{ . }}{{ end }}"
    }
  }
}`

What you can do is in the deployment environment run consul template in the step before the deployment. This will create new file or temp file that have the secrets. The files are then deleted after the apply. You still have the original compiled files for reference of what been applied. 

I would recommend not to go too crazy with it. There will most likely be clashies somewhere but its a workable solutions until the vault integration is added. I am fairly certain it would be there before version one.

calvinmorrow:
I started looking into whether Vault KV support was something I could add, but currently a bit tripped up on how to handle the `RefBackend.__getitem__` method.  

[...]

PR #310, opened 2019-06-13
Vault secret backend

vaibhavk:
Fixes issue #105 

## Proposed Changes

- [x]  Added Vault as a secret backend for kv/kv-v2 secret engine
- [x]  Add Tests


vaibhavk:
@adrianchifor @ramaro review required

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,84 @@
...
+* VAULT_CLIENT_KEY: the path to an unencrypted PEM-encoded private key matching the client certificate  
+* VAULT_CLIENT_CERT: the path to a PEM-encoded client certificate for TLS authentication to the Vault server  
+* VAULT_CACERT: the path to a PEM-encoded CA cert file to use to verify the Vault server TLS certificate  
+* VAULT_CAPATH: the path to a directory of PEM-encoded CA cert files to verify the Vault server TLS certificate  
+* VAULT_NAMESPACE: specify the Vault Namespace, if you have one  
+  
+Considering a key-value pair like `my_key`:`my_secert` ( in our case let’s store hello:batman inside the vault ) in the path `secret/foo` on the vault server, to use this as a secret either follow:  
+  
+```shell  
+$ echo “{'path':'secert/foo','key':'my_key'}” > somefile.txt
```
ramaro: typo: "secert"

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,84 @@
...
+* VAULT_SECRET_ID: (required by approle)  
+* VAULT_USER: username to login to vault  
+* VAULT_PASSWORD: password to login to vault  
+* VAULT_CLIENT_KEY: the path to an unencrypted PEM-encoded private key matching the client certificate  
+* VAULT_CLIENT_CERT: the path to a PEM-encoded client certificate for TLS authentication to the Vault server  
+* VAULT_CACERT: the path to a PEM-encoded CA cert file to use to verify the Vault server TLS certificate  
+* VAULT_CAPATH: the path to a directory of PEM-encoded CA cert files to verify the Vault server TLS certificate  
+* VAULT_NAMESPACE: specify the Vault Namespace, if you have one  
+  
+Considering a key-value pair like `my_key`:`my_secert` ( in our case let’s store hello:batman inside the vault ) in the path `secret/foo` on the vault server, to use this as a secret either follow:
```
ramaro: https://www.meetup.com/London-Kapitan-Meetup/events/261900327/

ramaro: typo: "secert"

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,84 @@
+# Hashicorp Vault  
+  
+This feature allows the user to fetch secrets from [Hashicorp Vault](https://www.vaultproject.io/), an online Kapitan Secret backend `vault` will be introduced.
```
loknathteja: Minor sentence formation issue. Remove the suffix "will be introduced"

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,167 @@
...
+        env['verify'] = False
+    elif verify.lower() == 'false':
+        cert = getenv( 'VAULT_CACERT' )
+        if not cert:
+            cert_path = getenv( 'VAULT_CAPATH' )
+            if not cert_path:
+                raise Exception('Neither VAULT_CACERT nor VAULT_CAPATH specified')
+            env['verify'] = cert_path
+        else:
+            env['verify'] = cert
```
loknathteja: Make sure to add test cases to cover all the conditions coded here

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._decrypt(ref_data)
+
+    def _decrypt(self, data):
+        """Decrypt data & return value for the key from Vault Server
+
+        :returns: secret in plain text
+
+        """
+        try:
+            if data.decode() == "secret_test_key":
```
loknathteja: Avoid having test case data inside method. Rather use mock python unit test library to achieve the same. Change the coding design to incorporate composition pattern. Refer to https://docs.python.org/3/library/unittest.mock.html

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,167 @@
...
+                return "secret_value"
+            else:
+                client = vault_obj()
+                data = safe_load(data)
+                response = client.read(data['path'])
+                return response['data']['data'][data['key']]
```
loknathteja: Can you explain this and suspect there is need for some checks with json data extraction to deal with error scenarios..

`['data']['data'][data['key']]` what format are you expecting the data to be ? how do you handle scenarios if the response format is out of order ? focus on these aspects

tests/test_cli.py
```diff
@@ -262,6 +262,37 @@ def test_cli_secret_write_base64_ref(self):
 
         os.remove(test_tag_file)
 
+    def test_cli_secret_write_vault(self):
```
loknathteja: Add test cases to cover other workflows/scenarios discussed in the previous comments

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,167 @@
...
+
+from kapitan.refs.base import Ref, RefBackend, RefError
+from kapitan import cached
+from kapitan.errors import KapitanError
+
+class VaultError(KapitanError):
+    """Generic vault errors"""
+    pass
+
+def get_env():
```
ramaro: I'd advise you not to only support env variables - ensuring the values are in the inventory is more kapitanesque :)
I'd like to see this supporting vault settings in both env and inventory.

vaibhavk: The reason I didn't use inventory as a store for the variable is due to the security concerns, we would be revealing the server address or tokens, but instead of those all other variables I guess can be included. Would that be acceptable?

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,167 @@
...
+
+    """
+    Hashicorp Vault can be used if using KV Secret Engine
+    """
+
+    def __init__(self,data,encode_base64=False,**kwargs):
+        """
+        set encoding_base64 to True to base64 encoding key before encrypting and writing
+        """
+        self._encrypt(data, encode_base64)
```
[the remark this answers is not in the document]
vaibhavk: This would include a large number, would request @uberspot @ramaro to suggest the one they think should be included or are planning to include in the future.

kapitan/refs/secrets/vault.py
vaibhavk: the default format of response for the secret in JSON goes like,
```json
{
  "request_id": "3a14fc9c-2a00-5dd1-3f24-3556c2cc6248",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": {
    "data": {
      "my_key": "my_precious_secret"
    },
    "metadata": {
      "created_time": "2019-06-27T08:19:20.584891Z",
      "deletion_time": "",
      "destroyed": false,
      "version": 1
    }
  },
  "warnings": null
}
```
and if there are multiple keys, then
```json
"data": {
    "data": {
      "my_key": "my_precious_secret",
      "hello": "user",
    },
```
and this response is consistent throughout `vault get` or `vault read`
that was the reason I used `response['data']['data'][data['key']]`

tests/test_vault.py
```diff
@@ -0,0 +1,145 @@
...
+        tag = '?{vault:secret/batman}'
+        secret = { 'some_random_value':'somethin_secret' }
+        client.secrets.kv.v2.create_or_update_secret(
+            path='foo',
+            secret=secret,
+        )
+        os.environ['VAULT_TOKEN'] = TOKEN
+        os.environ['VAULT_AUTHTYPE'] = 'token'
+        file_data = "{'path':'foo','key':'some_random_value'}".encode()
+        REF_CONTROLLER[tag] = VaultSecret(file_data,encrypt=True,encoding='original')
```
daminisatya: can you explain what functionality of the original implementation are you testing?

tests/test_vault.py
```diff
@@ -0,0 +1,145 @@
...
+        os.environ['VAULT_AUTHTYPE'] = 'approle'
+        os.environ['VAULT_ROLE_ID'] = ROLE_ID
+        os.environ['VAULT_SECRET_ID'] = SECRET_ID
+        test_client = vault_obj()
+        self.assertTrue(test_client.is_authenticated())
+        test_client.adapter.close()
+
+    def test_vault_write_reveal(self):
+        '''
+        Write secret, confirm secret file exists, reveal and compare content
```
daminisatya: it defeats the purpose of unit testing. you testing multiple activities in one test function. 

How exactly are you testing the following methods? 

- reveal(self)
- decrypt
- dump

daminisatya: We need tests for the cli option as well. test_cli.py needs those test cases.

Makefile
```diff
@@ -2,6 +2,10 @@ all: clean package
 
 .PHONY: test
 test:
+	@echo ----- Running docker container for vault tests-----
+	docker run --cap-add=IPC_LOCK --rm -p 8200:8200 -d -e \
+		'VAULT_LOCAL_CONFIG={"backend": {"file": {"path": "/vault/file"}}, "listener":{"tcp":{"address":"0.0.0.0:8200","tls_disable":"true"}}}'\
+		vault server
```
daminisatya: I believe we don't need this. We can definitely mock all these values in the test case itself. If you have written this to test the call to the vault server during the reveal, it's not required. It's like we are testing the vault, not the feature.

Makefile
```diff
@@ -2,6 +2,10 @@ all: clean package
 
 .PHONY: test
 test:
+	@echo ----- Running docker container for vault tests-----
+	docker run --cap-add=IPC_LOCK --rm -p 8200:8200 -d -e \
+		'VAULT_LOCAL_CONFIG={"backend": {"file": {"path": "/vault/file"}}, "listener":{"tcp":{"address":"0.0.0.0:8200","tls_disable":"true"}}}'\
```
daminisatya: I believe we don't need this. We can definitely mock all these values in the test case itself. If you have written this to test the call to the vault server during the reveal, it's not required. It's like we are testing the vault, not the feature.

adrianchifor: Agreed, this should be covered in python tests, we should not pollute the Makefile with specific cases.

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,84 @@
+# Hashicorp Vault  
+  
+This feature allows the user to fetch secrets from [Hashicorp Vault](https://www.vaultproject.io/), an online Kapitan Secret backend `vault`.  
+  
+Author: [@vaibahvk](https://github.com/vaibhavk) [@daminisatya](https://github.com/daminisatya)  
+## Specification  
+  
+The following variables need to be exported to the environment where you run this script in order to authenticate to your HashiCorp Vault instance:
```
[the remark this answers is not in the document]
adrianchifor: Also, the way we configure secret backend variables is within the inventory, not env vars. e.g. the recipients for gpg or keys for gkms/awskms as part of `kapitan.secrets.<backend>`. Maybe it would make sense for the vault token but not much else.

requirements.txt
```diff
@@ -11,5 +11,6 @@ boto3==1.9.138
 requests==2.21.0
 addict==2.2.1
 yamllint>=1.15.0
+hvac==0.9.1
```
adrianchifor: 0.9.2 is now available

docs/kap_proposals/kap_6_hashicorp_vault.md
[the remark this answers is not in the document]
vaibhavk: I have added the support to provide set these variables from within the inventory.

Makefile
vaibhavk: Can make a separate section for this specific test. But would making the mock tests for vault feature be just like hard-coding the results?

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,186 @@
...
+        env['cert'] = (client_cert,client_key)
+    return env, auth_type
+
+def vault_obj():
+    env, auth_type = get_env()
+    client = hvac.Client(
+        **{k:v for k,v in env.items() if k not in ( 'username','password','role_id','secret_id','token' )}
+    )
+
+    if auth_type == 'token':
```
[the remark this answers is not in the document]
vaibhavk: That can be done but currently, this doesn't include a lot of cases

kapitan/cli.py
```diff
@@ -398,8 +400,34 @@ def secret_write(args, ref_controller):
...
 
+    elif token_name.startswith("vault:"):
+        type_name, token_path = token_name.split(":")
+        _data = data.encode()
+        parameter = {}
+        encoding = 'original'
+        if args.base64:
+            _data = base64.b64encode(_data).decode()
+            _data = _data.encode()
+            encoding = 'base65'
```
loknathteja: Fix the encoding scheme name.. base64 instead of base65.. Its surprising that despite setting the base65 encoding, the decoding is successful. This is potential bug scenario which needs to be fixed and new workflow test to be added as well.

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,167 @@
...
+        else:
+            self.data = data.encode()
+
+
+    def reveal(self):
+        """
+        returns decrypted data
+        """
+        # can't use super().reveal() as we want bytes
+        ref_data = base64.b64decode(self.data)
```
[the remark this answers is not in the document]
loknathteja: @daminisatya The refactoring in cli.py removing encode_base64 flag in older patch has addressed part of my concern. However, my concern is yet to be fixed to deal with various encoding schemes specifically and case when encoding scheme is wrongly set , otherwise can lead to potential bug scenario pointed out in 411 of cli.py (please see my latest comment) although setting up non-existing/unacceptable encoding scheme base65 still successfully decode using base64. When the reference controller holding the secret_obj during encode in cli.py is passed to the reveal in cli.py, the state seems to be available for the reveal(..) which has to be utilised for conditional based decoding based on encoding scheme. @adrianchifor  please clarify my understanding of the encrypt and reveal feature.

kapitan/refs/secrets/vault.py
vaibhavk: Support for both (env variables & values from inventory) are added now

kapitan/cli.py
vaibhavk: This wasn't detected since I removed the `vault` tests from `test_cli.py`. Also in unittest each component of the vault.py was tested. I will be adding the test for `cli` again after concluding the discussion regarding testing using docker container.

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,186 @@
...
+    def __init__(self,data,encrypt=False,**kwargs):
+        """
+        set encoding_base64 to True to base64 encoding key before encrypting and writing
+        """
+        self.encoding = kwargs.get('encoding', 'original')
+        self.data = data
+        super().__init__(self.data,**kwargs)
+        self.type_name = 'vault'
+
+    def reveal(self):
```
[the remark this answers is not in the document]
vaibhavk: This is the default function name used to reveal for all secret backends [called from base.py](https://github.com/deepmind/kapitan/blob/640aba6531a1a6a82aafc286b3677c3143106aed/kapitan/refs/base.py#L265).

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,186 @@
...
+        returns decrypted data
+        """
+        # can't use super().reveal() as we want bytes
+        if self.encoding == 'base64':
+            self.data = base64.b64decode(self.data)
+
+        ref_data = base64.b64decode(self.data)
+        return self._decrypt(ref_data)
+
+    def _decrypt(self, data):
```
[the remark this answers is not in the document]
vaibhavk: Following the same naming of function as specified in other secret back-ends. `reveal` decodes the data, passes the data in plain text to `_decrypt` which then authenticate with `vault` server and returns the secret in plain text.
[_decrypt in gpg](https://github.com/deepmind/kapitan/blob/640aba6531a1a6a82aafc286b3677c3143106aed/kapitan/refs/secrets/gpg.py#L146)   
[_decrypt in gkms](https://github.com/deepmind/kapitan/blob/640aba6531a1a6a82aafc286b3677c3143106aed/kapitan/refs/secrets/gkms.py#L146)  
[_decrypt in awskms](https://github.com/deepmind/kapitan/blob/640aba6531a1a6a82aafc286b3677c3143106aed/kapitan/refs/secrets/awskms.py#L131)  
Can remove the function if found necessary and move all the code into `reveal` itself

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+        Decrypt data & return value for the key from Vault Server
+
+        :returns: secret in plain text
```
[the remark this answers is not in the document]
vaibhavk: By sample response are you referring to the JSON response vault give?

tests/test_vault.py
daminisatya: For the first iteration the tests are good. please resolve this.

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,229 @@
...
+            return response['data'][data['key']]
+        elif data['key'] in response['data']['data']:
+            return response['data']['data'][data['key']]
+        else:
+            exit(
+                "'{key}' doesn't exist on '{path}'".format(key=data['key'],
+                                                           path=data['path'])
+            )
+
+    def dump(self):
```
ramaro: Why are we dumping to disk? What happens if a secret is changed by anything else other than Kapitan?
Vault's KV engine is the canonical source for the content of the secret. Dumping it to file will potentially make revealing wrong or misleading.

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,229 @@
...
+    * VAULT_USERNAME: username to login to vault
+    * VAULT_PASSWORD: password to login to vault
+    * VAULT_CLIENT_KEY: path to an unencrypted PEM-encoded private key matching the client certificate
+    * VAULT_CLIENT_CERT: path to a PEM-encoded client certificate for TLS authentication to the Vault server
+    * VAULT_CACERT: path to a PEM-encoded CA cert file to use to verify the Vault server TLS certificate
+    * VAULT_CAPATH: path to a directory of PEM-encoded CA cert files to verify the Vault server TLS certificate
+    * VAULT_NAMESPACE: specify the Vault Namespace, if you have one
+    """
+    # Helper lambda function to get varibles either from parameters or environment
+    get_variable = lambda x: parameter.get(x, getenv('VAULT_'+x.upper(),default=''))
```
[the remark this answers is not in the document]
vaibhavk: There was no need for the lambda function but just to keep the code simple, readable & short.

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,229 @@
...
+    env = {}
+    env['url'] = get_variable('addr')
+    env['namespace'] = get_variable('namespace')
+    # AUTHENTICATION TYPE
+    if parameter.get('auth','token') in ['token','github']:
+        env['token'] = getenv('VAULT_TOKEN')
```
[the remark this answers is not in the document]
vaibhavk: Using get_variable to grab vault token would be inappropriate here since the token would be either in the environment or in `~/.vault-token` but never in the inventory.

ramaro:
@vaibhavk any plan on fixing the items we've reviewed?

prakashrj:
@ramaro. Can you please review these new updates?

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,136 @@
...
+or in a single line  
+```shell  
+$ echo “{'path':'secret/foo','key':'hello'}” | kapitan secrets --write vault:path/to/secret_inside_kapitan -t dev-sea -a userpass -f -  
+```  
+The entire string __{'path':'secret/foo','key':'hello'}__ is base64 encoded and stored in the secret_inside_kapitan. Now secret_inside_kapitan contains the following  
+  
+```yaml    
+data: 4oCccGF0aDpzZWNlcnQvZm9v4oCdIOKAnGtleTpteV9rZXnigJ0K  
+encoding: original  
+parameter:
```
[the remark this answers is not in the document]
vaibhavk: Added commit for loading parameters from inventory.
Any suggestions on how to test it? Since testing on temporary directory wouldn't work and now revealing the secret requires `--target` so testing the CLI require an example `vaultkv` secret in either of the `examples/` that we already have.

kapitan/refs/secrets/vault.py
```diff
@@ -0,0 +1,229 @@
...
+                self.data = base64.b64decode(self.data, validate=True)
+
+            ref_data = base64.b64decode(self.data, validate=True)
+        except b_error:
+            exit(
+                "non-alphabet characters in the data"
+            )
+
+        if ref_data.decode() == "secret_test_key":
+            return "secret_value"
```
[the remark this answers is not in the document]
vaibhavk: Will add all the tests in `test_vault.py`, including the one that requires testing if the CLI works or not with `vaultkv` secrets.

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,136 @@
...
+$ kapitan secrets --write vault:path/to/secret_inside_kapitan --file somefile.txt --target dev-sea --auth userpass 
+```  
+or in a single line  
+```shell  
+$ echo “{'path':'secret/foo','key':'hello'}” | kapitan secrets --write vault:path/to/secret_inside_kapitan -t dev-sea -a userpass -f -  
+```  
+The entire string __{'path':'secret/foo','key':'hello'}__ is base64 encoded and stored in the secret_inside_kapitan. Now secret_inside_kapitan contains the following  
+  
+```yaml    
+data: 4oCccGF0aDpzZWNlcnQvZm9v4oCdIOKAnGtleTpteV9rZXnigJ0K
```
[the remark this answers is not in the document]
vaibhavk: Currently, we are only encoding the `data` since they are not even secrets.

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,136 @@
...
+* VAULT_USERNAME: username to login to vault  
+* VAULT_PASSWORD: password to login to vault  
+* VAULT_CLIENT_KEY: the path to an unencrypted PEM-encoded private key matching the client certificate  
+* VAULT_CLIENT_CERT: the path to a PEM-encoded client certificate for TLS authentication to the Vault server  
+* VAULT_CACERT: the path to a PEM-encoded CA cert file to use to verify the Vault server TLS certificate  
+* VAULT_CAPATH: the path to a directory of PEM-encoded CA cert files to verify the Vault server TLS certificate  
+* VAULT_NAMESPACE: specify the Vault Namespace, if you have one  
+  
+Considering a key-value pair like `my_key`:`my_secret` ( in our case let’s store `hello`:`batman` inside the vault ) in the path `secret/foo` on the vault server, to use this as a secret either follow:  
+
```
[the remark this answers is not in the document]
vaibhavk: As you suggested, have changed the name of the secret backend `vault` -> `vaultkv`.
Adding support for `vaultts` i.e. transit engine will come soon after this :sweat_smile:

vaibhavk:
> Vault's KV engine is the canonical source for the content of the secret. Dumping it to file will potentially make revealing wrong or misleading.
One method is to check it every time command  `kapitan compile` is done on the inventory, or we can add a validate flag for the same. Both of these cases require the user to have authentication & authorizations of vault. Anyother method that you can suggest?

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,124 @@
...
+* VAULT_CLIENT_CERT: the path to a PEM-encoded client certificate for TLS authentication to the Vault server  
+* VAULT_CACERT: the path to a PEM-encoded CA cert file to use to verify the Vault server TLS certificate  
+* VAULT_CAPATH: the path to a directory of PEM-encoded CA cert files to verify the Vault server TLS certificate  
+* VAULT_NAMESPACE: specify the Vault Namespace, if you have one  
+  
+Considering a key-value pair like `my_key`:`my_secret` ( in our case let’s store `hello`:`batman` inside the vault ) in the path `secret/foo` in a kv-v2(KV version 2) secret engine on the vault server, to use this as a secret either follow:  
+  
+```shell  
+$ echo "foo:hello" > somefile.txt  
+$ kapitan secrets --write vaultkv:path/to/secret_inside_kapitan --file somefile.txt --target dev-sea
```
daminisatya: what does dev-sea contain?

requirements.txt
```diff
@@ -14,5 +14,7 @@ yamllint>=1.15.0
 cffi
 rfc3987==1.3.8
 GitPython==2.1.11
+hvac==0.9.2
+docker==4.0.2
```
daminisatya: We need docker for?

vaibhavk: `docker` or `docker-py` is needed for testing the authentication and getting the secrets from Hashicorp Vault. Currently, I'm spinning up a docker container of Vault and initializing it with test policies, authentication methods, and secrets, then testing it.

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -0,0 +1,124 @@
...
+        auth: userpass
+        addr: http://127.0.0.1:8200
+        engine: kv-v2
+        namespace: CICD-alpha
+        mount: team-alpha-secret
+        skip_verify: false
+        client_key: /path/to/key
+        client_cert: /path/to/cert
+```
+Almost all the environment variables can be defined in kapitan inventory except the token, password, secret_id, etc. Environment like `VAULT_TOKEN`,`VAULT_USERNAME`,`VAULT_PASSWORD`,`VAULT_ROLE_ID`,` VAULT_SECRET_ID`
```
uberspot: To avoid user confusion, can we rewrite this as a clearer note? Maybe split it to two sentences saying which envs can be declared in the kapitan inventory and which can't?

kapitan/cli.py
```diff
@@ -134,8 +135,8 @@ def main():
                                 action='store_true',
                                 default=from_dot_kapitan('compile', 'ignore-version-check', False))
     compile_parser.add_argument('--schemas-path',
-                                default=from_dot_kapitan('validate', 'schemas-path', './schemas'),
-                                help='set schema cache path, default is "./schemas"')
+                                 default=from_dot_kapitan('validate', 'schemas-path', './schemas'),
```
uberspot: Is this space for indentation?

kapitan/cli.py
```diff
@@ -357,6 +359,10 @@ def _search_imports(cwd, imp):
 
     elif cmd == 'refs':
         ref_controller = RefController(args.refs_path)
+        if args.target_name:
+            ref_controller.target = args.target_name
```
uberspot: Can you make another Constructor for RefController and pass these extra arguments with that one instead?
Defining class variables from an external class can easily lead to harder to trace bugs in the future. 
Also name those variables `target_name` and `inventory_path` so that the name tells exactly what they are. :)

kapitan/cli.py
```diff
@@ -449,6 +455,28 @@ def ref_write(args, ref_controller):
...
+            encoding = 'base64'
+        if args.target_name:
+            inv = inventory_reclass(args.inventory_path)
+            kap_inv_params = inv['nodes'][args.target_name]['parameters']['kapitan']
+            if 'secrets' not in kap_inv_params:
+                raise KapitanError("parameters.kapitan.secrets not defined in {}".format(args.target_name))
+
+            parameter = kap_inv_params['secrets']['vaultkv']
+        if parameter.get('auth') is None:
+            raise KapitanError("No Authentication type parameter Specified. Specify it"
```
uberspot: Change `Specified` to lowercase.

kapitan/cli.py
```diff
@@ -620,6 +652,11 @@ def secret_update_validate(args, ref_controller):
                         secret_obj.update_key(awskey)
                         ref_controller[token_path] = secret_obj
 
+            elif token_path.startswith("?{vaultkv:"):
+                if not vaultkv:
+                    logger.debug("secret_update_validate: target: %s has no inventory vaultkv parameters 'auth', skipping %s", target_name, token_path)
```
uberspot: Change to `no inventory vaultkv parameter 'auth',`

kapitan/cli.py
```diff
@@ -620,6 +652,11 @@ def secret_update_validate(args, ref_controller):
...
+                continue
```
uberspot: Is this `continue` indented correctly? Should it be under the `if not vaultkv:` ? 
I'm not sure if that's the case or the github UI is confusing me.

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+    """Generic vault errors"""
+    pass
+
+def get_env(parameter):
+    """
+    The following variables need to be exported to the environment depending on authentication needed
+    * VAULT_ADDR: url for vault
+    * VAULT_SKIP_VERIFY=true: if set, do not verify presented TLS certificate before communicating with Vault server.
+    * VAULT_TOKEN: token for vault
+    * VAULT_ROLE_ID: (required by approle)
```
uberspot: Remove parenthesis here and below.

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+    * VAULT_CLIENT_KEY: path to an unencrypted PEM-encoded private key matching the client certificate
+    * VAULT_CLIENT_CERT: path to a PEM-encoded client certificate for TLS authentication to the Vault server
+    * VAULT_CACERT: path to a PEM-encoded CA cert file to use to verify the Vault server TLS certificate
+    * VAULT_CAPATH: path to a directory of PEM-encoded CA cert files to verify the Vault server TLS certificate
+    * VAULT_NAMESPACE: specify the Vault Namespace, if you have one
+    """
+    # Helper lambda function to get varibles either from parameters or environment
+    get_variable = lambda x: parameter.get(x, os.getenv('VAULT_'+x.upper(),default=''))
+    env = {}
+    env['url'] = get_variable('addr')
```
uberspot: Nitpick: Can you rename the keys here to match exactly what env var they correspond to? 
It's easier to remember `env['addr']` comes from `VAULT_ADDR` than to remember what each key is called in source code vs env vars. :)

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+    env['namespace'] = get_variable('namespace')
+    # VERIFY VAULT SERVER TLS CERTIFICATE
+    skip_verify = get_variable('skip_verify')
+    if skip_verify.lower() == 'false':
+        cert = get_variable('cacert')
+        if not cert:
+            cert_path = get_variable('capath')
+            if not cert_path:
+                raise Exception('Neither VAULT_CACERT nor VAULT_CAPATH specified')
+            env['verify'] = cert_path
```
uberspot: Same for the `verify` key

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+        else:
+            env['verify'] = cert
+
+    else:
+        env['verify'] = False
```
uberspot: Don't reuse a key for either a boolean or the actual cert or the cert path. 
The name should indicate exactly what's in that dictionary key. Otherwise other developers reviewing this code will have to re-run through what codepath was taken here and guess what sort of variable will be in this key.
This is how Type Confusion bugs occur. 
Quick random link found googling: https://nebelwelt.net/publications/files/18SyScan360-presentation.pdf

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+    # CLIENT CERTIFICATE FOR TLS AUTHENTICATION
+    client_key = get_variable('client_key')
+    client_cert = get_variable('client_cert') 
+    if client_key != '' and client_cert != '':
+        env['cert'] = (client_cert,client_key)
```
uberspot: Same here. Each key should be self-explanatory.

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+
+    auth_type = vault_parameters['auth']
+    # GET TOKEN EITHER FROM ENVIRONMENT OF FILE
+    if auth_type in ['token','github']:
+        env['token'] = os.getenv('VAULT_TOKEN')
+        if not env['token']:
+            try:
+                with open(os.path.join(os.path.expanduser('~'),'.vault-token'),'r') as f:
+                    env['token'] = f.read()
```
uberspot: This might also need a check to see if the file exists but is empty (?).

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+            except IOError:
+                VaultError("Cannot read file ~/.vault-token")
+    # DIFFERENT LOGIN METHOD BASED ON AUTHENTICATION TYPE
+    if auth_type == 'token':
+        client.token = env['token']
+    elif auth_type == 'ldap':
+        client.auth.ldap.login(
```
uberspot: This needs a check to see if both env vars here exist before using them (?). 
If they don't exist log an info message?

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+                target_inv = inv['nodes'][kwargs['target_name']]
+                if target_inv is None:
+                    raise ValueError('target not set')
+
+                _kwargs['parameter'] = target_inv['parameters']['kapitan']['secrets']['vaultkv']
+                kwargs.update(_kwargs)
+                return cls(obj['data'], from_base64=True, **kwargs)
+
+        except IOError as ex:
+            if ex.errno == errno.ENOENT:
```
uberspot: Log a debug message if this occurs.

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+        return self._decrypt(ref_data)
+
+    def _decrypt(self, data):
+        """
+        Authenticate with Vault server & returns value of the key from secret
+
+        :returns: secret in plain text
+        """
+        try:
+            client = vault_obj(self.parameter)
```
uberspot: Could you make a small new class in this file with the VaultParameters instead of passing an object with an expected structure stated in comments? That way you can do all the evaluations of whether specific tokens exist in that constructor. :)

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,285 @@
...
+        "init VaultBackend ref backend type"
+        super().__init__(path, ref_type)
+        self.type_name = 'vaultkv'
+        self.ref_type = ref_type
+        self.target = target
+        self.inventory = inventory
+
+    def __getitem__(self, ref_path):
+        # remove the substring notation, if any
+        ref_file_path = re.sub(r"(@[\w\.\-\_]+)", "", ref_path)
```
uberspot: Shouldn't substrings work in the vault backend as well? Why strip them here?

tests/test_vault.py
```diff
@@ -0,0 +1,162 @@
...
+        cls.client.create_role('test_role')
+        os.environ['VAULT_ROLE_ID'] = cls.client.get_role_id('test_role')
+        os.environ['VAULT_SECRET_ID'] = cls.client.create_role_secret_id('test_role')['data']['secret_id']
+        os.environ['VAULT_TOKEN'] = cls.client.create_token(policies=[policy],lease='1h')['auth']['client_token']
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.client.adapter.close()
+        vault_container.stop()
+        shutil.rmtree(REFS_HOME, ignore_errors=True)
```
uberspot: Remember to clear `VAULT_ROOT_TOKEN` and all the other env vars to "" after tests run.

tests/test_vault.py
```diff
@@ -0,0 +1,162 @@
...
+            "I am a file with {}".format(secret['some_random_value']), revealed
+        )
+
+    def test_vault_base64_write_reveal(self):
+        '''
+        Write secret, confirm secret file exists, reveal and compare content
+        '''
+        tag = '?{vaultkv:secret/batman}'
+        secret = {
+            'some_random_value':'somethin_secret'
```
uberspot: something_secret*

tests/test_vault.py
```diff
@@ -0,0 +1,162 @@
...
+        # confirming secret file exists
+        self.assertTrue(os.path.isfile(os.path.join(REFS_HOME,'secret/batman')),
+                        msg="Secret file doesn't exist")
+        file_with_secret_tags = tempfile.mktemp()
+        with open(file_with_secret_tags,'w') as fp:
+            fp.write('I am a file with {}'.format(tag))
+        revealed = REVEALER.reveal_raw_file(file_with_secret_tags)
+        # confirming secerts are correctly revealed
+        self.assertEqual(
+            "I am a file with {}".format(secret['some_random_value']), revealed
```
uberspot: Change message to `File contents revealed: `

kapitan/cli.py
vaibhavk: That wasn't indented correctly, thanks for pointing that out.

kapitan/refs/secrets/vaultkv.py
homuncculus: `verify` is an `hvac.Client` parameter that expects either a boolean to indicate whether TLS verification should be performed when sending requests to Vault, or a string pointing at the CA bundle to use for verification.
I have added appropriate comments in the function.

vaibhavk: `verify` is an `hvac.Client` parameter that expects either a boolean to indicate whether TLS verification should be performed when sending requests to Vault, or a string pointing at the CA bundle to use for verification.
I have added appropriate comments in the function.

kapitan/refs/secrets/vaultkv.py
vaibhavk: There is a `client.is_authenticated()`at the end of the function that I expected to handle the authentication failure.

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,295 @@
...
+        client.auth_approle(
+            os.getenv('VAULT_ROLE_ID'),
+            secret_id=os.getenv('VAULT_SECRET_ID')
+        )
+    elif auth_type == 'github':
+        client.auth.github.login(token=env['token'])
+    else:
+        raise "Authentication type '{auth}' not supported".format(auth=auth_type)
+
+    assert (
```
uberspot: Change this to a normal `if` check. Under certain cases asserts are stripped from optimized binaries and this check might not be there in the end. 
Throw a VaultError if the `if` isn't authenticated().

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,296 @@
...
+        self.ref_type = ref_type
+        self.target_name = target_name
+        self.inventory_path = inventory_path
+
+    def __getitem__(self, ref_path):
+        full_ref_path = os.path.join(self.path, ref_path)
+        ref = self.ref_type.from_path(full_ref_path, target_name=self.target_name,
+                                      inventory_path=self.inventory_path)
+
+        if ref is not None:
```
uberspot: Change to `if ref:` That should hopefully work too?

tests/test_vault.py
```diff
@@ -0,0 +1,164 @@
...
+REFS_HOME = tempfile.mkdtemp()
+REF_CONTROLLER = RefController(REFS_HOME,target_name='minikube-mysql',
+                               inventory_path=os.path.join('.','examples','kubernetes','inventory'))
+REVEALER = Revealer(REF_CONTROLLER)
+# Create Vault docker container
+client = docker.from_env()
+env = {'VAULT_LOCAL_CONFIG':'{"backend": {"file": {"path": "/vault/file"}}, "listener":{"tcp":{"address":"0.0.0.0:8200","tls_disable":"true"}}}'}
+vault_container= client.containers.run(image='vault',cap_add=['IPC_LOCK'],ports={'8200':'8200'},
+                                       environment=env,detach=True,remove=True,command='server')
+sleep(4)
```
uberspot: Sleep here is quite unreliable to wait until the container is up. On different CI/CDs it might take longer than that and then the test will fail.
Use `docker inspect` in a loop to see if the container is actually in Ready state, and if not, sleep 1s again each time.
The docker bindings also give you the "status" https://docker-py.readthedocs.io/en/stable/containers.html so you can use that too to see if the container is Ready. :)

vaibhavk: This makes more sense, thank you, I actually changed it from 1s -> 4s (just in case) but checking the status is generic method :)

kapitan/refs/base.py
```diff
@@ -355,13 +355,15 @@ def __init__(self, content, content_type):
 
 
 class RefController(object):
-    def __init__(self, path):
+    def __init__(self, path,target_name = '',inventory_path = './inventory'):
```
uberspot: I had another look at this PR and compared with the GPG implementation for secrets. Actually you don't need to add `target_name` here at all. 
If you look at `kapitan/kapitan/refs/secrets/gpg.py` in the line `target_name = ref_params.kwargs['target_name']` it's actually getting the target_name from the inventory since it already exists there. 
You could easily do the same in your vaultkv class and get it that way instead of adding this here.

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,295 @@
...
+                "type": self.type_name}
+
+class VaultBackend(Base64RefBackend):
+    def __init__(self, path, target_name, inventory_path, ref_type=VaultSecret):
+        "init VaultBackend ref backend type"
+        super().__init__(path, ref_type)
+        self.type_name = 'vaultkv'
+        self.ref_type = ref_type
+        self.target_name = target_name
+        self.inventory_path = inventory_path
```
uberspot: Where is this inventory_path actually used? Do we need to set it here and in `RefController` or is it a leftover after changes?

kapitan/targets.py
```diff
@@ -314,6 +314,7 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, **kwa
     compile_objs = target_obj["compile"]
     ext_vars = target_obj["vars"]
     target_name = ext_vars["target"]
+    ref_controller.target_name = target_name
```
uberspot: This can probably be removed once you remove target_name from RefController as well.

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,295 @@
...
+
+class VaultSecret(Base64Ref):
+    """
+    Hashicorp Vault support for KV Secret Engine
+    """
+    def __init__(self, data, **kwargs):
+        """
+        Set vault parameter and encoding of data
+        """
+        self.data = data.rstrip()
```
uberspot: How come you `rstrip()` data here? Can you add a comment above and also explain what the data format is at that point?

kapitan/cli.py
```diff
@@ -620,6 +648,11 @@ def secret_update_validate(args, ref_controller):
                         secret_obj.update_key(awskey)
                         ref_controller[token_path] = secret_obj
 
+            elif token_path.startswith("?{vaultkv:"):
+                if not vaultkv:
+                    logger.debug("secret_update_validate: target: %s has no inventory vaultkv parameter 'auth', skipping %s", target_name, token_path)
+                    continue
+
```
uberspot: I might be mistaken, but should `secret_obj = ref_controller[token_path]` be set at this point? Same as in the cases of awskms/gkms/gpg above? Just warning if the key doesn't exist seems like it could lead to several *silent* bugs later on.

tests/test_cli.py
```diff
@@ -381,6 +383,84 @@ def test_cli_secret_subvar_gpg(self):
         self.assertEqual(expected.format("hello", "world"), stdout.getvalue())
         os.remove(test_tag_file)
 
+    @patch.object(vaultkv.VaultSecret,'_decrypt')
+    def test_cli_secret_write_vault(self,mock_reveal):
+        """
+        run $ kapitan refs --write vault:test_secret
```
uberspot: `vaultkv:` instead of `vault:`

tests/test_cli.py
```diff
@@ -381,6 +383,84 @@ def test_cli_secret_subvar_gpg(self):
...
+        self.assertEqual("revealing: {value}".format(value=test_secret_content_value),
+                         stdout.getvalue())
+ 
+        os.remove(test_tag_file)
+ 
+    @patch.object(vaultkv.VaultSecret,'_decrypt')
+    def test_cli_secret_write_base64_vault(self,mock_reveal):
+ 
+        """
+        run $ kapitan refs --write vault:test_secret --base64
```
uberspot: Same here.

examples/kubernetes/inventory/classes/component/mysql.yml
```diff
@@ -44,3 +48,5 @@ parameters:
 
         password_subvar: ?{gpg:targets/${target_name}/mysql/subvars@var1.password}
         password_sha256_subvar: ?{gpg:targets/${target_name}/mysql/subvars@var2.password_sha256}
+
+        password_vaultkv: ?{vaultkv:targets/${target_name}/mysql/vault_secret}
```
uberspot: One small observation about this. 
This will require all the VAULT related env vars to be set and a vault backend to be present in order for this line to be compiled. Which means that from a previous state of people being able to compile all examples, now they'll get an error in this one. 
I'd instead recommend you add this here in a comment just for reference.
And to keep your tests working, just introduce this line again in your tests setup() function (?). 
It would be best to keep the examples working for the majority of the users. :)

kapitan/refs/base.py
vaibhavk: The difference between the GPG & vaultkv  implementation is that I was asked not to store the metadata or the parameters in the secret file & instead get those from the inventory.
I couldn't find a method to get `target_name` and `inventory_path` once we have compiled the inventory. These parameters were required to load the vaultkv's `parameters` for revealing the secrets using the appropriate target's parameters.

kapitan/refs/secrets/vaultkv.py
vaibhavk: `inventory_path` will be used when we are revealing the secrets, this was actually useful only for testing purpose till now.

uberspot:
I tested this PR locally btw and no regression popped up. 
Two "nice to have" remain:
 - Use target_name from the inventory in the vault class instead of changing the RefBackend constructor (@ramaro might have a tip for that)
 - Test vaultkv properly in the CI/CD since now the vaultkv: secret isn't in the examples/ so tests don't run on it in Travis. 

Other than that the PR looks good. 
I'll approve but ideally would like a second approval from @ramaro or @adrianchifor before merging. :)

kapitan/cli.py
```diff
@@ -202,6 +202,9 @@ def main():
     refs_parser.add_argument('--key', '-K', help='set KMS key',
                              default=from_dot_kapitan('refs', 'key', ''),
                              metavar='KEY')
+    refs_parser.add_argument('--auth', help='set authentication type for vaultkv secrets',
```
ramaro: rename to `--vault-auth`?

docs/kap_proposals/kap_6_hashicorp_vault.md
```diff
@@ -35,6 +35,8 @@ The entire string __"foo:hello"__ is base64 encoded and stored in the secret_ins
 data: Zm9vOmhlbGxvCg==  
 encoding: original  
 type: vaultkv  
+vault_client_param:
```
ramaro: rename to `vault_params`? `

kapitan/refs/secrets/vaultkv.py
```diff
@@ -0,0 +1,253 @@
...
+                "'{key}' doesn't exist on '{path}'".format(key=data[1], path=data[0])
+            )
+        return return_data
+
+    def dump(self):
+        """
+        Returns dict with keys/values to be serialised.
+        """
+        return {"data": self.data, "encoding": self.encoding,
+                "type": self.type_name, "vault_client_param": self.vault_client_param}
```
ramaro: change `vault_client_param` to `vault_params`?

vaibhavk:
I'll add the `vaultkv` information on docs by making a new PR for `gh-pages` branch

uberspot:
> I'll add the `vaultkv` information on docs by making a new PR for `gh-pages` branch

That won't work unfortunately because of permissions. Ideally if you update the docs under docs/ on the master branch I can easily update the gh-pages part. :)

docs/kap_proposals/kap_6_hashicorp_vault.md
uberspot: One last one here @vaibhavk ^

Issue #194, opened 2019-01-22
Validate compile operation using jsonschema lib


uberspot:
Have kapitan use the jsonschema lib (if you enable it in your inventory). 
jsonschema lib will validate the compiled file output structure for k8s/terraform/whatever other yaml resources.
links:
https://json-schema.org/implementations.html
https://pypi.org/project/jsonschema/

pranavgupta1234:
Hi @uberspot , I would love to work on this issue under GSOC 2019. I have started exploring the Kapitan source and it would be great if you can point me in some specific direction to begin with.

pranavgupta1234:
Hey @uberspot, what tasks are involved in validating the file output structure ? Currently I think that we are required to define a jsonschema which will define properties for compiled files like their type, location, etc.

Issue #258, opened 2019-04-02
Support to validate the inventory before compiling


maheswarasunil:
**Support to validate the inventory before compiling**
   We are trying to explore the options for inventory value validations. Example 
In inventory 
```
 mysql:
    storage: 10G
    storage_class: standard
    image: mysql:latest  
```
we like to validate the value of 10G before compile. 
As jsonnet is not having many libraries to support like regex validations etc. We are not able to achieve this. 
  
It would be great if we can have validation support.

ramaro:
Fixed in #380 with jsonschema support

PR #317, opened 2019-07-05
Add kubernetes manifest validation

yoshi-1224:
(Partly) fixes issue #194

## Proposed Changes
  - Add support for validation of compiled k8s manifests using json schemas from `https://kubernetesjsonschema.dev`
  - Can be extended for other manifests/compiled output by extending the base `validator` class. E.g. we might at the same time tackle #258


yoshi-1224:
@uberspot @ramaro @adrianchifor 
I have two options in mind regarding how `validate` should be specified in the inventory:

#### Option 1

```yaml
kapitan:
   validate:
      - output_path: <corresponding_path_in_kapitan.compile>
        type: kubernetes
        kind: <kind>
        version: <kubectl-version>
```

#### Option 2
I think it is also possible to put validate options under `kapitan.compile`, such as 

```yaml
kapitan:
   compile:
      - output_path: <output_path>
        input_paths:
           - <input_path>
        validate: kubernetes
        kind: <kind>
        version: <kubectl-version>
```

But option 2 is not so great because it doesn't work well when one jsonnet file compiles to multiple files, or when there are more than one `input_paths` (or when `input_path` is a directory).

uberspot:
> @uberspot @ramaro @adrianchifor
> I have two options in mind regarding how `validate` should be specified in the inventory:
> 
> #### Option 1
> ```yaml
> kapitan:
>    validate:
>       - output_path: <corresponding_path_in_kapitan.compile>
>         type: kubernetes
>         kind: <kind>
>         version: <kubectl-version>
> ```
> 
> #### Option 2
> I think it is also possible to put validate options under `kapitan.compile`, such as
> 
> ```yaml
> kapitan:
>    compile:
>       - output_path: <output_path>
>         input_paths:
>            - <input_path>
>         validate: kubernetes
>         kind: <kind>
>         version: <kubectl-version>
> ```
> 
> But option 2 is not so great because it doesn't work well when one jsonnet file compiles to multiple files, or when there are more than one `input_paths` (or when `input_path` is a directory).
> 
> If we are going with Option 1, ~optimally we would want to validate while the object is still in memory, but it can be a bit complicated to do the matching of output_paths. Is it ok if I read the files written to `compile_path` (i.e. the temporary path before copy to the actual path) to validate them?~ we can first create hashmap of the output_paths in `kapitan.validate`, and then after each `render` operations in jsonnet and so on, we can find the corresponding validation option for that object, if exists.
> 
> If there is a better way over both options, feel free to let me know!

Option 1 does seem more versatile.
It makes sense to do Validation as a separate step and NOT combining it with compiling. Semantically it's a separate operation. And if it's implemented separately we can also in the future split it to a separate command `kapitan validate`, if it's something people want to use. :) 

All compiled/ outputs should be generated by kapitan compile anyway (and non valid files are deleted after compilation). So I'd assume it's safe to just run schema validation on the compiled target directory directly (assuming those outputs are k8s manifests in this case).

kapitan/validator/kubernetes_validator.py
```diff
@@ -0,0 +1,81 @@
...
+        logger.debug("Fetching schema from {}".format(url))
+        content, _ = make_request(url)
+        if content is None:
+            raise RequestUnsuccessfulError("schema failed to fetch from {}"
+                                           "\nThe specified version or kind may not be supported".format(url))
+        logger.debug("schema fetched  from {}".format(url))
+        return yaml.safe_load(content)
+
+    def _get_request_url(self, kind, version):
+        return 'https://kubernetesjsonschema.dev/' + self.file_path_format.format(version, kind)
```
uberspot: Maybe put this URL in a Constant in this file as well like you did for `DEFAULT_VERSION` ? 
It's best to keep these sort of constants near each other in code and not scattered to make debugging easier. :)

tests/test_kubernetes_validator.py
```diff
@@ -0,0 +1,70 @@
...
+from kapitan.validator.kubernetes_validator import KubernetesManifestValidator
+
+
+class KubernetesValidatorTest(unittest.TestCase):
+    def setUp(self):
+        self.cache_dir = tempfile.mkdtemp()
+        self.validator = KubernetesManifestValidator(self.cache_dir)
+
+    def test_download_and_cache(self):
+        downloaded_schema = self.validator._get_schema_from_web('service', '1.14.0')
```
uberspot: Ideally re-use the Constant you defined already in `kubernetes_validator.py` so when we increase that version we only change it in one file.

tests/test_kubernetes_validator.py
```diff
@@ -0,0 +1,70 @@
...
+        self.validator._cache_schema('service', '1.14.0', downloaded_schema)
+        self.assertTrue(os.path.isfile(os.path.join(self.cache_dir, 'v1.14.0-standalone-strict', 'service.json')))
```
uberspot: Same for the `standalone-strict` here. I'd keep it in a constant and just re-use it everywhere.

tests/test_kubernetes_validator.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def test_load_from_cache(self):
+        kind = 'deployment'
+        version = '1.11.0'
```
uberspot: How come the version here is different?

tests/test_kubernetes_validator.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+
+        service_manifest = yaml.safe_load(service_manifest_string)
+        self.validator.validate(service_manifest, kind='service', version='1.14.0')
+
+        with self.assertRaises(KubernetesManifestValidationError):
+            self.validator.validate(service_manifest, kind='deployment', version='1.14.0',
+                                    file_path='service/manifest', target_name='example')
+
+    def test_compile_with_validate(self):
```
uberspot: `test_validate_after_compile` ? Given the discussion in the main thread of this PR? :P

yoshi-1224:
@uberspot 
Thanks for the quick reply! I'll just run the validation on the compiled output dir then, after these lines in `compile_targets.py`:

```python
# compile done in parallel above
shutil.rmtree(compile_path)
shutil.copytree(temp_path, compile_path)
logger.debug("Copied %s into %s", temp_path, compile_path)
```

kapitan/validator/kubernetes_validator.py
```diff
@@ -0,0 +1,81 @@
...
+
+DEFAULT_VERSION = '1.14.0'
+
+logger = logging.getLogger(__name__)
+
+
+class KubernetesManifestValidator(Validator):
+    def __init__(self, cache_dir, **kwargs):
+        super().__init__(cache_dir, **kwargs)
+        self.schema_type = 'standalone-strict'
```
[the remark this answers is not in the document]
yoshi-1224: Yup I'll add this link https://github.com/garethr/kubernetes-json-schema#kubernetes-json-schemas

tests/test_kubernetes_validator.py
yoshi-1224: I was just using a different version just to increase test coverage. I'll just explain that in a comment!

ramaro:
How about we detach validation from compile by default? I'm worried compilation will become too slow.

I propose we add a new `$ kapitan validate` command which will do this instead. We might also add (similar to what was done for deps) a `validate_force` inventory option to run validate on compile and even perhaps a flag: `$ kapitan compile --validate`.

I believe this will make it more versatile - but just a thought!

yoshi-1224:
@uberspot, @ramaro 
I've created a separate command `kapitan validate`, which takes in

```
--schemas-path: defaults to ./schemas. If doesn't exist, will be created
--targets: same as compile i.e. targets to validate
--inventory-path: same as compile, defaults to ./inventory
--compiled-path: same as compile, defaults to ./compiled
-p: same as compile
```

`kapitan compile` now has `--validate` flag as well as `--schemas-path` (semantically the same as described above).

yoshi-1224:
@uberspot 
Actually I realized that by making this process parallel as it is now, we bump into the same issue we had with external dependencies fetch (e.g. imagine a situation where two different targets use the same schema for validation). Let me change the code in such a way that each `kubernetes_validator` corresponds to a unique pair of `(kind, version)` with a list of files (across different targets) that need to be validated against it, such that we can safely perform validation without race condition.

yoshi-1224:
@uberspot @ramaro 
PR is ready for review, thanks!

README.md
```diff
@@ -637,6 +639,41 @@ $ kapitan searchvar parameters.elasticsearch.replicas
 ./inventory/classes/component/elasticsearch.yml   1
 ```
 
+### kapitan validate
+
+Validates the schema of compiled output (currently supports Kubernetes manifests).
+
+Refer to the `minikube-es` inventory in [kapitan inventory](#kapitan-inventory). To validate the schema of the compiled StatefulSet manifest at `compiled/minikube-es/manifests/es-client.yml` (created by `components/elasticsearch/maiin.jsonnet`), add `kapitan.validate` parameters in `minikube-es` inventory.
```
uberspot: Typo in `maiin.jsonnet`

Issue #333, opened 2019-08-02
KAP-5 - Implement plain ref type, revamp secrets into refs


ramaro:
See [KAP-5](https://github.com/deepmind/kapitan/blob/master/docs/kap_proposals/kap_5_ref_types_redesign.md)

uberspot:
This has been fixed in code, but needs to be announced in all kapitan channels when the feature reaches a non-rc build.

ramaro:
all done since 0.25

PR #334, opened 2019-08-02
Kap5 ref types

ramaro:
Implements  #333


ademariag:
Could we merge this to master so I can get it to test it properly?
or a :dev branch?

uberspot:
It is a breaking change so i'd avoid merging before a complete review. If you want to test it locally, you can clone ramaro's branch and run it from there with `./bin/kapitan`. :)

kapitan/cli.py
```diff
@@ -429,20 +436,32 @@ def secret_write(args, ref_controller):
         tag = '?{{awskms:{}}}'.format(token_path)
         ref_controller[tag] = secret_obj
 
-    elif token_name.startswith("ref:"):
+    elif token_name.startswith("base64:"):
```
uberspot: This will break secret parding for all old formats right? Woukd it make sense to support the old type here too? Maybe at least print a warning message for the user so they know what to do/how to upgrade?

kapitan/refs/base.py
```diff
@@ -99,6 +91,54 @@ def from_params(cls, data, ref_params):
...
             return cls(data, **ref_params.kwargs)
 
+    def dump(self):
+        """
+        Returns dict with keys/values to be serialised.
+        """
+        return {"data": self.data.decode(), "encoding": self.encoding,
+                "type": self.type_name}
+
+
```
uberspot: Would it make sense to split these classes to separate files? They're more implementations of the base classes than actual "base.py"(?).

ramaro: makes sense, will work on it

kapitan/cli.py
ramaro: I'm thinking it might be better to provide a migration script. Sure, we could support `ref` but also other things will change and break, like renaming `--secrets-path` into `--refs-path` and the default value from `./secrets/` to `./refs/`. So I believe it's wiser to break and provide a clear migration script.

ademariag:
PING? :)

uberspot:
Merged, I'd like to wait a tiny bit before an rc release. Till https://github.com/deepmind/kapitan/issues/358 is fixed. :)

Issue #345, opened 2019-08-23
Add standalone binary to released files


uberspot:
After https://github.com/deepmind/kapitan/pull/323 we have a process to generate a standalone binary of kapitan and it's built in travis too.
The produced binary lives in dist/
It would be great to also include that to the files released by travis on each release.
i.e. in .travis.yml we already do that with CHANGELOG.md
```
deploy:
  - provider: releases
    api_key:
      secure: blabla=
    file: CHANGELOG.md
    prerelease: $PRERELEASE
    on:
      tags: true
      repo: deepmind/kapitan
```

yoshi-1224:
@uberspot
I may be misunderstanding, but is this issue already addressed in #349?

uberspot:
Yes it is. :) i just made an issue to track that work. I think the binary is now on the releases page on github and every build works fine on travis so i'd consider this issue resolved for now. :)

PR #349, opened 2019-08-23
Attempt #3 to fix travis pypi deployment, bin/kapitan artifacts are moved to bindist/

uberspot:
In an effort to fix this https://github.com/deepmind/kapitan/issues/345 
I added dist/kapitan-linux-amd64 to the list of artifacts that will be uploaded in `.travis.yml`

The build fails with https://travis-ci.org/deepmind/kapitan/builds/575917567 
```
Deploying application
dist/kapitan-linux-amd64 already exists, no checkout
kapitan/__pycache__/__init__.cpython-37.pyc already exists, no checkout
kapitan/__pycache__/version.cpython-37.pyc already exists, no checkout
Could not restore untracked files from stash entry
PyPI upload failed.
failed to deploy
```
For which the dist/ binary stays there because of the `warning: failed to remove dist/kapitan-linux-amd64: Permission denied` error just above in the build error.

The other .pyc artifacts might be another thing that still should be cleaned after the build. 

Trying to see if this MR fixes anything

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.bak
 build/
 dist/
+bindist/
 _site/
 site/
 kapitan.egg-info/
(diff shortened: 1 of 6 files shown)


yoshi-1224:
@uberspot 
Thanks for this. I believe this is to do with the binary created inside docket image as root whose permission is passed on to the host.

PR #356, opened 2019-08-28
Update instruction for gh-pages deploy

yoshi-1224:
My apologies for another minor documentation update.

When contributors want to make a PR for gh-pages, they must do it from a branch that shares commit history with our gh-pages, or Github would throw `There isn't anything to compare` message and prevent PRs to be created. This means we cannot create some new branch using `mkdocs gh-deploy` but should rather commit to a branch that is tracking our gh-pages branch. I added a note on this.


ramaro:
Good spot! thanks @yoshi-1224

PR #359, opened 2019-09-03
Shift to debian image

yoshi-1224:
This PR serves more as a suggestion.

The Dockerfile in this PR is changed to `FROM python:3.7-slim-stretch`. This was chosen because it is one of the smaller debian images which have python 3.7 installed by default (the custom image is named as kapitan-debian below).

```
$ docker image ls
REPOSITORY           TAG                 IMAGE ID            CREATED             SIZE
kapitan-debian       latest              728041763c54        2 hours ago         1.04GB
deepmind/kapitan     latest              0670fbc69cde        4 days ago          860MB
python               3.7-alpine          39fb80313465        7 days ago          98.7MB
python               3.7-slim-stretch    7f8ea8958a02        7 days ago          155MB
```

I think the image size difference is not so bad (~ 160MB) given that musl-libc is going to be smaller than glibc to start with. 

So far the image is working just fine (tested with the examples folder, plus helm binding works as well). Got `decryption failed` error when using --reveal option but I also get that using `deepmind/kapitan` image so I may just be doing it wrongly.

If this is acceptable, we can move on to change the base image of `Dockerfile.ci` as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,21 +1,27 @@
-FROM python:3.7-alpine
+FROM python:3.7-slim-stretch
 
 RUN mkdir /kapitan
 WORKDIR /kapitan
 COPY kapitan/ kapitan/
 COPY requirements.txt ./
 
-RUN apk add --no-cache --virtual build-dependencies g++ make musl-dev && \
-    apk add --no-cache libstdc++ gnupg yaml-dev libffi-dev openssl-dev git && \
-    pip install --upgrade pip && \
-    pip install -r requirements.txt && \
-    apk del build-dependencies && \
-    rm -r /root/.cache
+RUN apt-get update && apt-get install -y \
+    build-essential git wget \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && pip install --upgrade pip \
+    && pip install -r requirements.txt
 
-# install additional dependencies required to build helm binding
-RUN apk add --no-cache git bash go libc-dev && \
+# build go from source
+RUN wget https://dl.google.com/go/go1.12.7.linux-amd64.tar.gz -q && \
+    tar -C /usr/local -xvf go1.12.7.linux-amd64.tar.gz && \
+    rm go1.12.7.linux-amd64.tar.gz
+
+# build helm binding
+RUN export PATH=$PATH:/usr/local/go/bin && \
     chmod +x kapitan/inputs/helm/build.sh && \
-    ./kapitan/inputs/helm/build.sh
+    ./kapitan/inputs/helm/build.sh && \
+    rm /usr/local/go -rf
 
 ENV PYTHONPATH="/kapitan/"
 ENV SEARCHPATH="/src"
--- a/kapitan/inputs/helm/__init__.py
+++ b/kapitan/inputs/helm/__init__.py
@@ -65,11 +65,15 @@ class Helm(InputType):
 
 def render_chart(chart_dir, output_path, **kwargs):
     """renders helm chart located at chart_dir, and stores the output to output_path"""
+    binding_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "libtemplate.so")
+    if not os.path.exists(binding_path):
+        raise HelmBindingUnavailableError("Helm binding is not available. Run 'make build_helm_binding' to create it")
     try:
         # lib is opened inside the function to allow multiprocessing
-        lib = ffi.dlopen(os.path.join(os.path.dirname(os.path.abspath(__file__)), "libtemplate.so"))
-    except (NameError, OSError):
-        raise HelmBindingUnavailableError("Helm binding is not available. Run 'make build_helm_binding' to create it")
+        lib = ffi.dlopen(binding_path)
+    except (NameError, OSError) as e:
+        raise HelmBindingUnavailableError('There was an error opening helm binding. '
+                                          'Refer to the exception below:\n' + str(e))
 
     if kwargs.get('helm_values_file', None):
         helm_values_file = ffi.new("char[]", kwargs['helm_values_file'].encode('ascii'))
--- a/kapitan/inputs/helm/build.sh
+++ b/kapitan/inputs/helm/build.sh
@@ -17,4 +17,4 @@ else
     echo "$so_name is missing. Exiting"
     exit 1
 fi
-python cffi_build.py
+python3 cffi_build.py


ademariag:
I think we should also fix `Dockerfile.ci`

kapitan/inputs/helm/__init__.py
```diff
@@ -18,6 +19,21 @@ def __init__(self, compile_path, search_paths, ref_controller):
         super().__init__("helm", compile_path, search_paths, ref_controller)
         self.helm_values_file = None
         self.helm_params = {}
+        self.lib = self.initialise_binding()
+
+    def initialise_binding(self):
+        """returns the dl_opened library (.so file) if exists, otherwise None"""
+        if platform.system() not in ('Linux', 'Darwin'):  # TODO: later add binding for Mac
+            return None
+        binding_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "libtemplate.so")
```
uberspot: Add a small comment above this to say what path this builds exactly (where it's pointing too). 
So it's intuitive when changing this in the future for OSX. :)

kapitan/inputs/helm/__init__.py
```diff
@@ -18,6 +19,21 @@ def __init__(self, compile_path, search_paths, ref_controller):
...
+        if not os.path.exists(binding_path):
+            return None
```
uberspot: Print logger.debug message here if path doesn't exist.

kapitan/inputs/helm/template.go
```diff
@@ -301,7 +301,7 @@ func writeToFile_c(outputDir string, name string, data string) error {
 		return err
 	}
 
-	// fmt.Printf("wrote %s\n", outfileName)
```
uberspot: Remove comment or change to Logger.debug. It's ok if it prints it as Debug output. :)

yoshi-1224: I'll remove the comment since the bug is already fixed!

yoshi-1224:
@uberspot 
Should we change the name of the helm binding shared object according to which platform this is targeted, just as we did for kapitan binary? In my understanding @ramaro wants the so distributed for each of Linux and Mac. I'm working on cross compile for OSX now, which hopefully takes us somewhere

uberspot:
> @uberspot
> Should we change the name of the helm binding shared object according to which platform this is targeted, just as we did for kapitan binary? In my understanding @ramaro wants the so distributed for each of Linux and Mac. I'm working on cross compile for OSX now, which hopefully takes us somewhere

Sure, go for it! 
Just don't forget to update the docs too if there's any mention of the .so there too. :)

ademariag:
Is there an update on this?

yoshi-1224:
Hi @ademariag 
Sorry I'd been busy and couldn't really get on this task until now.

@uberspot 
Could you please have a look at the new the docker CI image and confirm whether everything that was there on the previous alpine version is available as before?

Dockerfile.ci
```diff
@@ -31,5 +52,7 @@ ENV KBLD_URL=https://github.com/k14s/kbld/releases/download/v0.11.0/kbld-linux-a
 RUN curl -L -o /usr/local/bin/kapp ${KAPP_URL} && chmod +x /usr/local/bin/kapp
 RUN curl -L -o /usr/local/bin/kbld ${KBLD_URL} && chmod +x /usr/local/bin/kbld
 
-COPY --from=jsonnet /usr/local/bin/jsonnet /usr/local/bin/
-COPY --from=promtool /go/bin/promtool /usr/local/bin/
+COPY --from=go-env /go/bin/promtool /usr/local/bin/
+
+COPY tests/test_resources /examples
```
yoshi-1224: This tests line is going to be removed later.

Dockerfile
```diff
@@ -1,21 +1,27 @@
-FROM python:3.7-alpine
+FROM python:3.7-slim-stretch
```
taharah: We can utilize multi-stage builds here to end up with a "cleaner" final image as shown below. The same is probably possible for the CI image, but I haven't looked at that yet.

```
# First build the Helm binding
FROM golang:1.12.9-stretch AS go-builder

RUN mkdir /kapitan
WORKDIR /kapitan
COPY . .

RUN ./kapitan/inputs/helm/build.sh

# Build the virtualenv for Kapitan
FROM python:3.7-slim-stretch AS python-builder

COPY --from=go-builder /kapitan /kapitan
WORKDIR /kapitan

ENV PATH="/opt/venv/bin:${PATH}"

RUN apt-get update \
  && apt-get install --no-install-recommends -y build-essential gcc \
  && python -m venv /opt/venv \
  && pip install --upgrade pip setuptools wheel \
  && pip install -r requirements.txt \
  && ./kapitan/inputs/helm/build.sh \
  && pip install .

# Final image with virtualenv built in previous step
FROM python:3.7-slim-stretch

COPY --from=python-builder /opt/venv /opt/venv

ENV PATH="/opt/venv/bin:${PATH}"
ENV SEARCHPATH="/src"
VOLUME ${SEARCHPATH}
WORKDIR ${SEARCHPATH}

# Install runtime dependencies and run as a non-root user for good measure
RUN apt-get update \
  && apt-get install --no-install-recommends -y git \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* \
  && useradd --create-home --no-log-init --user-group kapitan

USER 1000

ENTRYPOINT ["kapitan"]
```

The build script just needs to be modified as shown below in order to allow it to work for the multi-stage builds:

```sh
#!/usr/bin/env bash

cd $(dirname "$0")
pwd
so_name=libtemplate.so

# Compile the binding if a Go runtime exists
if [[ -z $(which go) ]]; then
    echo "[WARN] go is not available on this system -- skipping Helm binding build!"
else
    go build -buildmode=c-shared -o $so_name template.go
fi

# Validate that the compiled binding exists
if [[ -e $so_name ]]; then
    echo "$so_name built successfully"
else
    echo "[ERROR] $so_name does not exist!"
    exit 1
fi

# Compile the Python ffi binding if Python is available
if [[ -z $(which python3) ]]; then
    echo "[WARN] python3 is not available on this system -- skipping cffi build!"
else
    echo "Building the Python binding using cffi"
    python3 cffi_build.py
fi

# We made it! Time to exit cleanly...
exit 0
```

uberspot: The multi-stage build would be awesome and we should definitely add it, but ideally in a separate PR since it's semantically separate from the current one. :) 
Feel free to open a PR after this is merged or I'll have a look at it after these are fixed and a new RC is out. :+1:

taharah: Makes sense! I'll open the PR when possible.

yoshi-1224:
@uberspot 
what would be the most convenient way to build and distribute helm binding for Mac in addition to the existing Linux binding? Is cross-compiling on Travis the way, or do you think there might be an easier path (e.g. is it possible to build natively on OSX machine on Travis from which we can copy the built shared object file onto our Debian one)?

ademariag:
May I suggest that we look at the helm binding in mac in another PR? I feel this one if getting more than it came for. This way we can unlock testing on the current state.

yoshi-1224:
@ademariag 
I totally agree. That way @taharah can get started on the multistage Docker image(s) too. @uberspot Let me work on OSX binding in another PR. I'm going to make this one ready for review.

Issue #341, opened 2019-08-21
Base64 encoded secrets are not revealed


taharah:
**Describe the bug/feature**
When using an encrypted value via Kapitan secrets to create a Kubernetes secret, if the encrypted value is nested within a document, then the output will include the base64 encoded reference to the secret.

**To Reproduce**
1. Encrypt a value with Kapitan secrets
2. Create a jsonnet manifest that base64 encodes a document that has the value embedded and compile.
3. Decode the base64 string and you will find the embedded reference to the Kapitan secret.

**Expected behavior**
I would expect the output to include the decrypted value.

**Screenshots**
If applicable, add screenshots to help explain your problem/idea.

** If it's a bug (please complete the following information):**
 - `python --version`: 3.7.4
 - `pip3 --version`: 19.2.2
 - `Are you using pyenv or virtualenv?` virtualenv

**Additional context**

ramaro:
@taharah running base64 in jsonnet on a secret won't work because it will base64 the reference and not the content as you say.

I suggest you remove the base64 function call from jsonnet and create your secret instead with:
`$ kapitan secrets --write gpg:my/secret -f value.txt --base64`.

The `--base64` flag here will encode the content before encrypting it, which will have the desired effect.

I believe this is what is happening, but it would be great if you can provide some snippets.

taharah:
I have used the flag `--base64` previously for individual values that I am trying to protect; however, the issue is when I need to only protect a small portion of a configuration file, e.g., an API key. In this case, I would need to encrypt the entire file, which makes updating and maintaining the file far more difficult. Additionally, since Kapitan does not currently use envelope encryption, at least for KMS, files are limited in size to 4 KB.

uberspot:
How about templating the configuration file as well? Then you can just add the api key you mention as a secret reference in that template.

ramaro:
Sorry, can you share some snippets to help understand your problem clearly?

taharah:
For example, if I have something like the following in my inventory:

```yaml
parameters:
  alertmanager:
    configuration:
      resolve_timeout: 5m
      opsgenie_api_key: ${gpg:targets/${target_name}/opsgenie/api.key}
```

Then try to create a secret file from the entire configuration blob like shown below:

```jsonnet
{
  secret: kube.Secret(name) {
    data: {
      config.yml: std.base64(
        std.manifestYamlDoc(inv.parameters.alertmanager.configuration)
      ),
    },
  },
}
```

[...]

PR #363, opened 2019-09-07
Resolve and compile refs when inventory is cached

taharah:
Signed-off-by: Trevor Wood <Trevor.G.Wood@gmail.com>

Fixes issue #341. 

## Proposed Changes

* Resolves and compiles all refs when the inventory is cached rather than doing a find and replace in the final compiled files.

### DIFF
--- a/kapitan/resources.py
+++ b/kapitan/resources.py
@@ -275,7 +275,8 @@ def inventory_reclass(inventory_path):
             class_mappings = reclass_config.get("class_mappings")  # this defaults to None (disabled)
             _reclass = reclass.core.Core(storage, class_mappings, reclass.settings.Settings(reclass_config))
 
-            cached.inv = _reclass.inventory()
+            # Reveal or compile refs if necessary
+            cached.inv = reveal_or_compile_refs(_reclass.inventory())
         except ReclassException as e:
             if isinstance(e, NotFoundError):
                 logger.error("Inventory reclass error: inventory not found")
@@ -284,3 +285,30 @@ def inventory_reclass(inventory_path):
             raise InventoryError(e.message)
 
     return cached.inv
+
+
+def reveal_or_compile_refs(inv):
+    """
+    Reveals or compiles the reclass inventory inv for all targets specified if a Revealer has been instantiated.
+    Otherwise, it does not modify the inventory.
+    Returns a reclass style dictionary
+    """
+
+    # Return the unmodified inventory if there is no revealer_obj
+    if cached.revealer_obj is None:
+        return inv
+
+    # Check if any targets were specified
+    if cached.revealer_obj.targets:
+        targets_list = cached.revealer_obj.targets
+    else:
+        targets_list = inv["nodes"]
+
+    # Reveal or compile refs for all necessary targets
+    for target_name in targets_list:
+        if cached.revealer_obj.reveal:
+            cached.revealer_obj.reveal_obj(inv["nodes"][target_name])
+        else:
+            cached.revealer_obj.compile_obj(inv["nodes"][target_name], target_name=target_name)
+
+    return inv
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -36,9 +36,7 @@ from reclass.errors import NotFoundError, ReclassException
 logger = logging.getLogger(__name__)
 
 
-def compile_targets(
-    inventory_path, search_paths, output_path, parallel, targets, labels, ref_controller, **kwargs
-):
+def compile_targets(inventory_path, search_paths, output_path, parallel, targets, labels, **kwargs):
     """
     Searches and loads target files, and runs compile_target() on a
     multiprocessing pool with parallel number of processes.
@@ -73,13 +71,7 @@ def compile_targets(
 
         # append "compiled" to output_path so we can safely overwrite it
         compile_path = os.path.join(output_path, "compiled")
-        worker = partial(
-            compile_target,
-            search_paths=search_paths,
-            compile_path=temp_path,
-            ref_controller=ref_controller,
-            **kwargs,
-        )
+        worker = partial(compile_target, search_paths=search_paths, compile_path=temp_path, **kwargs)
 
         if not target_objs:
             raise CompileError("Error: no targets found")
@@ -318,15 +310,13 @@ def load_target_inventory(inventory_path, targets):
         targets_list = inv["nodes"]
 
     for target_name in targets_list:
-        try:
-            target_obj = inv["nodes"][target_name]["parameters"]["kapitan"]
-            valid_target_obj(target_obj)
-            validate_matching_target_name(target_name, target_obj, inventory_path)
-            logger.debug("load_target_inventory: found valid kapitan target %s", target_name)
-            target_objs.append(target_obj)
-        except KeyError:
-            logger.debug("load_target_inventory: target %s has no kapitan compile obj", target_name)
-            pass
+        # Validate the target name and get the target object
+        target_obj = validate_matching_target_name(target_name, inv, inventory_path)
+
+        # Validate the target object
+        valid_target_obj(target_obj)
+        logger.debug("load_target_inventory: found valid kapitan target %s", target_name)
+        target_objs.append(target_obj)
 
     return target_objs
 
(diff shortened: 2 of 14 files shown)


taharah:
@googlebot I signed it!

taharah:
I realized this change would not work in all use cases, e.g., when compiling only a specific of targets for which you have access to decrypt the secrets. I'll push the necessary changes to fix this use case soon.

kapitan/targets.py
```diff
@@ -491,12 +488,13 @@ def valid_target_obj(target_obj):
                              format(e.message, json.dumps(e.instance, indent=2, sort_keys=False)))
 
 
-def validate_matching_target_name(target_filename, target_obj, inventory_path):
+def validate_matching_target_name(target_filename, inv, inventory_path):
     """Throws *InventoryError* if parameters.kapitan.vars.target is not set,
     or target does not have a corresponding yaml file in *inventory_path*
```
uberspot: Update docstring to include the new return value. :)

kapitan/resources.py
```diff
@@ -250,3 +252,13 @@ def inventory_reclass(inventory_path):
             raise InventoryError(e.message)
 
     return cached.inv
+
+
+def reveal_or_compile_refs(inv):
```
uberspot: I like the consolidation of this behaviour to just one function instead of the previous duplication in code.
But if I understand this correctly it will always by default reveal when compiling ? If that's the case, that'd be a breaking change and it might break some user cases as well. 

I might be misinterpreting it, but isn't issue #341 already fixed in one of our examples?  
https://github.com/deepmind/kapitan/blob/master/examples/kubernetes/components/mysql/secret.jsonnet#L9
password declared here: https://github.com/deepmind/kapitan/blob/master/examples/kubernetes/inventory/classes/component/mysql.yml#L43 
I think you just need the same thing, but without the `|sha256` in the secret generation ? 
We already have several secret values in k8s secrets this way.

taharah: It should not be a breaking change since it will still reveal or compile the refs in the inventory based on the arguments given to the command. The reveal argument is passed to the Revealer when it is instantiated now.

However, I did realize there was one issue with the implementation this morning. I need to filter the reveal down to specific targets if they are specified. Otherwise, it will always try to reveal the entire inventory, which may not always be possible (e.g., I only have access to the decryption secrets for a subset of the targets). I'll push up the necessary changes soon.

The examples unfortunately do not resolve the issue. It works great when you only want to set single key/value pairs in a Secret; however, if I want to build an entire configuration file from the inventory data and only encrypt one value, then it is not possible to do so. In the current state, you either need to encrypt the entire file or nothing at all.

taharah: @uberspot I've updated the logic to include only revealing or compiling specified targets rather than always doing it for all targets. I've included some additional testing for the logic as well. 

Additionally, if you take a look at https://github.com/deepmind/kapitan/pull/363/files#diff-7626220eb9df789da70769fc68e3580aR313, you can see that the reveal argument is indeed passed to the Reveal object at instantiation. Then used here https://github.com/deepmind/kapitan/pull/363/files#diff-f524753f276883cad143588359f4fde7R276-R279 to flex between revealing and compiling.

tests/test_compile.py
```diff
@@ -70,25 +72,73 @@ def test_compile_not_matching_targets(self):
...
+    def test_inventory_reveal_or_compile_with_invalid_ref(self):
+        inventory_path = "inventory"
+        refs_path = "refs"
+
+        # Build inventory
+        ref_controller = RefController(refs_path)
+        cached.ref_controller_obj = ref_controller
+        cached.revealer_obj = Revealer(ref_controller, reveal=False)
+        inv = inventory_reclass(inventory_path)
+        inv["nodes"]["minikube-es"]["secret"] = "?{gkms:secret/value}"
```
uberspot: I have the impression this is always going to fail anyway since gkms isn't tested on Travis yet. 
Maybe gpg would be a saner option here.

taharah: It shouldn't matter since this test is for when a ref that is defined in the inventory cannot be found, i.e., it doesn't exist, there was a typo, etc.

uberspot:
I tested this branch locally with some templates/inventory we use internally for infrastructure. I just run `kapitan compile` and it fails to resolve the following from a class in the inventory:
```
.....
  some_secret: ?{gpg:common/some_key}
.....
``` 
I get `Could not find ref: ?{gpg:common/some_key}` as the only output. 

Is it maybe unable to find a __cached__ version of the reference? 

----------------
I also tested just revealing an existing secret:
```
~/code/kapitan/bin/kapitan refs --reveal -f ./compiled/my_target_name/manifests/my_component_configmap.yml
  File "~/code/kapitan/kapitan/cli.py", line 539, in ref_reveal
    for rev_obj in revealer.reveal_path(file_name):
  File "~/code/kapitan/kapitan/refs/base.py", line 170, in reveal_path
    content, content_type = self._reveal_file(path)
  File "~/code/kapitan/kapitan/refs/base.py", line 194, in _reveal_file
    rev_obj = self.reveal_obj(obj)
  File "~/code/kapitan/kapitan/refs/base.py", line 332, in reveal_obj
    obj = [self.reveal_obj(item) for item in obj]
  File "~/code/kapitan/kapitan/refs/base.py", line 332, in <listcomp>
    obj = [self.reveal_obj(item) for item in obj]
  File "~/code/kapitan/kapitan/refs/base.py", line 330, in reveal_obj
    obj[k] = self.reveal_obj(v)
  File "~/code/kapitan/kapitan/refs/base.py", line 330, in reveal_obj
    obj[k] = self.reveal_obj(v)
  File "~/code/kapitan/kapitan/refs/base.py", line 334, in reveal_obj
    obj = self.regex.sub(self._reveal_replace_match, obj)
  File "~/code/kapitan/kapitan/refs/base.py", line 239, in _reveal_replace_match
    return self._reveal_tag_without_subvar(tag)
  File "~/code/kapitan/kapitan/refs/base.py", line 258, in _reveal_tag_without_subvar
    ref = self.ref_controller[tag_without_subvar]
  File "~/code/kapitan/kapitan/refs/base.py", line 491, in __getitem__
    ref = self._get_from_token(token)
  File "~/code/kapitan/kapitan/refs/base.py", line 445, in _get_from_token
    ref = backend[path]
  File "~/code/kapitan/kapitan/refs/base.py", line 128, in __getitem__
    raise KeyError(ref_path)
KeyError: 'targets/my_target_name/my_component/some_secret'
```
----------------
I also tested on a target with only one secret in a template (jinja2) and just re-compiled that target without changing anything else and after this PR, it removed the hash from the secret in the output files.

compiled line before: 
`+export SOMETHING="$(echo -n '?{gkms:common/some_secret:41452404}' | kapitan secrets --reveal -f -)"

compiled line after: 
`+export SOMETHING="$(echo -n '?{gkms:common/some_secret}' | kapitan secrets --reveal -f -)"


------------

I hope these outputs help in debugging this a bit. :+1:

taharah:
Thank you for testing! I had been basing most of my assumptions on the functionality of the changes that the tests were passing without any modifications. I'll do some additional "real world" testing myself over the weekend to try and hammer out any remaining bugs. I'll also try and see if we can further harden the test suite to detect drifts in functionality better going forward.

taharah:
I've tried to recreate the errors @uberspot, but unfortunately I am unable to do so. I was only able to recreate the error when the `refs` directory did not exist. Otherwise, I always experience the expected behavior, i.e., the refs are either compiled or revealed.

uberspot:
Could it be due to caching? When i tried i hadn't compiled that project before so it could be related(?). Could you try in a container or another machine in case it's that?

taharah:
@uberspot I've rebased the change with the latest upstream changes. However, I still have yet to reproduce the error you posted and have been using the changes internally for several weeks now without issue.

uberspot:
> @uberspot I've rebased the change with the latest upstream changes. However, I still have yet to reproduce the error you posted and have been using the changes internally for several weeks now without issue.

Have you been using GKMS when testing? 
I checked out this branch again, ran it in our repo and it worked.
Then I deleted one of the refs for a target, which should re-generate the ref when compiling.
I ran `~/my_kapitan_dir/kapitan/bin/kapitan compile -t <my_target>` and got `Could not create GoogleKMSSecret: target_name missing` 
I also tested with two different targets and fails the same way.

bin/kapitan
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env bash
```
uberspot: We use this file for local testing sometimes. Any reason to remove it?

taharah: Actually not sure what happened here. I must of messed up the most recent rebase I did. I'll update the PR to not remove this file.

PR #371, opened 2019-09-14
Improve init templates

ademariag:
Fixes issue #

## Proposed Changes

`kapitan init` is a nice way to get users up and running, but the current set of templates are far from being ideal.

My suggestion is to have a much more minimal setup.

Included:
- wrapper `kapitan` script to use kapitan without installing it locally (using docker)

TODO:
- add apply.sh and setup_scripts


kapitan/inputs/templates/kapitan
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
uberspot: Change to `#!/usr/bin/env bash`

adrianchifor:
@ademariag are you still working on this or can we close?

PR #377, opened 2019-09-19
added fire for automatic cli generation

pranavgupta1234:
Fixes issue https://github.com/deepmind/kapitan/issues/240 

## Proposed Changes

  - Use fire for automatic CLI generation
  - paths will have to be passed comma separated now
  - kapitan --help will work now 
  - shorthand notation supported, eg. for an argument named recursive, --recursive, -recursive and -r works. Note that -r works smooth if there is no other parameter starting with r else fire will show both options to user. Capital (-R) is not supported.

Note that only those flags are updated which have same first letter in their argument name, like prune and parallelism in compile. Another change is passing comma separated paths. 

Although updating tests in never a good idea but still some changes had to be made due to change in way of using CLI after this update.

### DIFF
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -24,8 +24,11 @@ import json
 import logging
 import os
 import sys
+import fire
 
 import yaml
+
+from collections import namedtuple
 from kapitan import cached
 from kapitan.errors import KapitanError, RefHashMismatchError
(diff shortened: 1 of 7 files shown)


pranavgupta1234:
@adrianchifor @uberspot @ramaro suggestions ?

PR #379, opened 2019-09-28
support to create secrets on vault if file doesn't exist

vaibhavk:
Fixes issue #105 
Current vaultkv secret backend works for read-only vault secrets, where creating a secret through kapitan isn't possible.
## Proposed Changes
  - Create secret on vault if the current user is authorised to do so
  - Automatic vaultkv secret creation using the similar syntax of subvar i.e. `?{vaultkv:path/on/vault@key|randstr}`
  - if the file exists, using the content of the file instead of creating new secrets
cc @ramaro

### DIFF
--- a/kapitan/refs/secrets/vaultkv.py
+++ b/kapitan/refs/secrets/vaultkv.py
@@ -201,6 +201,12 @@ class VaultSecret(Base64Ref):
 
         return self._decrypt()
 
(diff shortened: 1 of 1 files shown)


uberspot:
Closing as stale, if anyone wants to have a go at this in the future, feel free to re-open.

Issue #346, opened 2019-08-23
Add jsonschema function for jsonnet, jinja2 and kadet 


ramaro:
Allow jsonschema validation via a jsonnet/kadet function or jinja2 filter.

This would allow a developer to define and validate a schema in the component's code. One example is validating the inventory parameters for a component.

uberspot:
Closing this as fixed by #380

PR #380, opened 2019-10-01
Jsonschema function for jsonnet

ramaro:
Introduces new `jsonschema` function in jsonnet.

It can be used to validate the inventory with jsonschema.

Given this inventory:

```yaml
mysql:                   
  storage: 10G           
  storage_class: standard
  image: mysql:latest    
```
The yaml inventory structure can be validated with the new `jsonschema()` function:

```jsonnet
local schema = { type: "object",
                 properties: {
                   storage: { type: "string", pattern: "^[0-9]+[MGT]{1}$"},
                   image: { type: "string" },
                 }};
// run jsonschema validation
local validation = kap.jsonschema(inv.parameters.mysql, schema);
// assert valid, otherwise error with validation.reason
assert validation.valid: validation.reason;
```

If `validation.valid` is not true, it will then fail compilation and display `validation.reason`.

For example, if defining the `storage` value with an invalid pattern (`10Z`), compile fails:

```
Jsonnet error: failed to compile /code/components/mysql/main.jsonnet:
 RUNTIME ERROR: '10Z' does not match '^[0-9]+[MGT]{1}$'

Failed validating 'pattern' in schema['properties']['storage']:
    {'pattern': '^[0-9]+[MGT]{1}$', 'type': 'string'}

On instance['storage']:
    '10Z'

/code/mysql/main.jsonnet:(19:1)-(43:2)

Compile error: failed to compile target: minikube-mysql
```

Partially fixes issue #346

### DIFF
--- a/docs/compile.md
+++ b/docs/compile.md
@@ -118,6 +118,51 @@ gzip_b64 - returns base64 encoded gzip of obj
 inventory - returns a dictionary with the inventory for target
 jsonschema - validates obj with schema, returns object with 'valid' and 'reason' keys
 ```
+##### Jsonschema validation from jsonnet
+
+Given the follow example inventory:
+
+```
+mysql:
+  storage: 10G
+  storage_class: standard
+  image: mysql:latest
+```
+
+The yaml inventory structure can be validated with the new `jsonschema()` function:
+
+```
+local schema = {
+    type: "object",
+    properties: {
+        storage: { type: "string", pattern: "^[0-9]+[MGT]{1}$"},
+        image: { type: "string" },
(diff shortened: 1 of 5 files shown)


uberspot:
Note: Add the example schema validation code in this PR to `docs/compile.md` for reference. :) 
Other than that, lgtm!

prakashrj:
Will it support reuse of object definitions like explained here? 
https://json-schema.org/understanding-json-schema/structuring.html

ramaro:
@prakashrj not that the moment, but we can look into extending support.

Issue #373, opened 2019-09-16
Allow `kapitan refs --reveal` to descend subdirectories


ademariag:
**Describe the bug/feature**
Especially with the new helm integration that produces manifests of subcharts in subdirectories, it makes sense to allow `kapitan refs --reveal` to descend subdirectories

```compiled/poc01-istio/chart/
└── istio                           
    ├── charts                                                           
    │   ├── galley                
    │   │   └── templates          
    │   │       ├── clusterrole.yaml
    │   │       ├── clusterrolebinding.yaml
    │   │       ├── configmap.yaml
    │   │       ├── deployment.yaml
    │   │       ├── poddisruptionbudget.yaml
    │   │       ├── service.yaml    
    │   │       └── serviceaccount.yaml    
    │   ├── gateways              
    │   │   └── templates                                                
    │   │       ├── autoscale.yaml 
    │   │       ├── deployment.yaml
    │   │       ├── poddisruptionbudget.yaml
    │   │       ├── role.yaml     
    │   │       ├── rolebindings.yaml
    │   │       ├── service.yaml    
    │   │       └── serviceaccount.yaml    
....
    └── templates
        ├── clusterrole.yaml
        ├── clusterrolebinding.yaml
        ├── configmap.yaml
        ├── serviceaccount.yaml
        └── sidecar-injector-configmap.yaml
```

**Additional context**

ademariag:
Also related: `reveal` complains if if finds subdirectories. This IMHO makes deploying charts with subcharts not possible or very awkward.
 
```
kapitan.sh refs --reveal -f ./compiled/poc01-istio-system/scripts/../manifests/istio
Traceback (most recent call last):
  File "/opt/venv/bin/kapitan", line 11, in <module>
    load_entry_point('kapitan==0.24.1rc3', 'console_scripts', 'kapitan')()
  File "/opt/venv/lib/python3.7/site-packages/kapitan/cli.py", line 364, in main
    ref_reveal(args, ref_controller)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/cli.py", line 540, in ref_reveal
    for rev_obj in revealer.reveal_path(file_name):
  File "/opt/venv/lib/python3.7/site-packages/kapitan/refs/base.py", line 171, in reveal_path
    content_yaml, content_json, content_raw = self._reveal_dir(path)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/refs/base.py", line 225, in _reveal_dir
[...]

PR #386, opened 2019-10-04
Add support for revealing files in subdirectories

ramaro:
Fixes issue #373


adrianchifor:
Should we add a test specific to subdirs?

ramaro:
@adrianchifor yep, will add that before merging

PR #388, opened 2019-10-04
Label selectors for compilation

### LINKED ISSUE #387: Label selectors for compilation
## Problem 
Currently we can only selectively compile targets using `-t` flag. In cases where we can't remember target names or there are just too many targets to input, we need an alternative.

## Proposal
Similar to how you can filter Kubernetes objects by label on kubectl with `-l` flag, I'm proposing an additional kapitan compile flag `--labels, -l`.

### Spec and usage
class-inherited-by-dev-targets.yml:
```
kapitan:
  labels:
    env: dev
```

class-inherited-by-prod-targets.yml:
```
kapitan:
  labels:
    env: prod
```

Only compile dev targets:
```
kapitan compile -l env=dev
```

adrianchifor:
Fixes #387

### DIFF
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -45,7 +45,7 @@ from reclass.errors import NotFoundError, ReclassException
 logger = logging.getLogger(__name__)
 
 
-def compile_targets(inventory_path, search_paths, output_path, parallel, targets, ref_controller, **kwargs):
+def compile_targets(inventory_path, search_paths, output_path, parallel, targets, labels, ref_controller, **kwargs):
     """
     Searches and loads target files, and runs compile_target() on a
     multiprocessing pool with parallel number of processes.
@@ -53,7 +53,7 @@ def compile_targets(inventory_path, search_paths, output_path, parallel, targets
     """
     # temp_path will hold compiled items
     temp_path = tempfile.mkdtemp(suffix='.kapitan')
-    updated_targets = targets
+    updated_targets = search_targets(inventory_path, targets, labels)
     # If --cache is set
     if kwargs.get('cache'):
         additional_cache_paths = kwargs.get('cache_paths')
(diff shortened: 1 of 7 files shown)


kapitan/targets.py
```diff
@@ -306,6 +306,43 @@ def load_target_inventory(inventory_path, targets):
...
 
+def search_targets(inventory_path, targets, labels):
+    """returns a list of targets where the labels match, otherwise just return the original targets"""
+    if not labels:
+        return targets
+
+    try:
+        labels_dict = dict(label.split('=') for label in labels)
+    except ValueError:
+        logger.error("Compile error: Failed to parse labels, should be formatted like: kapitan compile -l env=prod app=example")
```
ramaro: Raise CompileError instead (example https://github.com/deepmind/kapitan/blob/master/kapitan/targets.py#L80)

kapitan/targets.py
```diff
@@ -306,6 +306,43 @@ def load_target_inventory(inventory_path, targets):
...
+                logger.debug("search_targets: label %s=%s didn't match target %s", label, value, target_name)
+
+            matched_all_labels = False
+            break
+
+        if matched_all_labels:
+            targets_found.append(target_name)
+
+    if len(targets_found) == 0:
+        logger.error("No targets found with labels: {}".format(labels))
```
ramaro: raise CompileError

kapitan/targets.py
adrianchifor: Not sure if we should raise here as it would print a stacktrace, we only want to print a simple message with a hint.

kapitan/targets.py
adrianchifor: Same here, no stacktrace required as it's just a message, similar to what kubectl returns if no objects are found with labels.

kapitan/targets.py
ramaro: CompileError is clean in this case since it is triggered when compilation fails and no stack trace should appear

adrianchifor: Gotcha, making another PR now.

PR #392, opened 2019-10-10
Start release-v0.25.0-rc.2

uberspot:


Makefile
```diff
@@ -19,7 +19,7 @@ test:
 test_coverage:
 	@echo ----- Testing code coverage -----
 	coverage run --source=kapitan --omit="*reclass*" -m unittest discover
-	coverage report --fail-under=60 -m
+	coverage report --fail-under=64 -m
```
uberspot: Increasing this as the new coverage we have is 68% so there's still a safe limit there.

Issue #402, opened 2019-10-21
`kapitan init` generates a project which does not compile


jperville:
**Describe the bug/feature**

`kapitan init` generates a project which does not compile.

**To Reproduce**

```bash
$ mkdir xxx

$ cd xxx

$ docker run --rm -ti -u `id -u`:`id -g` -v $PWD:/src:delegated deepmind/kapitan:0.25 init   
Populated . with:
.
./inventory
./inventory/targets
	 my_target.yml
./templates
./templates/scripts
	 my_script.sh

$ docker run --rm -ti -u `id -u`:`id -g` -v $PWD:/src:delegated deepmind/kapitan:0.25 compile
Unknown (Non-Kapitan) Error occured
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/opt/venv/lib/python3.7/site-packages/kapitan/targets.py", line 383, in compile_target
    input_compiler.compile_obj(comp_obj, ext_vars, **kwargs)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/inputs/base.py", line 47, in compile_obj
    self.compile_input_path(input_path, comp_obj, ext_vars, **kwargs)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/inputs/base.py", line 67, in compile_input_path
    target_name=target_name, **kwargs)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/inputs/jsonnet.py", line 72, in compile_file
    fp.write_yaml(item_value)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/inputs/base.py", line 119, in write_yaml
    yaml.dump_all(obj, stream=self.fp, indent=indent, Dumper=PrettyDumper, default_flow_style=False)
  File "/opt/venv/lib/python3.7/site-packages/yaml/__init__.py", line 277, in dump_all
    for data in documents:
TypeError: 'bool' object is not iterable
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/venv/lib/python3.7/site-packages/kapitan/targets.py", line 94, in compile_targets
    [p.get() for p in pool.imap_unordered(worker, target_objs) if p]
  File "/opt/venv/lib/python3.7/site-packages/kapitan/targets.py", line 94, in <listcomp>
[...]

PR #403, opened 2019-10-21
Fix template for kapitan init so that the results compiles out of the box

jperville:
Fixes issue #402


jperville:
@googlebot I signed it!

ramaro:
Thanks for the fix @jperville !

PR #411, opened 2019-10-25
Fix Helm input type refs reveal or compile

taharah:
`yaml.safe_load_all(...)` returns a generator that needs to be converted to a
list. Otherwise, any refs will not be revealed or compiled when used with the
Helm input type.

Signed-off-by: Trevor Wood <Trevor.G.Wood@gmail.com>

## Proposed Changes

* Casts the generator returned by the function `yaml.safe_load_all(...)` to a list, so that any ref(s) will be revealed or compiled when writing the file to disk.

Reference: [https://stackoverflow.com/a/42805032](https://stackoverflow.com/a/42805032)


ramaro:
@taharah thanks for reporting. Can you provide an example how to reproduce this?

taharah:
Just use any ref(s) in a Helm chart and they will not be revealed or compiled in the final output. I can look to add testing if you'd like this to be a more comprehensive change.

adrianchifor:
This look good @taharah, thanks! For validation and future reference, do you mind adding a test with what the output should look like?

taharah:
@uberspot I've added tests as requested. Let me know if you would like me to add and/or change anything else.

Issue #368, opened 2019-09-12
Add support for helm bindings in OSX


uberspot:
As a followup from https://github.com/deepmind/kapitan/pull/359
This ticket is to track the work for OSX helm bindings + testing them on Travis.

cc @yoshi-1224

ademariag:
This would be very cool to have.... Any updates?

yoshi-1224:
@uberspot Happy to help

uberspot:
Go for it :) happy to review any PRs when you have em.

yoshi-1224:
@uberspot
So any suggestions as to what we should go with?

uberspot:
Yes, sorry I had to think a bit on the possible solutions. 
Dynamically fetching the .so file when kapitan runs might be a bit "dodgy". It's very rare that templating tools download shared libraries dynamically from the internet. And that might be also hard to do in environments that have restricted access to the internet.

I'd like to try and keep the steps of building the .so library in travis (osx) independent from the packaging steps to simplify maintenance.

So I'd suggest as a first step on this, let's focus on creating a travis job/pipeline that would build the osx bindings on travis (osx) and just release them to Github Releases. This should only happen on the master branch, same way we currently release to pypi and GH. 

A second step after that would be actually packaging the bindings to our pypi release. 
That could either be a manual step a maintainer does (less preferred) or if possible we could modify the existing travis job to wait (on the master branch) until the bindings are available (https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out  `travis_wait` referenced here).

That is a less elegant solution so I'd leave that as a further step and focus on first building and releasing the bindings on travis. :)

PR #414, opened 2019-11-03
Support helm template on osx

yoshi-1224:
Fixes issue #368 

## Proposed Changes
  - change "libtemplate.so" to "libtemplate_Linux.so" or "libtemplate_Darwin.so" depending on the platform
  - add osx jobs to travis for deploy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,22 @@
 language: python
 cache: pip
-matrix:
+jobs:
   include:
-  - python: 3.7
+  - stage: ubuntu-test-deploy
+    python: 3.7
     dist: xenial
     sudo: true
+  - stage: osx-deploy
+    if: tag IS present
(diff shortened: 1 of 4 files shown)


yoshi-1224:
@uberspot I set up Travis on my fork to test the osx environment. Everything seems fine except that when I run the tests on helm at once, it throws `fatal: morestack on g0` error. When I run the tests individually using `python -m unittest tests.test_helm_input.HelmInputTest.<test_method>` then they seem to run fine. Link to the build is [here](https://travis-ci.org/yoshi-1224/kapitan/builds/606771122?utm_source=github_status&utm_medium=notification). This is likely due to stack size difference in Linux and Mac from a quick research, as we know that the tests run without error on Ubuntu.
How should we deal with this?

Issue #235, opened 2019-03-08
Allow wildcards or directory for compile_input path.


rbounds:
**Describe the bug/feature**
Allow wildcards or directory for compile_input path.

It would be nice to define one class that handles inputs and outputs for everything in a directory. 
``` 
kapitan:
    compile:
      - output_path: manifests/$SUBDIRNAME/
        output_type: yaml
        input_paths:
          - components/*/*
        input_type: jsonnet
```

harsh-98:
We can use glob library it supports wildcard when searching for files and directory. Using glob, support for expressions like `directory/**/*`, `directory/*` , `direct*/*` can be implemented but then we need to check whether the path is a file and its extension. @ubespot what do you suggest?

uberspot:
Sounds reasonable. :+1:

adrianchifor:
You can't compile a whole folder as jsonnet won't know what to call the outputs and in what order to process them. This is why we use `main.jsonnet` in most components which then reference `deployment.jsonnet`, `service.jsonnet`, `secret.jsonnet` etc. 

Kadet will however default to `__init__.py` if you point it to a folder. Not sure about jinja.

harsh-98:
So, only for jinja2, there should be support for wildcard and directories.

harsh-98:
This functionality is already there in kapitan. For jinja2, we can pass directory and it will render for the files in that directory. Look at render_jinja2 in https://github.com/deepmind/kapitan/blob/master/kapitan/utils.py#L128. And this is directory tranversal cannot be used for jsonnet as mentioned by @adrianchifor

rbounds:
The core of the issue is more about the repetitive nature of `kapitan.compile` values in each component class. `input_paths` and `output_paths` follow a common pattern so I was thinking it would be nice for the user to define the pattern and use a derived value but now knowing a bit more about kapitan I don't see how it would cleanly work.

ramaro:
We could still support wildcards and leave it up to the user to define whatever is valid. For example, `directory/*/main.jsonnet` could still be possible and valid.

PR #419, opened 2019-11-12
add input_paths globbing support

ramaro:
Fixes issue #235


kapitan/cli.py
```diff
@@ -656,14 +656,7 @@ def secret_update_validate(args, ref_controller):
                         secret_obj.update_key(awskey)
                         ref_controller[token_path] = secret_obj
 
-            elif token_path.startswith("?{vaultkv:"):
```
uberspot: Is this intended?

ramaro: Yes, this is a cleanup and this code doesn't do anything!

PR #423, opened 2019-11-25
Copy - input that allows you to get your components plain unaltered to your compile targets

Cyanopus:
# Proposed Changes

  - The code provides an input feature of copy
  - This feature is to give an opportunity for a plain copy of a component without any alterations. Those can happen by other compiled steps
  - The use case come from the fact that the ArgoCD helm plugin requires a helm chart as it is and value files. By default, the directory plugin of Argo do not understand helm hooks. With this feature you can copy the helm chart and render additional value files at convenience.

### DIFF
--- a/kapitan/inputs/copy.py
+++ b/kapitan/inputs/copy.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+#
+# Copyright 2019 The Kapitan Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+import shutil
+
+from kapitan.inputs.base import InputType
+
+logger = logging.getLogger(__name__)
+
+
+class Copy(InputType):
+    def __init__(self, compile_path, search_paths, ref_controller):
+        super().__init__("copy", compile_path, search_paths, ref_controller)
+
+    def compile_file(self, file_path, compile_path, ext_vars, **kwargs):
+        """
+        Write items in path as plain rendered files to compile_path.
+        path can be either a file or directory.
+        """
+
+        try:
+            logger.debug("Copying {} to {}.".format(file_path, compile_path))
+            if os.path.isfile(file_path):
+                if os.path.isfile(compile_path):
+                    shutil.copy2(file_path, compile_path)
+                else:
+                    os.makedirs(compile_path, exist_ok=True)
+                    shutil.copy2(file_path, os.path.join(
+                        compile_path, os.path.basename(file_path)))
+            else:
+                if os.path.exists(compile_path):
+                    shutil.rmtree(compile_path)
+                shutil.copytree(file_path, compile_path)
+        except OSError as e:
+            logger.exception("Input dir not copied. Error: {}".format(e))
+
+    def default_output_type(self):
+        # no output_type options for copy
+        return None
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+#
+# Copyright 2019 The Kapitan Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"copy tests"
+import os
+import sys
+import shutil
+import hashlib
+import unittest
+
+from kapitan.cli import main
+from kapitan.utils import directory_hash
+from kapitan.inputs.copy import Copy
+
+
+search_path = ""
+ref_controller = ""
+test_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "test_copy")
+compile_path = os.path.join(test_path, "output")
+file_path = os.path.join(test_path, "input")
+test_file_path = os.path.join(file_path, "test_copy_input")
+test_file_compiled_path = os.path.join(compile_path, "test_copy_input")
+test_file_content = """
+apiVersion: v1
+kind: Pod
+metadata:
+  name: alpine
+  namespace: default
+spec:
+  containers:
+  - image: alpine:3.2
+    command:
+      - /bin/sh
+      - "-c"
+      - "sleep 60m"
+    imagePullPolicy: IfNotPresent
+    name: alpine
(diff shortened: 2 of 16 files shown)


kapitan/inputs/copy.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def compile_file(self, file_path, compile_path, ext_vars, **kwargs):
+        """
+        Write items in path as plain rendered files to compile_path.
+        path can be either a file or directory.
+        kwargs:
+            reveal: default False, set to reveal refs on compile
+            target_name: default None, set to current target being compiled
+        """
+        reveal = kwargs.get('reveal', False)
```
uberspot: This reveal and target_name aren't actually used anywhere in this function.
We can probably remove them from here (and the docstring above) to simplify the code?

kapitan/inputs/copy.py
```diff
@@ -0,0 +1,61 @@
...
+            logging.debug("Copying {} to {}.".format(file_path, compile_path))
+            if os.path.isfile(file_path):
+                if os.path.isfile(compile_path):
+                    shutil.copy2(file_path, compile_path)
+                else:
+                    os.makedirs(compile_path, exist_ok=True)
+                    shutil.copy2(file_path, compile_path + "/" +
+                                os.path.basename(file_path))
+            else:
+                #os.makedirs(compile_path, exist_ok=True)
```
uberspot: Remove commented out code. :)

kapitan/inputs/copy.py
```diff
@@ -0,0 +1,61 @@
...
+                if os.path.exists(compile_path):
+                    shutil.rmtree(compile_path)
+                shutil.copytree(file_path, compile_path)
+        except OSError as e:
+            logger.exception("Input dir not copied. Error: {}".format(e))
+
+    def default_output_type(self):
+        # no output_type options for jinja2
```
uberspot: `jinja2` in comment.

kapitan/inputs/copy.py
```diff
@@ -0,0 +1,61 @@
...
+        target_name = kwargs.get('target_name', None)
+
+        try:
+            logging.debug("Copying {} to {}.".format(file_path, compile_path))
+            if os.path.isfile(file_path):
+                if os.path.isfile(compile_path):
+                    shutil.copy2(file_path, compile_path)
+                else:
+                    os.makedirs(compile_path, exist_ok=True)
+                    shutil.copy2(file_path, compile_path + "/" +
```
uberspot: Avoid using harcoded path separators as they may differ between different operating systems.
Use `os.path.join` instead https://www.geeksforgeeks.org/python-os-path-join-method/

tests/test_copy.py
```diff
@@ -0,0 +1,97 @@
...
+
+from kapitan.cli import main
+from kapitan.utils import directory_hash
+from kapitan.inputs.copy import Copy
+
+
+search_path = ""
+ref_controller = ""
+test_path = os.path.dirname(os.path.abspath(__file__)) + "/test_copy"
+compile_path = test_path + "/output"
```
uberspot: Same for all these paths. Use `os.path.join`.

tests/test_copy.py
```diff
@@ -0,0 +1,97 @@
+import os
```
uberspot: This file is missing the Copyright header, as well as a docstring. 
You can run `make codestyle` in the root directory of kapitan and it will warn you about code style errors as well. :)

tests/test_copy.py
```diff
@@ -0,0 +1,97 @@
...
+class CopyTest(unittest.TestCase):
+
+    def setUp(self):
+
+        try:
+            shutil.rmtree(test_path)
+        except FileNotFoundError:
+            pass
+
+        for folder in [file_path, compile_path]:
```
uberspot: The code starting from here up to the end of this function should probably be in a separate helper function and called from `test_copy_file_folder` and `test_copy_folder_folder`? 
`setUp` is usually minimal and then in each test you run the compilation.

uberspot:
Hey @Cyanopus  :) Thanks for the PR. I added a few comments.

Could you also undo the "chmod -x" change to the shell scripts?

kapitan/inputs/copy.py
Cyanopus: Removed.

kapitan/inputs/copy.py
Cyanopus: Changed to copy.

kapitan/inputs/copy.py
Cyanopus: Changed it everywhere to be os.path.join.

tests/test_copy.py
Cyanopus: Changed them with os.path.join.

tests/test_copy.py
Cyanopus: Added the copyright, added docstring like the rest of the tests If I recall it correctly. :)
Ran `make codestyle` and fixed all style issues.

tests/test_copy.py
Cyanopus: Separated the folder structure to a helper function.

Cyanopus:
@uberspot Thanks for the feedback! I've addressed all of the notes, can you please review again? I will do my best to deliver better quality next time :)

kapitan/inputs/copy.py
Cyanopus: Removed them from the code and the docstring.

Cyanopus:
> Looks good to go. :) Thank you for the fixes.
> One last thing I'd ask for is to just update `docs/compile.md` with this new input type so the docs reflect the new functionality.
> Specifically add it to the `## Supported input types` and then maybe a short paragraph below that explains what this new input type does.
> Something like:
> 
> ```
> ### Copy
> 
> This input type simply copies the input templates to the output directory without any rendering/processing.
> For Copy, `input_paths` can be either a file or a directory: in case of a directory, all the templates in the directory will be copied and outputted to `output_path`.
> 
> *Supported output types*: N/A (no need to specify `output_type`)
> ```
> 
> 👍

Added the description to compile.md. Thanks for making it easier with writing it. Couldn't formulate it better :)

kapitan/inputs/copy.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__("copy", compile_path, search_paths, ref_controller)
+
+    def compile_file(self, file_path, compile_path, ext_vars, **kwargs):
+        """
+        Write items in path as plain rendered files to compile_path.
+        path can be either a file or directory.
+        """
+
+        try:
+            logging.debug("Copying {} to {}.".format(file_path, compile_path))
```
adrianchifor: `logger.debug`

Cyanopus: Fixed. @adrianchifor  thanks for the feedback!

adrianchifor:
Well done @Cyanopus and thanks for the contribution!

LGTM, let's wait for the tests to pass and we can merge. Will start a release candidate for 0.26 right after as we've added a few new things since the last major one.

paulbeattie:
FWIW this breaks the readme/getting started as `copy` is not a valid option in version `<0.26.0`. 

Error for those searching in the near future
```
Invalid inventory structure

Error: 'copy' is not one of ['jsonnet', 'kadet']
On instance:
"copy"
```

Cyanopus:
> FWIW this breaks the readme/getting started as `copy` is not a valid option in version `<0.26.0`.
> 
> Error for those searching in the near future
> 
> ```
> Invalid inventory structure
> 
> Error: 'copy' is not one of ['jsonnet', 'kadet']
> On instance:
> "copy"
> ```

It's probably a good idea to have each new feature specify explicitly since which version it is available in docs to avoid confusion.

adrianchifor:
That's what a changelog is for :) https://github.com/deepmind/kapitan/blob/master/CHANGELOG.md

PR #426, opened 2019-11-29
Fix input_paths globbing when search paths are defined

simu:
`glob.glob` returns an empty array if it cannot resolve the provided
path, regardless of whether the provided path contains a globbed element
or not.

This commit reorganized the code in `compile_obj` and `compile_input_path`
to take into account the provided search paths during globbing.  As a
side-effect we always call `compile_input_path` with a fully resolved
path to the input file, and do not need to search the file in all search
paths again.

The CompileError for inputs which are not found has been moved to the
globbing phase, and is raised when `glob.glob` returns an empty array
for a given `input_path` in all provided search paths.

Fixes issue: no issue filed

## Proposed Changes

  - Fix `compile_obj` to take search paths into consideration when resolving globbed paths

### DIFF
--- a/kapitan/inputs/base.py
+++ b/kapitan/inputs/base.py
@@ -38,44 +38,42 @@ class InputType(object):
 
     def compile_obj(self, comp_obj, ext_vars, **kwargs):
         """
-        run compile_input_path() for each input_path in comp_obj
-        kwargss are passed into compile_input_path()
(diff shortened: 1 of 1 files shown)


simu:
@googlebot I signed it!

ramaro:
@simu excellent, thank you!

Issue #434, opened 2019-12-14
Output as a string does not interpolate secrets


tprobinson:
**Describe the bug/feature**
Using 0.26.0, when a jsonnet component outputs a string and not an object, secret references remain uninterpolated unless `kapitan reveal` is run.

**To Reproduce**
1. Create files in a new Kapitan project as seen in https://github.com/tprobinson/kapitan-issue-demonstration

2. Run `kapitan compile`.

**Expected behavior**
Secrets should be interpolated in the final output.

**Additional context**
If secrets are interpolated by walking the object that's output by jsonnet, this may be difficult. On a related note, the `?{ ... }` secret references get mangled when attempting to operate on them with functions like url encoding. If the secrets' actual value were passed in to jsonnet instead of just strings containing the `?{ ... }` references, that may solve two issues at once.

ramaro:
@tprobinson thanks for reporting. Currently, refs/secrets are designed to be used _only_ from the inventory. Manipulation of these objects via an input will result in undefined behavior as you are describing.
I think the best way to improve this is to implement ref aware handling functions for input types.

ramaro:
@tprobinson and now that I had a better second look at this, you're right, this should not be happening. Thanks again, will review.

ramaro:
@tprobinson this is fixed, let me know if that works for you as well. Thank you

tprobinson:
I'm trying to test this, but I can't seem to compile the binary myself. I'm not much of a Python dev so I'm probably doing something wrong, but running via the Dockerfile build gets me this:
```
$ docker run --rm kapitan:test help
Traceback (most recent call last):
  File "/opt/venv/bin/kapitan", line 11, in <module>
    load_entry_point('kapitan==0.26.0', 'console_scripts', 'kapitan')()
  File "/opt/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/opt/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    return ep.load()
  File "/opt/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2443, in load
    return self.resolve()
  File "/opt/venv/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2449, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
[...]

PR #438, opened 2019-12-20
refs compile/reveal interpolation fix and tests

ramaro:
Fixes issue #434


kapitan/inputs/base.py
```diff
@@ -129,9 +129,9 @@ def write_json(self, obj):
         reveal = self.kwargs.get('reveal', False)
         target_name = self.kwargs.get('target_name', None)
         if reveal:
-            self.revealer.reveal_obj(obj)
+            obj = self.revealer.reveal_obj(obj)
```
uberspot: one extra space in indentation here

Issue #430, opened 2019-12-02
kapitan compile should produce error messages with locations to kapitan sources


NicolasRouquette:
**Describe the bug/feature**

Since `kapitan compile` is intended to behave like a programming language compiler, it makes sense to expect that kapitan errors should refer to specific locations in the kapitan sources.

Currently, (0.26.0-rc.1), it seems that kapitan errors provide a description of the problem but there is no information about the location of the error in the kapitan sources.

I stumbled on this by following the blog article.

**To Reproduce**
1. Use kapitan 0.26.0-rc.1

2. Clone https://github.com/ademariag/kapitan-examples/commit/1115f98f8ec5091e308876641f6c0c73235d858d
3. compile

```bash
cd kapitan-examples/example-3
kapitan compile
```
```
no backend for ref type: ref
Compile error: failed to compile target: dev-sea
```

**Expected behavior**

Something like this:
```bash
Compiled dev-sea (0.32s)
Compiled prod-sea (0.32s)
```

** If it's a bug (please complete the following information):**
 - `python --version`: N/A
 - `pip3 --version`: N/A
 - `Are you using pyenv or virtualenv?` N/A

Ran: `docker run -t --rm -v $(pwd):/src:delegated deepmind/kapitan`

**Additional context**

MatteoVoges:
I think the linked pr's should have solved this issue. Please let me know if some other error messages are still unclear.

PR #439, opened 2019-12-20
Add key info for RefBackendError for easier debugging

uberspot:
Related to issue #430 
Slightly hard to generalize and add more information (like current filename processed for example)
because references are set/read in multiple places in cli.py as well in various use cases.
It might be feasible to add nicer exception messages there too with the same approach. 

New message looks like this:
```
no backend for ref type: gpsafasg
  for key ?{gpsafasg:targets/minikube-mysql/mysql/password||randomstr|base64}
Compile error: failed to compile target: minikube-mysql
```
Which is slightly more helpful but not perfect.

### DIFF
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -29,17 +29,17 @@ import yaml
 from kapitan import cached
 from kapitan.errors import KapitanError, RefHashMismatchError
 from kapitan.initialiser import initialise_skeleton
+from kapitan.inputs.jinja2_filters import DEFAULT_JINJA2_FILTERS_PATH
 from kapitan.lint import start_lint
 from kapitan.refs.base import PlainRef, RefController, Revealer
 from kapitan.refs.base64 import Base64Ref
 from kapitan.refs.secrets.awskms import AWSKMSSecret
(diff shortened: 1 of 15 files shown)


kapitan/cli.py
```diff
@@ -405,11 +406,11 @@ def ref_write(args, ref_controller):
             inv = inventory_reclass(args.inventory_path)
             kap_inv_params = inv['nodes'][args.target_name]['parameters']['kapitan']
             if 'secrets' not in kap_inv_params:
-                raise KapitanError("parameters.kapitan.secrets not defined in {}".format(args.target_name))
+                raise KapitanError("parameters.kapitan.secrets not defined in inventory of target {}".format(args.target_name))
 
             recipients = kap_inv_params['secrets']['gpg']['recipients']
         if not recipients:
-            raise KapitanError("No GPG recipients specified. Use --recipients or specify them in " +
+            raise KapitanError("No GPG recipients specified. Use --recipients or specify them in" +
```
adrianchifor: space was needed here

uberspot: Nice catch, fixed.

PR #442, opened 2019-12-20
Python black

uberspot:
I just ran `black -l 110 -t py37 --exclude reclass .` to enforce the same formatting on all kapitan source code for easier reviewing/coding in the future. 

https://pypi.org/project/black/

### DIFF
--- a/kapitan/inputs/helm/cffi_build.py
+++ b/kapitan/inputs/helm/cffi_build.py
@@ -5,18 +5,17 @@ def main():
     ffi = FFI()
 
     # declare functions to export
-    ffi.cdef("""
+    ffi.cdef(
+        """
         char* renderChart(char* p0, char* p1, char* p2, char* p3, char* p4, char* p5);
         void free(void* ptr);
-    """)
-
-    ffi.set_source(
-        "helm_binding",  # specify name for importing this module
-        None
+    """
     )
+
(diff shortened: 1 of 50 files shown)


ramaro:
We should make this part of the pipeline

uberspot:
> We should make this part of the pipeline

Done :)

kapitan/inputs/helm/cffi_build.py
```diff
@@ -5,18 +5,17 @@ def main():
...
         void free(void* ptr);
-    """)
-
-    ffi.set_source(
-        "helm_binding",  # specify name for importing this module
-        None
+    """
     )
 
+    ffi.set_source("helm_binding", None)  # specify name for importing this module
```
gburiola: interesting to see how black reformatted this block. before, the comment was in the middle of the parameters.

kapitan/lint.py
```diff
@@ -56,7 +56,15 @@
 """
 
 
-def start_lint(fail_on_warning, skip_class_checks, skip_yamllint, inventory_path, search_secrets, secrets_path, compiled_path):
+def start_lint(
```
gburiola: sooo much easier to read

uberspot: I know right :D

Issue #445, opened 2020-01-02
Kadet inputs with plain output type and dict objects fail to compile


uberspot:
**Describe the bug/feature**
When Kadet is used in conjunction with `plain` output type it fails to write the output to a file.
The issue is in CompiledFile.write as it assumes that data passed to it is strings but that's not always the case (especially if in Kadet where you can define the exact outputted object as a python dictionary).

Solution: Convert data explicitly to str in write() (PR coming soon).

To replicate:

Kapitan config:
```
  kapitan:
    compile:
      - output_path: ""
        input_type: kadet
        output_type: plain
        input_paths:
          - components/bla
```

Kadet init.py:
```
.....
def main():
    obj = kadet.BaseObj()
    inv = kadet.inventory()
    kubelib = kadet.load_from_search_paths("kubelib")

    data = {}

    obj.root['my_dir/{}'.format(file_name)] = d['raw']

    obj.root['my_other_dir/my_config.yml'] = kubelib.ConfigMap(
        name='some_config_map',
        labels={},
        data=data
    )

    return obj
```

Stacktrace:
```
Traceback (most recent call last):
  File "~/.pyenv/versions/3.7.1/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "~/kapitan/kapitan/targets.py", line 387, in compile_target
    input_compiler.compile_obj(comp_obj, ext_vars, **kwargs)
  File "~/kapitan/kapitan/inputs/base.py", line 59, in compile_obj
    self.compile_input_path(input_path, comp_obj, ext_vars, **kwargs)
  File "~/kapitan/kapitan/inputs/base.py", line 74, in compile_input_path
    target_name=target_name, **kwargs)
  File "~/kapitan/kapitan/inputs/kadet.py", line 134, in compile_file
    fp.write(item_value)
  File "~/kapitan/kapitan/inputs/base.py", line 107, in write
    self.fp.write(self.revealer.compile_raw(data, target_name=target_name))
  File "~/kapitan/kapitan/refs/base.py", line 314, in compile_raw
    compiled = self.regex.sub(self._compile_replace_match_with_args(**kwargs), data)
TypeError: expected string or bytes-like object
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "~/kapitan/kapitan/targets.py", line 95, in compile_targets
    [p.get() for p in pool.imap_unordered(worker, target_objs) if p]
[...]

PR #446, opened 2020-01-02
Fix issue #445: Explicitly create directories for all outputted CompiledFile objects. 

uberspot:
Fixes issue #445


uberspot:
I modified this PR to just create the needed dirs in the CompiledFile class for all input types vs creating in jinja2 explicitly. 
After investigating more it makes sense that converting to str() can be done explicitly in Kadet anyway so enforcing the conversion isn't necessary.

Issue #444, opened 2019-12-24
SUGGESTION: Allow prune per compile output


ademariag:
**Describe the bug/feature**
Sometimes global prune on/off is not enough.

I propose to add the optional "prune" option to a compile configuration:

```
    kapitan:
      compile:
        - output_path: terraform
          output_type: json
          output_prune: false
          input_paths:
            - components/terraform/main.jsonnet
          input_type: jsonnet
```

**Additional context**
In particular, to give a use case, I compile terraform together with spinnaker and kubernetes, and as for this "bug" https://github.com/terraform-providers/terraform-provider-google/issues/4431 terraform sometimes needs to see the nulls

PR #451, opened 2020-01-15
444: Enables per compile output prune

ademariag:
Fixes issue #444 

## Proposed Changes

  - Enables per compile output prune

User can now explicitly setup a prune policy on a per compile output basis

### DIFF
--- a/kapitan/inputs/jsonnet.py
+++ b/kapitan/inputs/jsonnet.py
@@ -53,7 +53,7 @@ class Jsonnet(InputType):
         output_obj = json.loads(json_output)
 
         output = kwargs.get("output", "yaml")
-        prune = kwargs.get("prune", False)
+        prune = kwargs.get("prune_input", False)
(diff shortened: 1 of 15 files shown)


uberspot:
Could we add a test for this? :)

kapitan/inputs/jsonnet.py
```diff
@@ -53,7 +53,7 @@ def _search_imports(cwd, imp):
         output_obj = json.loads(json_output)
 
         output = kwargs.get("output", "yaml")
-        prune = kwargs.get("prune", False)
+        prune = kwargs.get("prune_input", False)
```
uberspot: Not sure if this will work. I have the impression `kwargs` here is referencing `--prune` from the command line.

ademariag:
@uberspot please take another look

ademariag:
@uberspot can you merge as I am not authorized...

Issue #275, opened 2019-04-19
Ability to specify custom target path


afoninsky:
I’ve found that I need to share inventory targets and secrets between my customers keeping the rest of configuration in common maintainable repository. I can specify `--secrets-path` but it’s not possible to specify `--target-path` (`--inventory-path` does not fit as it includes classes too)

**Feature request**
(minimum) Add flag `--target-path[s?]` which points to custom directory with targets
(maximum) Add ability to specify pathes using configuration file, so users will be able to maintain own repository structures

supra08:
@uberspot I would like to work on this issue.

MatteoVoges:
As a workaround you could change the `nodes_uri` in your `reclass-config.yml` file. This file has to be in the inventory_path which you specify with `--inventory-path` or default `inventory`. Please let me know, if there are any further questions.

PR #468, opened 2020-03-10
[WIP] Add CLI option to select custom targets path

supra08:
Fixes issue #275 

## Proposed Changes

  - Add CLI flag to add custom targets path
  - Expand the path and save it as a reclass object

There is only one problem that I am facing:
`kapitan/lib/kapitan.libjsonnet` has this
```
  inventory(target=std.extVar("target"), inv_path="inventory/"):: std.native("inventory")(target, inv_path),
```
hardcoded. So is there a way to override this when dealing with custom target paths?
@uberspot

### DIFF
--- a/kapitan/resources.py
+++ b/kapitan/resources.py
@@ -231,7 +231,7 @@ def inventory(search_paths, target, inventory_path="inventory/"):
     return inventory_reclass(full_inv_path)["nodes"][target]
 
 
-def inventory_reclass(inventory_path):
+def inventory_reclass(inventory_path, targets_path=None):
     """
     Runs a reclass inventory in inventory_path
     (same output as running ./reclass.py -b inv_base_uri/ --inventory)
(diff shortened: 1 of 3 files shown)


supra08:
@googlebot I signed it!

kapitan/resources.py
```diff
@@ -231,7 +231,7 @@ def inventory(search_paths, target, inventory_path="inventory/"):
     return inventory_reclass(full_inv_path)["nodes"][target]
 
 
-def inventory_reclass(inventory_path):
+def inventory_reclass(inventory_path, targets_path=None):
     """
     Runs a reclass inventory in inventory_path
```
uberspot: Can you update this docstring as well to reflect the new behavior?

kapitan/targets.py
```diff
@@ -563,8 +573,11 @@ def schema_validate_compiled(targets, compiled_path, inventory_path, schema_cach
     worker = partial(schema_validate_kubernetes_output, cache_dir=schema_cache_path)
     pool = multiprocessing.Pool(parallel)
 
+    if targets_path == None:
+        targets_path = inventory_path
```
uberspot: Shouldn't this be set to `os.path.join(inventory_path, "targets")` ? 
Or am I missing something?

supra08: Yeah I missed out here, will fix it.

kapitan/resources.py
supra08: alright!

uberspot:
Besides my comment, the only other thing missing is a small test to check that this flag is actually used. Could you add one under `kapitan/tests/test_compile.py` ? :)

supra08:
Yeah sure! I'll add it ASAP. Apologies for the delay.

Issue #255, opened 2019-04-01
Document the Kadet input


th31nitiate:
I apologize as this is just a request for an understanding of what Kadet is and how it functions?

I am trying to gain a better understanding as to when using the dev version I am able to init a directory with examples, but it is not very clear as to the function.

It looks like a method to perform inventory from python scripts but this is not clear.

ramaro:
Kadet is undocumented on purpose because it is experimental. But have a read of  https://github.com/deepmind/kapitan/pull/190 and you'll get a feel of what it currently does.

uberspot:
I changed the ticket to track documenting the feature in the future.

PR #482, opened 2020-04-19
Add kadet usage documentation

th31nitiate:
Fixes issue #255

I was not too sure about how detailed to be and if the location was appropriate. I made the best effort to ensure compatibility with other sections of that particular page. My main aim was to ensure people can build up on the examples.

## Proposed Changes

  - Add kadet documentation

### DIFF
--- a/docs/compile.md
+++ b/docs/compile.md
@@ -188,7 +188,69 @@ It's up to you to decide what the output is.
 
 ### kadet
 
-This input type is experimental. See <https://github.com/deepmind/kapitan/pull/190> for its usage.
+Kadet is an extensible input type that enables you to generate templates using python. The key benefit being the ability to utilize familiar programing principles while having access to kapitan's powerful inventory system. 
+
+A library that defines resources as classes using the Base Object class (BaseObj) is required. These can then be utilized within components to render output. 
+
+The following functions are provided by the class BaseObj.
+
+Method definitions:<br/>
+
+* new: Provides parameter checking capabilities<br/>
+* body: Enables in-depth parameter configuration<br/>
+
+Method functions:<br/>
+
+* root: Defines values that will be compiled into the output<br/>
+* need: Ability to check & define input parameters<br/>
+* update_root: Updates the template file associated with the class<br/>
+
+A class can be a resource such as a kubernetes Deployment as shown here:
+
+```
+class Deployment(BaseObj):
+    def new(self):
+        self.need("name", "name string needed")
+        self.need("labels", "labels dict needed")
+        self.need("containers", "containers dict needed")
+        self.update_root("lib/kubelib/deployment.yml")
+
+    def body(self):
+        self.root.metadata.name = self.kwargs.name
+        self.root.metadata.namespace = inv.parameters.target_name
+        self.root.spec.template.metadata.labels = self.kwargs.labels
+        self.root.spec.template.spec.containers = self.kwargs.containers
+```
+
+The deployment is an BaseObj which has two main funcations. New can be used to perform parameter validation & template compilation. Body is utilized to set those parameters to be rendered. `self.root.metadata.name` is a direct reference to a key in the corresponding yaml.
+
+We have established that you may define a library which holds information on classes that represent resource objects. The library is then utilized by defined components to generate the required output.
+
+Here we import `kubelib` using `load_from_search_paths`. We then use kubelib to access the defined object classes. In this instance the Deployment & Service resource class.
+
+```
+...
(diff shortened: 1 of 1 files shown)


docs/compile.md
```diff
@@ -188,7 +188,69 @@ It's up to you to decide what the output is.
...
 ### kadet
 
-This input type is experimental. See <https://github.com/deepmind/kapitan/pull/190> for its usage.
+Kadet is an extensible input type that enables you to generate templates using python. The key benefit being the ability to utilize familiar programing principles while having access to kapitan's powerful inventory system. 
+
+A library that defines resources as classes using the Base Object class (BaseObj) is required. These can then be utilized within components to render output. 
+
+The following functions are provided by the class BaseObj.
+
+Method definitions:<br/>
```
uberspot, line 197: You don't need these <br/> breakpoints  in markdown, just leave an empty newline instead and it will render it  correctly.
Same for all  these breaklines below.

docs/compile.md
```diff
@@ -188,7 +188,69 @@ It's up to you to decide what the output is.
...
+        self.update_root("lib/kubelib/deployment.yml")
+
+    def body(self):
+        self.root.metadata.name = self.kwargs.name
+        self.root.metadata.namespace = inv.parameters.target_name
+        self.root.spec.template.metadata.labels = self.kwargs.labels
+        self.root.spec.template.spec.containers = self.kwargs.containers
+```
+
+The deployment is an BaseObj which has two main funcations. New can be used to perform parameter validation & template compilation. Body is utilized to set those parameters to be rendered. `self.root.metadata.name` is a direct reference to a key in the corresponding yaml.
```
uberspot, line 225: Typo in `functions`.

docs/compile.md
```diff
@@ -188,7 +188,69 @@ It's up to you to decide what the output is.
...
+def main():
+    output = kadet.BaseObj()
+    output.root.nginx_deployment = kubelib.Deployment(name=name, labels=labels, containers=[nginx_container])
+    output.root.nginx_service = kubelib.Service(
+        name=name, labels=labels, ports=[svc_port], selector=svc_selector
+    )
+    return output
+```
+
+Kadet uses a library called addict to organise the parameters inline with the yaml templates.
```
uberspot, line 250: Would be nice to add a hyperlink in `addict` linking to the library repo.

docs/compile.md
```diff
@@ -188,7 +188,69 @@ It's up to you to decide what the output is.
...
+As shown above we create a BaseObject named output. We update the root of this output with the data structure returned from kubelib. This output is what is then returned to kapitan to be compiled into the desired output type.
+
+For a deeper understanding of this input type please review the proposal document at [kadet](/kap_proposals/kap_0_kadet/) & the examples located at `examples/kubernetes/components/nginx`.
```
uberspot, line 253: Are you sure the link for `/kap_proposals/kap_0_kadet/` will be rendered and work? It might have to start with `./` ?

docs/compile.md
th31nitiate: Sorry for making you review these simple mistakes. Initially, it did not render as intended so I looked for a workaround. I added the newline's as recommended and it seems fine.

docs/compile.md
th31nitiate: school boy error :(, should be paying more attention

docs/compile.md
th31nitiate: I believe it will. When I prepend `./` it attempts to access `/compile/kap_proposals/kap_0_kadet/`

docs/compile.md
```diff
@@ -188,7 +188,69 @@ It's up to you to decide what the output is.
...
-This input type is experimental. See <https://github.com/deepmind/kapitan/pull/190> for its usage.
+Kadet is an extensible input type that enables you to generate templates using python. The key benefit being the ability to utilize familiar programing principles while having access to kapitan's powerful inventory system. 
+
+A library that defines resources as classes using the Base Object class (BaseObj) is required. These can then be utilized within components to render output. 
+
+The following functions are provided by the class BaseObj.
+
+Method definitions:
+
+* new: Provides parameter checking capabilities
```
ramaro, line 199: Maybe update method names to `function()`? So we'd have `new()`, `body()`, etc...

th31nitiate: True, that does make it more clear. I change references to BaseObj to reflect this also. This just helps readers better understand the meaning. 

I removed the following reference on line 193 since it generally did not feel right.

- Base Object class (`BaseObj()`) is required

Issue #483, opened 2020-04-19
Add terraform example documentation 


th31nitiate:
**Describe the bug/feature**
Update the Terraform example in the kapitan documentation

**To Reproduce**
There is a kubernete's example on the documentation but there is no terraform example.

**Expected behavior**

**Screenshots**
![docs-view](https://user-images.githubusercontent.com/14749635/79696213-3a594400-8273-11ea-831a-fcc61c50e8cf.png)

**Additional context**
It will enable adaptability & better understanding of the tools capabilities

PR #488, opened 2020-04-21
Terraform example docs

th31nitiate:
Fixes issue #483 

This add the Terraform example documentation as described in the issue. I attempted to explained things the way I understood. 

I also made sure to perform spell/grammer checking to avoid wasting your time while performing the review.

## Proposed Changes

  - Add Terraform usage example documentation

### DIFF
--- a/docs/example-terraform.md
+++ b/docs/example-terraform.md
@@ -1,19 +1,19 @@
 # Terraform example
 
-We will be looking at how to use  kapitan to compile terraform files with Jsonnet as the input type. It's possible to use other input types, however Jsonnet is recommended. For example we could use the kadet input to generate terraform files but this would require 
-templates to be written in yaml then to rendered into json. Jsonnet is the most suitable input type as you will see due to its functional nature. The only appropriate output type is `json` since this is the format that terraform consumes.
+We will be looking at how to use Kapitan to compile terraform files with Jsonnet as the input type. It's possible to use other input types, however, Jsonnet is recommended. 
+For example, we could use the Kadet input to generate terraform files but this would require templates to be written in YAML then to rendered into JSON.
+Jsonnet is the most suitable input type as you will see due to its functional nature. The only appropriate output type is JSON since this is the format that Terraform consumes.
 
 ## Directory structure
 
-There is a number of examples available in `examples/terraform`. This will be our working directory for this documentation. The directory structure is as follows:
-
+There are several examples available in `examples/terraform`. This will be our working directory for this documentation. The directory structure is as follows:
 ```
 ├── inventory
 └── templates
 ```
 
-It is possible to further extend this locally to include a `lib` directory where a `terraform.libjsonnet` file can be stored for use. This is generally dependent on the project scope & organisational patterns. 
-We will describe in more detail the role of each these folders in the following sections.
+It is possible to further extend this locally to include a `lib` directory where a `terraform.libjsonnet` file can be stored for use. This is generally dependent on the project scope & organizational patterns. 
+We will describe in more detail the role of each of these folders in the following sections.
 
 ### inventory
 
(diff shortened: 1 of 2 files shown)


docs/example-terraform.md
```diff
@@ -0,0 +1,187 @@
+# Terraform example
+
+We will be looking at how to use Kapitan to compile terraform files with Jsonnet as the input type. It's possible to use other input types, however, Jsonnet is recommended. 
+For example, we could use the Kadet input to generate terraform files but this would require templates to be written in YAML then to rendered into JSON.
```
uberspot, line 4: "then rendered into JSON" *

docs/example-terraform.md
```diff
@@ -0,0 +1,187 @@
...
+Jsonnet is the most suitable input type as you will see due to its functional nature. The only appropriate output type is JSON since this is the format that Terraform consumes.
```
uberspot, line 5: Maybe i'd rephrase "Jsonnet is the most suitable input type as you will see due to its functional nature." to "Jsonnet is the most straightforward input type as you will see due to its functional nature.". Suitable sounds a bit as if Kapitan is biased between input types. :)

docs/example-terraform.md
```diff
@@ -0,0 +1,187 @@
...
+export TF_DATA_DIR=$(realpath -m ${DIR}/../../../.TF_DATA_DIR/{{inventory.parameters.name}}) # Folder for TF initialization (preferable outside of compiled)
+export OUTPUT_DIR=$(realpath -m ${DIR}/../../../output/{{inventory.parameters.name}}) # Folder for storing output files (preferable outside of compiled)
+```
+
+It is good practice to utilize this method to improve integration with various CLI based tools. Scripts help to ensure terraform &
+kapitan can function with your CI/CD systems. It generally depends on your organizational workflows. 
+
+### secrets
+
+Although there are no particular secrets in this instance. It is possible to utilize Kapitan secrets as defined in [secrets management](secrets.md). There is truly no limit to the extensibility of this tool.
```
uberspot, line 187: "There is truly no limit to the extensibility of this tool." - Definitely appreciate the comment :) but I think it might be misinterpreted if we add it in the terraform docs? 
Might be better to keep this doc purely about terraform?

docs/example-terraform.md
```diff
@@ -0,0 +1,187 @@
...
+When terraform runs it will use the output module to generate your desired `README.md` using information from terraform's state.
+
+Scripts are located in the `scripts` directory. They are compiled using jinja2 as the templating language. An example is as follows:
+
+```shell
+export TF_DATA_DIR=$(realpath -m ${DIR}/../../../.TF_DATA_DIR/{{inventory.parameters.name}}) # Folder for TF initialization (preferable outside of compiled)
+export OUTPUT_DIR=$(realpath -m ${DIR}/../../../output/{{inventory.parameters.name}}) # Folder for storing output files (preferable outside of compiled)
+```
+
+It is good practice to utilize this method to improve integration with various CLI based tools. Scripts help to ensure terraform &
```
uberspot, line 182: All this looks great. :) Appreciate the effort!
One last comment is if you could replace `&` with `and` as it makes searching codebases/documentation more intuitive?

docs/example-terraform.md
ramaro: This is actually not true. Kadet will work with both JSON or YAML.

th31nitiate: Thanks for spotting that. 

@ramaro Thanks that sounds important. It means integration would be much simpler since we can convert HCL to JSON. Kadet can then use the pre-existing JSON as  input templates.

docs/example-terraform.md
th31nitiate: That is true. It is important to highlight that Kapitan is not biased on input types. It is dynamic which enables creativity.

docs/example-terraform.md
th31nitiate: I felt like I was not doing the secrets section much justice.

I do agree with you on that one though.

docs/example-terraform.md
th31nitiate: I never considered that before. Good point :D

PR #492, opened 2020-04-23
Fix #491

### LINKED ISSUE #491: Prune empty file behaviour
**Describe the bug/feature**

NOTE: Offline chat with team decided that the above behaviour is acceptable. Reporting for posterity.

With the `jsonnet` output, so we have 2 ways to generate files: let’s call them ***single file*** and ***multi file*** mode.

[relevant code](https://github.com/deepmind/kapitan/blob/master/kapitan/inputs/jsonnet.py#L56) 

### single file

a jsonnet file called `main.jsonnet` that produces a content that "is not a json dict" will add that content in a file `main.yml`

***Note:*** this condition (not a dict) can never normally be invoked, and it only happens in the rare situation where prune is enabled ***AND*** the produced json file is 
```
{}
```
resulting into a Null object.
In this case, we take the filename of the jsonnet itself: in my case, `main.yml`, and generate it as an empty file.

### multi file modea

a jsonnet file that produces a dict of values, for instance 
```
{ 
  "manifest": { 
    foo: true
  }
}
```

And we write out the “manifest.yml” file with content:
```
{
  "foo": true
}
```

If the content of the "manifest" key is empty, ***AND*** prune is enabled, kapitan will ***NOT*** generate the file 

My points:
* Given that the switch between single and multi file mode is based on a “is this a dict” [check](https://github.com/deepmind/kapitan/blob/master/kapitan/inputs/jsonnet.py#L56), which can never be triggered under normal use (no actual use case), why do we have it at all? It creates confusion, and as seen above, it leads to inconsistencies.

For a real life example where this behaviour is a problem, see: 

https://github.com/kapicorp/kapitan-reference/blob/bc92712591c23ba46f342ef11564586fa8c58e85/components/generators/ingresses/main.jsonnet#L17

and in general with every generator

**To Reproduce**
1. Create files as described in the above examples
2. run kapitan compile
3. observe behaviour as described

**Expected behavior**

Single mode file should EITHER not exist, or be consistent with multi file and not create an empty file when prune is on.

**Additional context**
Offline chat with team decided that the above behaviour is acceptable.

ademariag:
Fixes issue #491 

## Proposed Changes

* check that the generated content is a dictionary

for JSON to be valid, the content of `output_obj` needs to be:

```
{ "key": "value"}
```

or even just:
```
{}
```

which in both cases would be a "dict", which would make it ignore the
`if not isinstance(output_obj, dict)` check.

This means that the only moment that check is evaluated is when, with `prune: true`, `{}` becomes NoneType.

## Test
adding the following code
```
 56         print("output_obj is of type: ", type(output_obj))
 57         return
```

### Without `prune` enabled:
```
output_obj is of type:  <class 'dict'> <- because it is {}
```

### With `prune` enabled:
```
output_obj is of type:  <class 'NoneType'> <- prune makes it NoneType, and this triggers the "single file" mode
```

### DIFF
--- a/kapitan/inputs/jsonnet.py
+++ b/kapitan/inputs/jsonnet.py
@@ -53,53 +53,46 @@ class Jsonnet(InputType):
             output_obj = prune_empty(output_obj)
             logger.debug("Pruned output for: %s", file_path)
 
-        if not isinstance(output_obj, dict):
-            tmp_output_obj = output_obj
-            # assume that the output filename is the
-            # same as the input jsonnet filename
-            filename = os.path.splitext(os.path.basename(file_path))[0]
-            output_obj = {}
-            output_obj[filename] = tmp_output_obj
-
-        for item_key, item_value in output_obj.items():
-            # write each item to disk
-            if output == "json":
-                file_path = os.path.join(compile_path, "%s.%s" % (item_key, output))
-                with CompiledFile(
-                    file_path,
-                    self.ref_controller,
-                    mode="w",
-                    reveal=reveal,
-                    target_name=target_name,
-                    indent=indent,
-                ) as fp:
-                    fp.write_json(item_value)
-            elif output == "yaml":
-                file_path = os.path.join(compile_path, "%s.%s" % (item_key, "yml"))
-                with CompiledFile(
-                    file_path,
-                    self.ref_controller,
-                    mode="w",
-                    reveal=reveal,
-                    target_name=target_name,
-                    indent=indent,
-                ) as fp:
-                    fp.write_yaml(item_value)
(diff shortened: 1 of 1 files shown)


ademariag:
BTW even with this patch, with `prune: false` Kapitan crashed quite badly. This PR does not address that issue, which I will re-open in another bug report

uberspot:
Crashed for which use case? We can't merge a crashing PR.  :P

ademariag:
The crash is unrelated to this PR. It already exists in the current codebase.
Will open another bug report :)

ramaro:
@ademariag can we add tests to cover what this fixes?

PR #508, opened 2020-05-08
Remote inventory proposal

AlphaRoy14:
# query related to .kapitan file 
In the proposal, I mentioned that `.kapitan`can be used to store the remote inventory configuration. From the codebase, I could figure out that `.kapitan` is used to store the arguments of various flags as well. However, after compilation when I list the items (` $ ls -al`) of the root directory  of the project I don't see any `.kapitan` file. Where can I find the `.kapitan` file so that I can manually see its contents?
# query related to command line arguments
## The proposed method of adding remote inventory details from cli is as follows

The command for adding the remote inventories from the cli and fetching it will be:
``` $ kapitan inventory --fetch <output-path> <source> ```
The command for only adding remote inventories to the .kapitan file from cli will be:
```$ kapitan inventory --add <inventory-name> <output-path> <source> ```
The command for fetching all the inventory declared in .kapitan file will be:
```$ kapitan inventory --fetch-all```
## Alternative 
Does it make more sense to have ``` $ kapitan inventory --fetch <inventory-name>  ``` instead of ``` $ kapitan inventory --fetch <output-path> <source> ``` ?
This way we can add the remote inventories using ```$ kapitan inventory --add <inventory-name> <output-path> <source> ``` and fetch it using ``` $ kapitan inventory --fetch <inventory-name>  ```. 
We can also force fetch an existing inventory using `$ kapitan inventory --fetch <inventory-name> --force`.

### DIFF
--- a/docs/kap_proposals/kap_7_remote_inventory.md
+++ b/docs/kap_proposals/kap_7_remote_inventory.md
@@ -19,11 +19,17 @@ remoteInventory:
         output_path: <relative_output_path>
         ... type specific arguments
 ```
+
 The command for adding the remote inventories from the cli and fetching it will be:
+
 ``` $ kapitan inventory --fetch <output-path> <source> ```
+
 The command for only adding remote inventories to the .kapitan file from cli will be:
+
 ```$ kapitan inventory --add <inventory-name> <output-path> <source> ```
+
 The command for fetching all the inventory declared in .kapitan file will be:
+
 ```$ kapitan inventory --fetch-all```
 
 ## Copying inventory files to the output location
--- a/docs/kap_proposals/kap_7_remote_inventory.md
+++ b/docs/kap_proposals/kap_7_remote_inventory.md
@@ -2,7 +2,6 @@
 
 This feature would add the ability to Kapitan to fetch parts of the inventory from remote locations (https/git). This would allow users to combine different inventories from different sources and build modular infrastructure reusable across various repos.
 
-
 Author: @alpharoy14
 
 ## Specification
@@ -19,19 +18,17 @@ remoteInventory:
         output_path: <relative_output_path>
         ... type specific arguments
 ```
-The command for adding remote inventories to the `.kapitan` file from cli will be:
-```$ kapitan inventory --add <inventory-name> <output-path> <source> ```
-The command for fetching remote inventories that is added to the `.kapitan` file will be:
-``` $ kapitan inventory --fetch <inventory-name> ```
-The command for fetching all the inventory declared in .kapitan file will be:
 
+The command for fetching a specific remote inventory declared in inventory configuration files will be:
+``` $ kapitan inventory --fetch <inventory-name> ```
+
+The command for fetching all the inventories declared in inventory configuration files will be:
 ```$ kapitan inventory --fetch-all```
 
 ## Copying inventory files to the output location
 The output path is the path to save the inventory items into. The path is relative to the `inventory/` directory. For example, it could be `/classes/`. The contents of the fetched inventory will be recursively copied.
 
-The fetched inventory files will be cached in a temp directory.
-The temp directory (`remote_inventories/`) will be in the path pointed by the $XDG_CACHE_HOME environment variable. In case that does not exist, we can check the `$KAPITAN_CACHE_DIR`. Which, by default, will be set to the root folder of the project.
+The fetched inventory files will be cached in the `.kapitan_cache` directory.
 
 ## Force fetching
 While fetching, the output path will be recursively checked to see if it contains any file with the same name. If so, kapitan will raise an error saying so.
--- a/docs/kap_proposals/kap_7_remote_inventory.md
+++ b/docs/kap_proposals/kap_7_remote_inventory.md
@@ -1,0 +1,62 @@
+# Remote Inventory Federation
+
+This feature would add the ability to Kapitan to fetch parts of the Inventory from remote locations (https/git). This would allow users to combine different inventories from different sources and build modular infrastructure reusable across various repos.
+
+
+Author: @alpharoy14
+
+## Specification
+
+The configuration and declaration of remote inventories would be done in the hidden file `​.kapitan`
+
+The file specifications are as follows:
+
+```yaml
+remoteInventory:
+    -   name: <inventory_name> #user defined
+        type: <inventory_type> #git\https
+        source: <source_of_inventory>
+        output_path: <relative_output_path>
+        ... type specific arguments
+```
+The command for adding the remote inventories from the cli and fetching it will be:
+``` $ kapitan inventory --fetch <output-path> <source> ```
+he command for only adding remote inventories to the .kapitan file from cli will be:
+```$ kapitan inventory --add <output-path> <source> ```
+The command for fetching all the inventory declared in .kapitan file will be:
+```$ kapitan inventory --fetch-all```
+
+## Copying inventory files to the output location
+The output path is the path to save the invertory intems into. The path is relative to the `inventory/` directory. For example, it could be `/classes/`. The contents of the fetched inventory will be reccursively copied.
+
+The fetched inveltory files will be cached in a temp directory.
+The temp directory (`remote_inventories/`) will be in the path pointed by the $XDG_CACHE_HOME environment variable. In case that does not exist, we can check the `$KAPITAN_CACHE_DIR`. Which, by default, will be set to the root folder of the project.
+
+## Force fetching
+While fetching, the output path will be recursively checked to see if it contains any file with the same name. If so, kapitan will raise an error saying so.
+
+To overwrite the files with the newly downloaded inventory items, we can add the --force flag to the fetch command as shown below.
+
+`$ kapitan inventory --fetch-all --force`
+`$ kapitan inventory --fetch <output-path> <source> --force`
+
+## URL type
+The URL type (Http/git) will be infered from the <source> address. Depending on the URL type, the configuration file may have aditional arguments.
+
+E.g Git type may also include aditional `ref` parameter as illustrated below:
+
+```yaml
+remoteInventory:
+    -   name: <inventory_name> #user defined
+        type: <inventory_type> #git\https
+        source: <source_of_inventory>
+        output_path: <relative_output_path>
+        ref: <commit_hash/branch/tag>
+```
+
+## Implementation details
+
+### Dependencies
+
+- [GitPython](https://github.com/gitpython-developers/GitPython) module (and git executable) for git type
+- requests module for http[s]
--- a/docs/kap_proposals/kap_7_remote_inventory.md
+++ b/docs/kap_proposals/kap_7_remote_inventory.md
@@ -6,24 +6,21 @@ Author: @alpharoy14
 
 ## Specification
 
-The configuration and declaration of remote inventories would be done in the hidden file `​.kapitan`.
+The configuration and declaration of remote inventories would be done in the inventory files.
 
 The file specifications are as follows:
 
 ```yaml
-remoteInventory:
-    -   name: <inventory_name> #user defined
-        type: <inventory_type> #git\https
-        source: <source_of_inventory>
-        output_path: <relative_output_path>
-        ... type specific arguments
+parameters:
+ kapitan:
+  inventory:
+   - type: <inventory_type> #git\https
+     source: <source_of_inventory>
+     output_path: <relative_output_path>
 ```
 
-The command for fetching a specific remote inventory declared in inventory configuration files will be:
-``` $ kapitan inventory --fetch <inventory-name> ```
+On executing the ``` $ kapitan compile``` command, first the remote inventories will be fetched followed by fetching of external dependencies and finally reclass will be used to compile.
 
-The command for fetching all the inventories declared in inventory configuration files will be:
-```$ kapitan inventory --fetch-all```
 
 ## Copying inventory files to the output location
 The output path is the path to save the inventory items into. The path is relative to the `inventory/` directory. For example, it could be `/classes/`. The contents of the fetched inventory will be recursively copied.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by AlphaRoy14** [comment]
@googlebot I signed it!

File: docs/kap_proposals/kap_7_remote_inventory.md [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+## Force fetching
+While fetching, the output path will be recursively checked to see if it contains any file with the same name. If so, kapitan will raise an error saying so.
+
+To overwrite the files with the newly downloaded inventory items, we can add the --force flag to the fetch command as shown below.
+
+`$ kapitan inventory --fetch-all --force`
+`$ kapitan inventory --fetch <output-path> <source> --force`
+
+## URL type
+The URL type (Http/git) will be infered from the <source> address. Depending on the URL type, the configuration file may have aditional arguments.
```
uberspot, line 50: `inferred` and `additional` *

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,68 @@
...
+```$ kapitan inventory --add <inventory-name> <output-path> <source> ```
+
+The command for fetching all the inventory declared in .kapitan file will be:
+
+```$ kapitan inventory --fetch-all```
+
+## Copying inventory files to the output location
+The output path is the path to save the invertory intems into. The path is relative to the `inventory/` directory. For example, it could be `/classes/`. The contents of the fetched inventory will be reccursively copied.
+
+The fetched inveltory files will be cached in a temp directory.
```
uberspot, line 38: `inventory` *

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,68 @@
...
+remoteInventory:
+    -   name: <inventory_name> #user defined
+        type: <inventory_type> #git\https
+        source: <source_of_inventory>
+        output_path: <relative_output_path>
+        ref: <commit_hash/branch/tag>
+```
+
+## Implementation details
+
```
uberspot, line 64: Looks good, I'd just add a TODO here under `Implementation details` to fill in later on, once more details have been ironed out. :)

uberspot:
```
Does it make more sense to have $ kapitan inventory --fetch <inventory-name> instead of $ kapitan inventory --fetch <output-path> <source> ?
```
Personally, I think yes. It makes more sense to always add an inventory, and then have a separate step that fetches it (sort of similar to how git submodules operate), than fetching and implicitly adding.  Also, given this is still a prototype, if we see that these arguments aren't as usable as expected we can easily iterate and improve them before adding this feature in a stable release. :) 

```
Where can I find the .kapitan file so that I can manually see its contents?
```
That file is just a yaml file, constructed by users themselves. 
An example would be something like this:

```
version: 0.27
compile:
    parallelism: 12
```

Each leaf-key corresponds to a cmd argument/command. and each node (like "compile") would be a command. All keys in this .kapitan file are parsed by the `from_dot_kapitan` method so you can see all available keys if you just grep for that through the source code:

```
$ grep -ir "from_dot_kapitan" kapitan/
kapitan/utils.py:def from_dot_kapitan(command, flag, default):
kapitan/cli.py:    from_dot_kapitan,
kapitan/cli.py:        default=from_dot_kapitan("eval", "output", "yaml"),
kapitan/cli.py:        default=from_dot_kapitan("eval", "vars", []),
kapitan/cli.py:        default=from_dot_kapitan("eval", "search-paths", ["."]),
kapitan/cli.py:        default=from_dot_kapitan("compile", "search-paths", [".", "lib"]),
kapitan/cli.py:        default=from_dot_kapitan("compile", "jinja2-filters", defaults.DEFAULT_JINJA2_FILTERS_PATH),
kapitan/cli.py:        default=from_dot_kapitan("compile", "verbose", False),
kapitan/cli.py:        default=from_dot_kapitan("compile", "prune", False),
kapitan/cli.py:        default=from_dot_kapitan("compile", "quiet", False),
kapitan/cli.py:        default=from_dot_kapitan("compile", "output-path", "."),
kapitan/cli.py:        default=from_dot_kapitan("compile", "fetch", False),
kapitan/cli.py:        default=from_dot_kapitan("compile", "validate", False),
kapitan/cli.py:        default=from_dot_kapitan("compile", "parallelism", 4),
kapitan/cli.py:        default=from_dot_kapitan("compile", "indent", 2),
kapitan/cli.py:        default=from_dot_kapitan("compile", "refs-path", "./refs"),
kapitan/cli.py:        default=from_dot_kapitan("compile", "reveal", False),
kapitan/cli.py:        default=from_dot_kapitan("compile", "embed-refs", False),
kapitan/cli.py:        default=from_dot_kapitan("compile", "inventory-path", "./inventory"),
kapitan/cli.py:        default=from_dot_kapitan("compile", "cache", False),
kapitan/cli.py:        default=from_dot_kapitan("compile", "cache-paths", []),
kapitan/cli.py:        default=from_dot_kapitan("compile", "ignore-version-check", False),
kapitan/cli.py:        default=from_dot_kapitan("validate", "schemas-path", "./schemas"),
kapitan/cli.py:        default=from_dot_kapitan("compile", "targets", []),
kapitan/cli.py:        default=from_dot_kapitan("compile", "labels", []),
kapitan/cli.py:        default=from_dot_kapitan("inventory", "target-name", ""),
kapitan/cli.py:        default=from_dot_kapitan("inventory", "inventory-path", "./inventory"),
kapitan/cli.py:        default=from_dot_kapitan("inventory", "flat", False),
kapitan/cli.py:        default=from_dot_kapitan("inventory", "pattern", ""),
kapitan/cli.py:        default=from_dot_kapitan("inventory", "verbose", False),
kapitan/cli.py:        default=from_dot_kapitan("searchvar", "inventory-path", "./inventory"),
kapitan/cli.py:        default=from_dot_kapitan("searchvar", "verbose", False),
kapitan/cli.py:        default=from_dot_kapitan("searchvar", "pretty-print", False),
kapitan/cli.py:        default=from_dot_kapitan("refs", "update-targets", False),
kapitan/cli.py:        default=from_dot_kapitan("refs", "validate-targets", False),
kapitan/cli.py:        default=from_dot_kapitan("refs", "base64", False),
kapitan/cli.py:        default=from_dot_kapitan("refs", "reveal", False),
kapitan/cli.py:        default=from_dot_kapitan("refs", "inventory-path", "./inventory"),
kapitan/cli.py:        default=from_dot_kapitan("refs", "recipients", []),
kapitan/cli.py:        "--key", "-K", help="set KMS key", default=from_dot_kapitan("refs", "key", ""), metavar="KEY"
kapitan/cli.py:        default=from_dot_kapitan("refs", "vault-auth", ""),
kapitan/cli.py:        default=from_dot_kapitan("refs", "refs-path", "./refs"),
kapitan/cli.py:        default=from_dot_kapitan("refs", "verbose", False),
kapitan/cli.py:        default=from_dot_kapitan("lint", "fail-on-warning", False),
kapitan/cli.py:        default=from_dot_kapitan("lint", "skip-class-checks", False),
kapitan/cli.py:        default=from_dot_kapitan("lint", "skip-yamllint", False),
kapitan/cli.py:        default=from_dot_kapitan("lint", "search-secrets", False),
kapitan/cli.py:        default=from_dot_kapitan("lint", "refs-path", "./refs"),
kapitan/cli.py:        default=from_dot_kapitan("lint", "compiled-path", "./compiled"),
kapitan/cli.py:        default=from_dot_kapitan("lint", "inventory-path", "./inventory"),
kapitan/cli.py:        default=from_dot_kapitan("init", "directory", "."),
kapitan/cli.py:        default=from_dot_kapitan("compile", "compiled-path", "./compiled"),
kapitan/cli.py:        default=from_dot_kapitan("compile", "inventory-path", "./inventory"),
kapitan/cli.py:        default=from_dot_kapitan("compile", "targets", []),
kapitan/cli.py:        default=from_dot_kapitan("validate", "schemas-path", "./schemas"),
kapitan/cli.py:        default=from_dot_kapitan("validate", "parallelism", 4),

```

docs/kap_proposals/kap_7_remote_inventory.md
AlphaRoy14: Thanks for the feedback. I have updated the proposal.

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,64 @@
+# Remote Inventory Federation
+
+This feature would add the ability to Kapitan to fetch parts of the inventory from remote locations (https/git). This would allow users to combine different inventories from different sources and build modular infrastructure reusable across various repos.
+
+
+Author: @alpharoy14
+
+## Specification
+
+The configuration and declaration of remote inventories would be done in the hidden file `​.kapitan`.
```
ramaro, line 10: I think we should leave the `.kapitan` reserved for setting default kapitan flag values only

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,64 @@
...
+
+The file specifications are as follows:
+
+```yaml
```
ramaro, line 14: How about using this spec instead in the inventory itself? This could go under the `inventory` key (since nothing else uses it, to me is makes it very clear that its remote. It would avoid duplicating similar functionality we already implemented on top of the inventory and you could keep the semantics you've suggested later on in your proposal.

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,64 @@
...
+remoteInventory:
+    -   name: <inventory_name> #user defined
+        type: <inventory_type> #git\https
+        source: <source_of_inventory>
+        output_path: <relative_output_path>
+        ... type specific arguments
+```
+The command for adding remote inventories to the `.kapitan` file from cli will be:
```
ramaro, line 22: I think we can leave this feature behind for now. Maybe later we can revisit a more generic way of managing `.kapitan` with tooling. For now we expect users to manually edit the file.

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,64 @@
...
+```$ kapitan inventory --add <inventory-name> <output-path> <source> ```
+The command for fetching remote inventories that is added to the `.kapitan` file will be:
+``` $ kapitan inventory --fetch <inventory-name> ```
```
ramaro, line 25: `--fetch` would still make sense if we move the remote spec into the inventory

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,64 @@
...
+The command for fetching all the inventory declared in .kapitan file will be:
+
+```$ kapitan inventory --fetch-all```
```
ramaro, line 28: likewise

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Copying inventory files to the output location
+The output path is the path to save the inventory items into. The path is relative to the `inventory/` directory. For example, it could be `/classes/`. The contents of the fetched inventory will be recursively copied.
+
+The fetched inventory files will be cached in a temp directory.
+The temp directory (`remote_inventories/`) will be in the path pointed by the $XDG_CACHE_HOME environment variable. In case that does not exist, we can check the `$KAPITAN_CACHE_DIR`. Which, by default, will be set to the root folder of the project.
```
ramaro, line 34: Maybe we can piggyback on the `.kapitan_cache` directory we already use for compile? We would make things more consistent.

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Force fetching
+While fetching, the output path will be recursively checked to see if it contains any file with the same name. If so, kapitan will raise an error saying so.
+
+To overwrite the files with the newly downloaded inventory items, we can add the --force flag to the fetch command, as shown below.
+
+`$ kapitan inventory --fetch-all --force`
+`$ kapitan inventory --fetch <inventory-name> --force`
```
ramaro, line 42: Just a thought: force fetching makes sense. But would this be the only option? Could we offer an interactive way of dealing with remote files (e.g. accept remote file instead, keep local file instead, etc...)

docs/kap_proposals/kap_7_remote_inventory.md
AlphaRoy14: I believe this will become something similar to external dependency management. The target files would have an option of `inventory` key. Something like 
``` yaml
parameters:
  kapitan:
    inventory:
    - name: <inventory_name>
      type: <inventory_type>
      output_path: relative/path/to/file/or/dir
      source: <source_of_inventory>
```
The differentiating factor from remote dependency would be using ` $ kapitan inventory --fetch <inventory-name> `  before compiling. This way the user can edit the fetched inventory configurations. Did I understand it correctly?

docs/kap_proposals/kap_7_remote_inventory.md
AlphaRoy14: Yes I think this will be a better alternative. I'll update the proposal.

docs/kap_proposals/kap_7_remote_inventory.md
AlphaRoy14: I mentioned in the proposal that Kapitan would check if a file with the same name exists in the output path and would display a message to the user saying so. The user can then use the `--force` flag if they want to replace the existing file. Would you suggest that we have a boolean key in the configuration instead? 
eg 
```yaml
parameters:
  kapitan:
    inventory:
    - name: <inventory_name>
      type: <inventory_type>
      output_path: relative/path/to/file/or/dir
      source: <source_of_inventory>
      force: True
```
Or we can display a message (eg File already exists, would you like to force fetch it?) to which the user can respond with Y/N. Please let me know if there's a better way to make it more interactive.

AlphaRoy14:
@ramaro I have made the changes that you suggested related to caching and cli. I will update the rest of the proposal as soon as you approve the changes related to the storage of the configuration and interactive way of dealing with remote inventories (as discussed in the comments) :)

docs/kap_proposals/kap_7_remote_inventory.md
uberspot: So instead would a solution be to define a new `.kapitanmodules` file for example?

docs/kap_proposals/kap_7_remote_inventory.md
uberspot: @ramaro how will that work if inventory fetching is done in a separate step before running reclass?
We'd have to build the inventory twice. Once before fetching external deps and once after fetching them? 
Or am I missing something?

docs/kap_proposals/kap_7_remote_inventory.md
ramaro: I think using the inventory to add remotes would make more sense

docs/kap_proposals/kap_7_remote_inventory.md
ramaro: We will only need to run it twice if there's something new to fetch - which we can detect, right?

ramaro: And yes, the other option is to make fetching manual using `--fetch`. But that wouldn't be consistent with the behaviour of `kapitan compile --fetch`, so I guess it makes sense to have `--fetch` for force fetch only if we implement a way of detecting that an remote inventory hasn't been fetched at all (e.g. check the output_path doesn't exist)

docs/kap_proposals/kap_7_remote_inventory.md
ramaro: Just a thought really! I think we can park this for now and revisit when we have the main feature done.

ramaro:
@AlphaRoy14 thanks for this! I added a new comment which I'd like us to be in consensus, but other than that, it looks great.

docs/kap_proposals/kap_7_remote_inventory.md
AlphaRoy14: Just a thought: manual fetching might turn out to be useful as users might want to tweak certain parameters before compiling. Besides this, if the entire process is automated then we might face the problem of infinite loop if the inventory is force compiled (using the `--fetch` flag).

ramaro: I think that could be easy to detect if it happens! Willing to take the risk for the sake of simplicity now. We can revisit if it is really an issue.

AlphaRoy14: great! So the way I understand it, it's similar to how external dependency is implemented except for the case of remote inventory, the `inventory` key will be parsed first to download and save the inventories in the `output_path` and then the `dependency` key will be parsed to download the remote dependencies. Did I miss something? If not, should I update this in the proposal and start with the implementation?

uberspot: That sgtm! :)

PR #520, opened 2020-06-08
Helm dependency

sebradloff:
## Proposed Changes

  - Introduces a new external dependency of type `helm`. Allows the user to download a helm chart and its subcharts.
  - Upgrades to golang 1.14.


ramaro:
Thanks for the great contribution! Reviewing...

kapitan/dependency_manager/base.py
```diff
@@ -13,12 +13,18 @@
...
-from kapitan.errors import GitSubdirNotFoundError
+from kapitan.errors import GitSubdirNotFoundError, HelmBindingUnavailableError
 from kapitan.utils import make_request
 
 from git import Repo
 
+try:
+    from kapitan.dependency_manager.helm.helm_fetch_binding import ffi
+except ImportError:
+    pass  # make this feature optional
```
uberspot, line 26: Add a logger.debug message here before "pass"

kapitan/dependency_manager/base.py
```diff
@@ -178,3 +189,56 @@ def fetch_http_source(source, save_dir):
...
+            logger.info(
+                "Dependency helm chart {} and version {}: saved to {}".format(
+                    chart_name, version, destination_dir
+                )
+            )
+
+
+def initialise_helm_fetch_binding():
+    """returns the dl_opened library (.so file) if exists, otherwise None"""
+    if platform.system() not in ("Linux", "Darwin"):  # TODO: later add binding for Mac
```
uberspot, line 231: Since Darwin is in this list, is this TODO comment outdated?

kapitan/dependency_manager/helm/build.sh
```diff
@@ -0,0 +1,31 @@
...
+if [[ -e $so_name ]]; then
+    echo "[INFO] $so_name built successfully or already exists"
+else
+    echo "[ERROR] $so_name does not exist!"
+    exit 1
+fi
+
+# Compile the Python ffi binding if Python is available
+if [[ -z $(which python3) ]]; then
+    echo "[WARN] python3 is not available on this system -- skipping cffi build!"
```
uberspot, line 25: Since these build steps are part of the CI/CD kapitan docker image, I'd say if this fails to build, the script should fail completely (?). That way travis correctly flags the build as broken. Otherwise the .so file won't build and the final released artifact will partially fail.

kapitan/dependency_manager/base.py
sebradloff: My understanding is that the binding is still not available on Mac as of version `0.27.4`. I think that's what this PR from @yoshi-1224 is trying to achieve: https://github.com/deepmind/kapitan/pull/414

kapitan/dependency_manager/helm/build.sh
sebradloff: Agreed. I was trying to stay consistent with the build script `kapitan/inputs/helm/build.sh` where I drew inspiration for using cgo in the PR. I'll add a `set -euo pipefail` at the top of this script.

kapitan/dependency_manager/base.py
uberspot: Oh I see, I got confused with the actual comment and misread the if statement. :)

uberspot:
It would be great to include along with this PR an update to https://github.com/deepmind/kapitan/blob/master/docs/kap_proposals/kap_2_helm_charts_input_type.md so that it reflects the new input capabilities after this is merged? 
To make sure it's rolled out at the same time as the code.

sebradloff:
> It would be great to include along with this PR an update to https://github.com/deepmind/kapitan/blob/master/docs/kap_proposals/kap_2_helm_charts_input_type.md so that it reflects the new input capabilities after this is merged?
> To make sure it's rolled out at the same time as the code.

@uberspot Agreed. I'd like to add to the dependency manger documentation and provide a "complete" helm example in the `examples` directory.

PR #521, opened 2020-06-09
Add support for remote inventory fetching

AlphaRoy14:
Fixes issue #508 

This PR is to track the status of the GSoC task of Remote Inventory Federation and also to get feedback on the implementation.

## TODO
  - ~recursive inventory building~
  - ~testing fetching from git sources~
  - ~HTTP[s] fetching implementation~
  - ~testing fetching and unpacking from http sources~
  - ~force fetch~
  - ~caching~


AlphaRoy14:
@uberspot @ramaro `--fetch` was decided to add the functionality of force fetching as [discussed] (https://github.com/deepmind/kapitan/pull/508#discussion_r431948592) in the proposal. However, it is compulsory to use `--fetch` to fetch external dependencies. To keep it consistent with external dependency management, I have made it compulsory to use `--fetch` for fetching remote inventories as well for now.

srueg:
This feature sounds awesome!

The main reason we built a wrapper around Kapitan was to build the inventory out of multiple Git repos: https://github.com/projectsyn/commodore

So if this would land in Kapitan natively we could get rid of a lot of the logic we implemented 🎉 


Would this design allow to specify remote inventories recursively? I.e. specify another remote inventory in one of the remote inventories?
I'm wondering because as far as I understood reclass can only parse the inventory once the complete tree is available. And if you don't parse the inventory with reclass, how would the hierarchical overwriting of parameters work for the `inventory` dict?

cc @simu

AlphaRoy14:
Hi @srueg 

Glad you liked the idea!
It will be designed to recursively fetch remote inventories followed by fetching external dependencies (if there are any) and finally will use reclass to compile. You can read more about the idea [here](https://github.com/deepmind/kapitan/blob/master/docs/kap_proposals/kap_7_remote_inventory.md).

srueg:
But won't the call to `load_target_inventory()` [here](https://github.com/deepmind/kapitan/pull/521/files#diff-280d4028c0e294072ab70d24452f7b8bR77) load the inventory with reclass which would fail if parts of the inventory are not available yet since they're remote and not fetched yet?

Sorry if it's a stupid questions, I'm just trying to understand how the inventory can be recursively without breaking reclass.

uberspot:
> But won't the call to `load_target_inventory()` [here](https://github.com/deepmind/kapitan/pull/521/files#diff-280d4028c0e294072ab70d24452f7b8bR77) load the inventory with reclass which would fail if parts of the inventory are not available yet since they're remote and not fetched yet?
> 
> Sorry if it's a stupid question, I'm just trying to understand how the inventory can be recursively without breaking reclass.

Correct, we might have to do a pre-fetch step before passing anything from the inventory to reclass if possible and then, after fetching, call inventory parsing. Alternatively a more crude way of doing it would be to strip remove inventory info first, then fetch external inventories then resend it to reclass to get the full inventory. The first approach might be simpler.

AlphaRoy14:
> > But won't the call to `load_target_inventory()` [here](https://github.com/deepmind/kapitan/pull/521/files#diff-280d4028c0e294072ab70d24452f7b8bR77) load the inventory with reclass which would fail if parts of the inventory are not available yet since they're remote and not fetched yet?
> > Sorry if it's a stupid question, I'm just trying to understand how the inventory can be recursively without breaking reclass.
> 
> Correct, we might have to do a pre-fetch step before passing anything from the inventory to reclass if possible and then, after fetching, call inventory parsing. Alternatively a more crude way of doing it would be to strip remove inventory info first, then fetch external inventories then resend it to reclass to get the full inventory. The first approach might be simpler.

If we intend to call reclass only once after all the remote inventories are fetched, then I think we'll need to have a centralized inventory configuration e.g only the target files under `./inventory/target` and we cannot have inventories declared in classes. Then we can parse the target files to fetch the inventories. Should we assume for now that inventories will only be declared in target files?

tests/test_remote_inventory/environment_one/inventory/targets/remoteinv-example.yml
```diff
@@ -0,0 +1,15 @@
+parameters:
+    kapitan:
+        vars:
+            target: remoteinv-example
+        inventories:
+            - type: git
+              output_path: targets/
+              source: https://github.com/deepmind/kapitan.git
+              ref: master
+              subdir: kapitan/tests/test_remote_inventory/environment_two/inventory/targets/
```
AlphaRoy14, line 10: @uberspot @ramaro Hi! To test the recursive fetching and compiling of inventories I created two dummy directories `kapitan/tests/test_remote_inventory/environment_one` and `kapitan/tests/test_remote_inventory/environment_two`. The target files of environment_one fetches the target files of  environment_two and the target files of environment_two fetches `kapitan/tests/test_resources/inventory/`. This raises an [error on Travis](https://travis-ci.org/github/deepmind/kapitan/builds/698962601) as after cloning https://github.com/deepmind/kapitan.git, there's no directory called `kapitan/tests/test_remote_inventory/environment_two`. Is making a separate PR for the dummy environments and then testing it on Travis the only way? Or would it be better if I host the dummy environments on my github for testing and later make a PR?

uberspot: You can definitely create the two dirs in a separate PR with just the test resources.
Just add a small readme.md file in those dirs explaining what they are for if that's easy? 
So it's more intuitive. :)

kapitan/remoteinventory/fetch.py
```diff
@@ -38,7 +46,15 @@ def fetch_inventories(inventory_path, target_objs, pool):
...
                 inv_type = inv["type"]
                 source_uri = inv["source"]
-                if source_uri in cached.inv_sources:
+                source_hash = hashlib.sha256(source_uri.encode())
+                # hashing the source and subdir together for git sources
+                # as different inventory items can have the same git uri
+                try:
+                    subdir = inv["subdir"]
+                    source_hash.update(subdir.encode())
+                except KeyError:
```
uberspot, line 55: before the `pass` make sure to always add a logger.debug("") message to make debugging easier. That is then displayed when devs run kapitan with `--verbose`

tests/test_remote_inventory.py
```diff
@@ -45,9 +58,11 @@ def test_compile_fetch(self):
             temp_output,
             "--inventory-path",
             temp_inv,
-            "-t",
+            "--targets",
             "remoteinv-example",
             "remoteinv-nginx",
+            "-p",
```
uberspot, line 64: we usually don't add parallelism options in tests. Is this to test fetching remote invs with parrallel threads?
If yes then it might be better added in an explicit test for that. 
If you want to make your local tests faster, you could add this parallelism option in your local `.kapitan` config to apply on all kapitan projects? :)

AlphaRoy14:
Hi @uberspot @ramaro, I encountered a bug while trying to compile with the `--inventory-path` flag. 

## command 
`/bin/kapitan compile --inventory-path=/Users/roy/Desktop/inv_test --output-path=/Users/roy/Desktop/OUTPUT -t zippedinv` 
## output
```
$ ../../../bin/kapitan compile --inventory-path=/Users/roy/Desktop/inv_test --output-path=/Users/roy/Desktop/OUTPUT -t zippedinv
"""
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/pool.py", line 121, in worker
  File "/Users/roy/Desktop/GIT/kapitan/kapitan/targets.py", line 422, in compile_target
    input_compiler.compile_obj(comp_obj, ext_vars, **kwargs)
  File "/Users/roy/Desktop/GIT/kapitan/kapitan/inputs/base.py", line 52, in compile_obj
    self.compile_input_path(input_path, comp_obj, ext_vars, **kwargs)
  File "/Users/roy/Desktop/GIT/kapitan/kapitan/inputs/base.py", line 74, in compile_input_path
    **kwargs,
  File "/Users/roy/Desktop/GIT/kapitan/kapitan/inputs/jinja2.py", line 35, in compile_file
    context["inventory"] = inventory(self.search_paths, target_name)
  File "/Users/roy/Desktop/GIT/kapitan/kapitan/resources.py", line 270, in inventory
    return inventory_reclass(full_inv_path)["nodes"][target]
KeyError: 'zippedinv'
"""

The above exception was the direct cause of the following exception:

  File "/Users/roy/Desktop/GIT/kapitan/kapitan/targets.py", line 103, in compile_targets
    [p.get() for p in pool.imap_unordered(worker, target_objs) if p]
  File "/Users/roy/Desktop/GIT/kapitan/kapitan/targets.py", line 103, in <listcomp>
    [p.get() for p in pool.imap_unordered(worker, target_objs) if p]
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/pool.py", line 748, in next
    raise value
KeyError: 'zippedinv'


'zippedinv'
```
I figured it was because the `inventory_path` was not being passed to the `kapitan.targets.compile_target()` function. I attempted to fix the issue and the changes can be viewed in the second last commit `fixed issue with compiling with --inventory-path flag`.

kapitan/errors.py
```diff
@@ -70,6 +70,12 @@ class GitSubdirNotFoundError(KapitanError):
     pass
 
 
+class GitFetchingError(KapitanError):
+    """repo not found and permission error"""
```
uberspot, line 74: `repo not found and/or permission error` perhaps?

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+                    continue
+                else:
+                    inv_output_path[source_uri].add(output_path)
+
+                if inv_type == "git":
+                    git_inventories[source_uri].append(inv)
+                elif inv_type in ("http", "https"):
+                    http_inventories[source_uri].append(inv)
+        except KeyError:
+            continue
```
uberspot, line 66: Add logger.debug() output in exceptions before continuing. It helps debugging. :)

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+        repo_path = os.path.join(save_dir, os.path.basename(source))
+        repo = Repo(repo_path)
+        output_path = os.path.join(inventory_path, inv["output_path"])
+        copy_src_path = repo_path
+
+        if "ref" in inv:
+            ref = inv["ref"]
+            repo.git.checkout(ref)
+        else:
+            repo.git.checkout("master")
```
uberspot, line 99: Doublechecking here that we mention in documentation that inventory fetcher defaults to master branch?

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+    logger.info("Dependency {} : successfully fetched".format(source))
+    if content is not None:
+        basename = os.path.basename(source)
+        # to avoid collisions between basename(source)
+        path_hash = hashlib.sha256(os.path.dirname(source).encode()).hexdigest()[:8]
+        full_save_path = os.path.join(save_dir, path_hash + basename)
+        with open(full_save_path, "wb") as f:
+            f.write(content)
+        return content_type
+    else:
```
uberspot, line 179: You can remove this "else:" line here. 
If the `if content is not None:` line isn't true it will just get to the last two lines even without the else: condition.

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+                source_uri = inv["source"]
+                source_hash = hashlib.sha256(source_uri.encode())
+                # hashing the source and subdir together for git sources
+                if "subdir" in inv:
+                    subdir = inv["subdir"]
+                    source_hash.update(subdir.encode())
+
+                sources.append(source_hash.hexdigest()[:8])
+        except KeyError:
+            continue
```
uberspot, line 200: Add logger.debug() here too.

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+    for inv in invs:
+        output_path = os.path.join(inventory_path, inv["output_path"])
+        if inv.get("unpack", False):
+            # ensure that the directory we are extracting to exists
+            os.makedirs(output_path, exist_ok=True)
+            is_unpacked = unpack_downloaded_file(copy_src_path, output_path, content_type)
+            if is_unpacked:
+                logger.info("Inventory {} : extracted to {}".format(source, output_path))
+            else:
+                logger.info(
```
uberspot, line 150: I'd change this to logger.warning. Since it's most likely user error that they're trying to fetch an unsupported file and we want to make it clear this failed here.

tests/test_remote_inventory.py
```diff
@@ -0,0 +1,113 @@
...
+            os.path.join(
+                os.path.abspath(os.path.dirname(__file__)), "test_remote_inventory", "environment_one"
+            )
+        )
+
+    def test_fetch_git_inventory(self):
+        temp_dir = tempfile.mkdtemp()
+        git_source = "https://github.com/deepmind/kapitan.git"
+        fetch_git_source(git_source, temp_dir)
+        self.assertTrue(os.path.isdir(os.path.join(temp_dir, "kapitan.git", "kapitan")))
```
uberspot, line 30: Can we check here for a file instead? Maybe the README.md one. 
That will always exist and it's slightly safer as a test than checking for the directory since directories could be created in various other scenarios when using git too.

tests/test_remote_inventory.py
```diff
@@ -0,0 +1,113 @@
...
+
+        http_sources = [
+            "https://github.com/deepmind/kapitan/raw/master/tests/test_remote_inventory/zipped_inventories/inventory.7z",
+            "https://raw.githubusercontent.com/deepmind/kapitan/master/tests/test_resources/inventory/classes/common.yml",
+            "https://github.com/deepmind/kapitan/raw/master/tests/test_remote_inventory/zipped_inventories/inventory.zip",
+        ]
+        for source in http_sources:
+            fetch_http_source(source, temp_dir)
+        self.assertTrue(os.path.isfile(os.path.join(temp_dir, "47c29a3binventory.7z")))
+        # self.assertTrue(os.path.isfile(os.path.join(temp_dir,"009a21cfmaster.zip")))
```
uberspot, line 54: Is this a leftover?

tests/test_remote_inventory.py
```diff
@@ -0,0 +1,113 @@
...
+class RemoteInventoryTest(unittest.TestCase):
+    def setUp(self):
+        os.chdir(
+            os.path.join(
+                os.path.abspath(os.path.dirname(__file__)), "test_remote_inventory", "environment_one"
+            )
+        )
+
+    def test_fetch_git_inventory(self):
+        temp_dir = tempfile.mkdtemp()
```
uberspot, line 27: These tests look good in general. Just one comment, add a pythondoc to all these test_ functions with just a sentence about what each test does. I know we can read it from code, but if you check other test_*.py files you'll see pydoc comments on all functions making code searching for relevant tests much much simpler for devs. :)

uberspot:
Nice work so far Arindaam.
Besides the last remaining comments this seems to be very very close to completion. :100: 
Make sure to also rebase your branch to latest master once done. :+1:

kapitan/targets.py
```diff
@@ -78,12 +79,23 @@ def compile_targets(
...
             **kwargs,
         )
 
         if not target_objs:
             raise CompileError("Error: no targets found")
 
+        if kwargs.get("fetch_inventories", False):
+            # new_source checks for new sources in fetched inventory items
+            new_sources = list(set(list_sources(target_objs)) - cached.inv_sources)
+            while new_sources:
```
ramaro, line 92: I think it's cleaner to turn this into a for loop

kapitan/utils.py
```diff
@@ -471,3 +475,41 @@ def make_request(source):
     else:
         r.raise_for_status()
     return None, None
+
+
+def unpack_downloaded_file(file_path, output_path, content_type):
```
ramaro, line 480: I don't understand the need for this. Can't we use/extend https://github.com/deepmind/kapitan/blob/7b3a4af419e9716faa11d76a6f0bc258e1b33b8b/kapitan/dependency_manager/base.py#L132 instead?

requirements.txt
```diff
@@ -17,5 +17,6 @@ rfc3987==1.3.8
 GitPython==3.1.3
 hvac==0.10.4
 docker==4.2.1
+py7zr==0.7.4
```
ramaro, line 20: is supporting 7zip really necessary?

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -19,7 +19,7 @@ parameters:
      output_path: <relative_output_path>
 ```
 
-On executing the ``` $ kapitan compile``` command, first the remote inventories will be fetched followed by fetching of external dependencies and finally reclass will be used to compile.
+On executing the ``` $ kapitan compile --fetch``` command, first the remote inventories will be fetched followed by fetching of external dependencies and finally reclass will be used to compile.
```
ramaro, line 22: change to "and finally merge the inventory to compile"

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+    git_worker = partial(fetch_git_inventories, inventory_path=inventory_path, save_dir=temp_dir)
+    http_worker = partial(fetch_http_inventories, inventory_path=inventory_path, save_dir=temp_dir)
+    [p.get() for p in pool.imap_unordered(git_worker, git_inventories.items()) if p]
+    [p.get() for p in pool.imap_unordered(http_worker, http_inventories.items()) if p]
+
+    rmtree(temp_dir)
+    logger.debug("Removed {}".format(temp_dir))
+
+
+def fetch_git_inventories(inv_mapping, inventory_path, save_dir):
```
ramaro, line 77: Can't we reuse https://github.com/deepmind/kapitan/blob/7b3a4af419e9716faa11d76a6f0bc258e1b33b8b/kapitan/dependency_manager/base.py#L90 ?

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+    logger.info("Inventory {} : fetching now".format(source))
+    try:
+        Repo.clone_from(source, os.path.join(save_dir, os.path.basename(source)))
+        logger.info("Inventory {} : successfully fetched".format(source))
+    except GitCommandError as e:
+        logger.error(e)
+        raise GitFetchingError("Inventory {} : fetching unsuccessful{}".format(source, e.stderr))
+
+
+def fetch_http_inventories(inv_mapping, inventory_path, save_dir):
```
ramaro, line 127: Can't we reuse https://github.com/deepmind/kapitan/blob/7b3a4af419e9716faa11d76a6f0bc258e1b33b8b/kapitan/dependency_manager/base.py#L132 ?

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+    for target_obj in target_objs:
+        try:
+            inventories = target_obj["inventories"]
+            for inv in inventories:
+                inv_type = inv["type"]
+                source_uri = inv["source"]
+                source_hash = hashlib.sha256(source_uri.encode())
+                # hashing the source and subdir together for git sources
+                # as different inventory items can have the same git uri
+                if "subdir" in inv:
```
ramaro, line 48: how will this work when 'ref' is updated? We should take this into consideration

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,201 @@
...
+    "returns list of all remote inventory sources"
+    sources = []
+    for target_obj in target_objs:
+        try:
+            invs = target_obj["inventories"]
+            for inv in invs:
+                source_uri = inv["source"]
+                source_hash = hashlib.sha256(source_uri.encode())
+                # hashing the source and subdir together for git sources
+                if "subdir" in inv:
```
ramaro, line 194: also hash 'ref'?

kapitan/remoteinventory/fetch.py
AlphaRoy14: Thanks for pointing it out, will hash 'ref' along with 'type' and 'source'.

> how will this work when 'ref' is updated? We should take this into consideration

kapitan/remoteinventory/fetch.py
AlphaRoy14: > Can't we reuse
> 
> https://github.com/deepmind/kapitan/blob/7b3a4af419e9716faa11d76a6f0bc258e1b33b8b/kapitan/dependency_manager/base.py#L90
> 
> ?

For remote inventories we have the provision to save the inventories relative to the user-defined inventory path, which is passed as an argument in the `fetch_git_inventories()` function. In the external dependency implementation, however, the relative output path is only considered in the testing environment. https://github.com/deepmind/kapitan/blob/7b3a4af419e9716faa11d76a6f0bc258e1b33b8b/kapitan/dependency_manager/base.py#L59
Would you suggest that I update the external dependency functions to consider relative paths and reuse the code?

kapitan/targets.py
AlphaRoy14: Since it is not known how many times the loop should run (as fetched inventories can fetch other inventories) I decide to use a while loop. I might be missing something but shouldn't a for loop be better only if it looped for a fixed number of times?

kapitan/utils.py
AlphaRoy14: I thought of shifting the code from `kapitan/kapitan/dependency_manager/base.py` to the utils section as unpacking a downloaded file might be useful while implementing other functionalities in the future?

requirements.txt
AlphaRoy14: We can definitely remove it and document the supported archiving/compression formats.

tests/test_remote_inventory.py
AlphaRoy14: Thank you for the reviews, this was a test case for fetching an entire git directory. I'll add it back when I push a commit with all the changes.

requirements.txt
uberspot: It might be preferred to start without it to keep our dependencies low and if users ask for it explicitly we can easily re-add it. :)

kapitan/utils.py
```diff
@@ -502,3 +505,67 @@ def unpack_downloaded_file(file_path, output_path, content_type):
...
+    Similar to distutils.file_util.copy_file except
+    if the file exists in 'dst' it's not clobbered
+    or overwritten.
+
+    returns a tupple (src, val)
+    file not copied if val = 0 else 1
+    """
+
+    if not os.path.isfile(src):
+        raise DistutilsFileError("can't copy {}: doesn't exit or is not a regular file".format(src))
```
uberspot, line 522: exist*

kapitan/utils.py
```diff
@@ -471,3 +477,95 @@ def make_request(source):
...
+
+    returns a tupple (src, val)
+    file not copied if val = 0 else 1
+    """
+
+    if not os.path.isfile(src):
+        raise DistutilsFileError("can't copy {}: doesn't exist or is not a regular file".format(src))
+
+    if os.path.isdir(dst):
+        dir = dst
```
uberspot, line 525: The `dir` variable in this function seems to be unused. Why assign values to it?

AlphaRoy14: Thanks for pointing it out, the `dir` variable was to add a debugging message after copying the file. I'll add it.

AlphaRoy14:
Hi @uberspot @ramaro We decided to piggyback on `.kapitan_cache` to store the inventory and dependency cache earlier. Since `.kapitan_cache` is a file, I've updated the proposal and implemented the cache to be stored in `.dependency_cache` which will be created in the root directory of the project. Since the caches are local to the project directories, I was wondering if we should explore our alternatives considerations such as `$XDG_CACHE_HOME`, `$KAPITAN_CACHE_DIR`. Having a centralized user-specific cache directory would be more useful I feel.

adrianchifor:
> Hi @uberspot @ramaro We decided to piggyback on `.kapitan_cache` to store the inventory and dependency cache earlier. Since `.kapitan_cache` is a file, I've updated the proposal and implemented the cache to be stored in `.dependency_cache` which will be created in the root directory of the project. Since the caches are local to the project directories, I was wondering if we should explore our alternatives considerations such as `$XDG_CACHE_HOME`, `$KAPITAN_CACHE_DIR`. Having a centralized user-specific cache directory would be more useful I feel.

For now we can keep `.kapitan_cache` and `.dependency_cache` separate as they store different info and some people might not be using `--cache`. Root dir is fine for now, we can work on supporting different cache dirs if people request it.

AlphaRoy14:
@uberspot @ramaro All the features are implemented and it is ready for review :)

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,101 @@
...
+                    # if the output_path is duplicated for the same source_uri
+                    logger.warning("Skipping duplicate output path for uri {}".format(source_uri))
+                    continue
+                else:
+                    inv_output_path[source_uri].add(output_path)
+
+                if inv_type == "git":
+                    git_inventories[source_uri].append(inv)
+                elif inv_type in ("http", "https"):
+                    http_inventories[source_uri].append(inv)
```
uberspot, line 65: It might be good to have an "else:" condition here in case the user specifies any inv_type different than git/http/https just print out a logger.warning message. To catch people making typos. :)

AlphaRoy14: I think a typo will be caught during schema verification while loading target inventory. Regardless, I'll add the else condition just to be safe :) https://github.com/deepmind/kapitan/blob/2fce9462e83a2a6e01ee674c1a6b7766bb95bf5a/kapitan/targets.py#L72

docs/kap_proposals/kap_7_remote_inventory.md
```diff
@@ -19,20 +19,21 @@ parameters:
...
-On executing the ``` $ kapitan compile``` command, first the remote inventories will be fetched followed by fetching of external dependencies and finally reclass will be used to compile.
+On executing the ``` $ kapitan compile --fetch``` command, first the remote inventories will be fetched followed by fetching of external dependencies and finally merge the inventory to compile.
 
 
 ## Copying inventory files to the output location
 The output path is the path to save the inventory items into. The path is relative to the `inventory/` directory. For example, it could be `/classes/`. The contents of the fetched inventory will be recursively copied.
 
-The fetched inventory files will be cached in the `.kapitan_cache` directory.
+The fetched inventory files will be cached in the `.dependency_cache` directory if `--cache` is set.
+eg. `$ kapitan compile --fetche --cache`
```
ramaro, line 29: typo, "fetch" not "fetche"

kapitan/remoteinventory/fetch.py
```diff
@@ -0,0 +1,103 @@
...
+                source_uri = inv["source"]
+                source_hash = hashlib.sha256(source_uri.encode())
+                # hashing the source, subdir and ref together for git sources
+                # as different inventory items can have the same git uri
+                if "subdir" in inv:
+                    subdir = inv["subdir"]
+                    source_hash.update(subdir.encode())
+                if "ref" in inv:
+                    ref = inv["ref"]
+                    source_hash = hashlib.sha256(ref.encode())
```
ramaro, line 46: don't we want to update() instead it resetting?

Issue #522, opened 2020-06-09
Embedded refs not persisting original encoding metadata


ramaro:
Embedded refs not persisting original encoding metadata

**To Reproduce**
Create a ref with `--base64` (encoding: base64), compile with --embed-refs and reveal to see encoding will now be `original`

**Expected behavior**
Encoding metadata is persisted

** If it's a bug (please complete the following information):**
 - `python --version`: 3.8.3
 - `kapitan --version`: 0.24.7

PR #523, opened 2020-06-09
fix embedded refs encoding

ramaro:
Fixes issue #522


uberspot:
```
----- Testing code formatting -----
black -l 110 -t py37 --check --exclude "reclass|helm_binding.py|version.py" .
would reformat /home/travis/build/deepmind/kapitan/tests/test_refs.py
Oh no! 💥 💔 💥
1 file would be reformatted, 58 files would be left unchanged.
```

Issue #294, opened 2019-05-16
Reading values from environment variables


xydinesh:
**Describe the bug/feature**

I like to populate jsonnet template or yaml parameters using an environment variable before running `kapitan compile`.  We are using environment variable `CI_PIPELINE_ID` to create docker images and want to use this value with the image for deployment. For example,

```
image: xydinesh/testimage:${CI_PIPELINE_ID}
```

Is there a way to do this with jsonnet templates ? Or do I have to generate a yaml with this new image before running `kapitan compile`

```
local kap = import "lib/kapitan.libjsonnet";
local inv = kap.inventory();
local replicas = inv.parameters.nginx.replicas;
local image = inv.parameters.nginx.image;
local namespace = inv.parameters.namespace;
```

Thanks !

uberspot:
Related to https://github.com/deepmind/kapitan/issues/211 
Implementation should be the same/similar for both issues.
It's not a fully recommended practice to have configuration as environment vs as code, but the functionality makes sense so this feature could be implemented. :) 
An example implementation would parse the inventory after reclass and check for $strings. 
Env vars would have to be in that format and without curly braces {}  to avoid reclass interpreting them as inventory variables.

xydinesh:
Thanks ! If I can access `CI_PIPELINE_ID` environment variable to create correct docker image tag for deployment, I wouldn't worry about the syntax ({}). For now, I will use a python script to create new yaml file with this value.

ramaro:
The new Ref proposal https://github.com/deepmind/kapitan/pull/298 could solve this problem for you with the support for updating refs via the command line. You could write/update refs with the value of your envs vars.

frittentheke:
@uberspot the Docker image ID which is known just before Kapitan is run is exactly what brought me here. Being able to consume environment variables would be awesome ... just please make Kapitan require them if referenced and if not just fail  ...

zoek1:
hey @uberspot could i work on this as first good issue?

uberspot:
Hello, :) you can always just start work on an issue without confirmation. We don't strictly assign open issues to people.

zoek1:
Thanks, i would like participate on the GSoC with the Kapitan so i want tackle this to familiarize. :smiley:

AlphaRoy14:
[...]

PR #530, opened 2020-06-18
Env Refs Backend

farvour:
Fixes issue #294 

## Proposed Changes

  - Create a new ref-type called "env" which is basically a modified inherited version of the PlainRef that automatically consults the environment for a variable prefixed with `$KAPITAN_VAR_`, and uses the last slice of the refs path as the variable to look up.
  - If it can't find the value, the refs data itself is used as the "default" value for the reveal.

I haven't written a test case for this yet, would like to, but I tested this out with a couple of different samples using a ref that was just a top level path (like foobar) and nested (foo/bar). In each case it tries to look up `$KAPITAN_VAR_foobar` and `$KAPITAN_VAR_bar` respectively.

PS - slightly off topic, PR #534 adds built-in TOML support for the j2 filters.

### DIFF
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -170,7 +170,7 @@ To reference `secret_foo`inside this file, you can specify it in the inventory a
 
 It may be useful in some occasions when revealing references to have the values for the reference dynamically
 come from the environment in which Kapitan is executing. This backend provides such functionality. It will attempt
-to locate a value for a reference from the environment using a prefixed variable *`$KAPITAN_VARS_*`* convention
+to locate a value for a reference from the environment using a prefixed variable *`$KAPITAN_VAR_*`* convention
 and use this value with the refs command.
 
 ```shell
@@ -191,12 +191,9 @@ parameters:
 When using the above parameters reference, values would be consulted in the environment from the following
 variables:
 
-* `$KAPITAN_ENV_VARS_mysql_secret_foo`
-* `$KAPITAN_ENV_VARS_mysql_secret_bar`
+* `$KAPITAN_VAR_mysql_secret_foo`
+* `$KAPITAN_VAR_mysql_secret_bar`
 
-*Note*: Automatic creation of these references has not been tested. Please ensure the refs file is created before attempting to use it in your parameters structure.
-
-*Note*: Only flat string variables have been tested to work with this reference backend. Please do not attempt to use any other data structures with these refs.
 
 ### Vaultkv Secret Backend (Read Only) - Addons
 
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -1,15 +1,16 @@
-# Kapitan secrets
+# Kapitan References (Formerly Secrets)
 
-Kapitan can manage secrets with the following key management services:
+Kapitan can manage references and secrets with the following key management services:
 
 - GPG
 - Google Cloud KMS (beta)
 - AWS KMS (beta)
+- Environment
 - Vaultkv (read only support)
 
 If you want to get started with secrets but don't have a GPG or KMS setup, you can also use the secret `ref` type. Note that `ref` is not encrypted and is intended for development purposes only. *Do not use ref secrets if you're storing sensitive information!*
 
-## Using secrets
+## Using Secrets
 
 The usual flow of creating and using an encrypted secret with kapitan is:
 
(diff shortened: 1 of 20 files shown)


farvour:
Filled out the CLA.

kapitan/inputs/jinja2_filters.py
```diff
@@ -95,6 +96,10 @@ def to_yaml(obj):
     return yaml.safe_dump(obj, default_flow_style=False)
 
 
+def to_toml(obj):
```
uberspot, line 99: You'll need to add this to the list in function load_jinja2_filters(env) for this to work. Also would be good to update docs to include this too and add a small test for it. :)

farvour: Will do! I am actually working on the tests for the env reftype right now. I will also update all the docs so that folks aren't left in the dark.

adrianchifor:
Thanks for the PR, but this seems like 2-3 things combined in one. Would you mind splitting toml support and env ref backend into 2 PRs?

It will make them easier to review and in case any of them breaks, we don't have to revert all.

farvour:
I will split these up into two separate PRs.

farvour:
I updated the documentation a bit. Changed secrets -> references where applicable. I also added some docs on how to use the new env reftype backend.

docs/secrets.md
```diff
@@ -163,6 +166,38 @@ To reference `secret_foo`inside this file, you can specify it in the inventory a
...
+    secret_bar: ?{env:my/mysql/mysql_secret_bar}
+```
+
+When using the above parameters reference, values would be consulted in the environment from the following
+variables:
+
+* `$KAPITAN_ENV_VARS_mysql_secret_foo`
+* `$KAPITAN_ENV_VARS_mysql_secret_bar`
+
+*Note*: Automatic creation of these references has not been tested. Please ensure the refs file is created before attempting to use it in your parameters structure.
```
uberspot, line 197: This point is slightly more crucial and worth testing. Otherwise it might be counterintuitive for users. I can help test this tomorrow too.

farvour: Okay. I figure it should just work, since I'm just extending off the PlainRef (and I presume that one works for auto-create?) or maybe not.

uberspot: Yup, it should work same as the other backends since autocreation is part of the base ref class. But I can briefly test tomorrow for completeness, no prob. :)

kapitan/refs/env.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        super().__init__(data, kwargs=kwargs)
+        self.type_name = "env"
+
+    def reveal(self):
+        """
+        Attempt to locate the variable in the environment w/ the suffix.
+        """
+        return os.getenv(
+            "{}{}".format(DEFAULT_ENV_REF_VAR_PREFIX, self.path.split("/")[-1]), default=self.data,
```
farvour, line 31: FWIW this is a very opinionated way to derive the prefix name. I'm open to other suggestions if this isn't very clear. I had also considered converting the path separarator characters to _, but in our our case we use  target_name prefix for a lot of these. I thought that might cause the variable names to get quite lengthly and cause other problems, so I settled on the "last octet" of the path.

docs/secrets.md
uberspot: I tested both usecases. In detail:
1) No ref created. The ref was `secret_foo: ?{env:my/mysql/mysql_secret_foo||randomstr}`. I ran compile. The secret is auto-generated correctly. Once the secret is revealed if the env var is defined, it displays that as intended.
2) I also tested the use case of the ref already existing and the ref not existing and not having `||randomstr` and both worked as expected. :) One minor issue in the doc.
You mention KAPITAN_ENV_VARS_* but inn code you have `KAPITAN_VAR_*` so let's just make that the same everywhere :) 
The rest LGTM and we can merge after that.

kapitan/refs/env.py
uberspot: I'm personally ok with it, it resembles the TF_VAR convention that terraform uses so it makes sense.

docs/secrets.md
ramaro: So can we remove this from the docs then?

docs/secrets.md
```diff
@@ -163,6 +166,38 @@ To reference `secret_foo`inside this file, you can specify it in the inventory a
...
+
+*Note*: Only flat string variables have been tested to work with this reference backend. Please do not attempt to use any other data structures with these refs.
```
ramaro, line 199: and this

uberspot: Done

PR #534, opened 2020-06-22
Add TOML Jinja2 Filters

farvour:
Fixes issue #

## Proposed Changes

  - Bake in the jinja2 TOML filter support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ tags
 .coverage
 *.so
 *.h
+.vscode/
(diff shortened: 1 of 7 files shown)


uberspot:
Nice :) this could use a small update in docs too for the existence of a new filter and a small single test to cover it. In test_jinja2.py most likely?

farvour:
Sure I'll throw a test in here and update the doc.

farvour:
Added test and documentation. :)

PR #544, opened 2020-06-26
Test resources to fetch inventory items over http[s]

AlphaRoy14:
Continuation of issue #531


AlphaRoy14:
Hi @uberspot, I was trying to figure out if there's a way to host zipped inventory file in the test directory itself but I could not find a way to do so. We can have inventories hosted on different accounts and can then fetch it e.g https://github.com/AlphaRoy14/kapitan_tests/archive/master.zip. But this will make the tests dependent on the availability of external URIs. I was just wondering what would be the best way to approach this. In the PR I'm just fetching a single inventory item and not a zipped file.

uberspot:
> Hi @uberspot, I was trying to figure out if there's a way to host zipped inventory file in the test directory itself but I could not find a way to do so. We can have inventories hosted on different accounts and can then fetch it e.g https://github.com/AlphaRoy14/kapitan_tests/archive/master.zip. But this will make the tests dependent on the availability of external URIs. I was just wondering what would be the best way to approach this. In the PR I'm just fetching a single inventory item and not a zipped file.

Hi @AlphaRoy14, how come hosting a zipped inventory file in these test directories didn't work? 
Ideally if we can host a test zip under these same directories it would be preferred to have everything under the same dir. 
You could even make a quick package script that would create that zip for you and add it in PRs so it stays updated. 
Alternatively if this is definitely not possible, I can make a new repo for test resources under https://github.com/kapicorp which is also used by kapitan. :) Let me know how the first step goes.

AlphaRoy14:
Hi @uberspot 

> > Hi @uberspot, I was trying to figure out if there's a way to host zipped inventory file in the test directory itself but I could not find a way to do so. We can have inventories hosted on different accounts and can then fetch it e.g https://github.com/AlphaRoy14/kapitan_tests/archive/master.zip. But this will make the tests dependent on the availability of external URIs. I was just wondering what would be the best way to approach this. In the PR I'm just fetching a single inventory item and not a zipped file.
> 
> Hi @AlphaRoy14, how come hosting a zipped inventory file in these test directories didn't work?
> Ideally if we can host a test zip under these same directories it would be preferred to have everything under the same dir.
> You could even make a quick package script that would create that zip for you and add it in PRs so it stays updated.
> Alternatively if this is definitely not possible, I can make a new repo for test resources under https://github.com/kapicorp which is also used by kapitan. :) Let me know how the first option goes.

Thanks for the reply @uberspot, I was having a local issue with zipping the inventories and uploading it to github. Anyway, I have resolved it and added all the zipped inventories in the `tests/test_remote_inventory/zipped_inventories/` directory. I have locally tested fetching and unpacking and once this PR is merged, I'll make a PR with all the unit tests to check if any issue arises with Travis.

uberspot:
Sounds good to me :) Could you just re-push a no-change commit to re-trigger a build on travis? 
I can't unfortunately as this is your own github branch/fork.

AlphaRoy14:
> Sounds good to me :) Could you just re-push a no-change commit to re-trigger a build on travis?
> I can't unfortunately as this is your own github branch/fork.

For the test cases for [Remote Inventory](https://github.com/deepmind/kapitan/pull/521) to work, this PR needs to be merged first. As the tests involve fetching the zipped inventories that I have added in this PR.

Issue #500, opened 2020-04-30
Ability to show a secret from a ref file


pvanderlinden:
It would be great have the ability to show a secret from a ref file.

As a workaround I currently use:

```
#!/usr/bin/env sh
set -uex
REFNAME=$(echo $1|sed s+^refs/++)
echo "?{gkms:$REFNAME}" | kapitan refs --reveal -f -|base64 -d
```

Example:
`./show-secret.sh refs/targets/production/db/superuser_password`

PR #548, opened 2020-07-02
Fix issue #500: Add --ref-file or --tag to refs --reveal option.

uberspot:


kapitan/refs/base.py
```diff
@@ -506,6 +503,24 @@ def ref_from_embedded(self, type_name, b64_path):
...
+            type_name = ref_file_obj.pop("type")
+            encoding = ref_file_obj["encoding"]
+            data = ref_file_obj.pop("data")
+            from_base64 = encoding == "base64"
+
+            backend = self._get_backend(type_name)
+
+            # create new ref with deserialised data and remaining keys as kwargs
+            # note that encrypt=False is only for secret ref types, non secret refs (e.g. base64) will ignore
+            # from_base64 is True because data is always base64 encoded in embedded form
```
ramaro, line 520: Change comment to "from_base64 is True because data is being loaded from a ref file where it is always base64"

kapitan/refs/base.py
```diff
@@ -506,6 +503,24 @@ def ref_from_embedded(self, type_name, b64_path):
...
+            ref = backend.ref_type(data, encrypt=True, from_base64=from_base64, **ref_file_obj)
```
ramaro, line 521: from_base64 should always be True here

tests/test_cli.py
```diff
@@ -42,6 +42,47 @@ def setUp(self):
         subprocess.run(["gpg", "--import-ownertrust", example_key_ownertrust])
         os.remove(example_key_ownertrust)
 
+    def test_cli_secret_reveal_tag(self):
```
ramaro, line 45: Maybe add another test withtout `--base64`?

tests/test_cli.py
```diff
@@ -42,6 +42,47 @@ def setUp(self):
...
+        """
+        run $ kapitan refs - -write gpg: test_secretb64 - -base64
```
ramaro, line 47: fix command line: `--` instead of `- -`

Issue #527, opened 2020-06-15
Support overlay / multiple render stages


chris-sanders:
**Describe the bug/feature**
The use of an overlay (Kustomize / ytt as examples) is a pattern I'm finding necessary when working with 3rd party manifests such as those generated via Helm. Using a Helm repository as a starting and using the Helm input type renders output directly to the compiled directory, which seems to be the norm for compile steps. What I'd like is to have the ability to compile to an interim folder or apply multiple compile steps to allow overlay rendering and file organization before landing in the final location.

**Additional context**
I'll give an example to show why I find this useful.

The cluster I'm working on has more than one storage provider available. Most of the time I want to use the default provider but for some mount points I want to use an alternate storage class. This is, generally speaking, not something that I'm finding typically exposed in Helm templates. I would prefer to stay as close to upstream as possible, and I have no desire to write the entire manifest specifically for myself. I want to use the Helm chart, customize its' deployment with site specific variables.yaml that I generate and maintain with Kapitan and then when necessary adjust specific volumes storage providers. I'd like to be able to manage the overlay file and the generated output with Kapitan so I can maintain a complete 'compiled' folder, I don't want to apply overlays at deploy time.

Specifically I'm interested in using ytt (#354) but I think this extra compile step is beyond just supporting overlay input types due to the fact that the compile output (like helm) automatically target a compiled folder and it's not actually my final output.

Issue #365, opened 2019-09-08
Allow metadata to be added to compile targets


taharah:
**Describe the feature**

Allow metadata to be added to each compile target. For instance, setting a "name" that could be used to get the top level key for the inventory data used in the component(s). It would allow components to be written in a way that is much more usable than what is possible currently.

Current state:

```yaml
parameters:
  kapitan:
    compile:
    - output_path: <output_path_in_target_dir>
      input_type: jinja | jsonnet | kadet | helm
      input_paths:
        - path/to/input/dir/or/file
      output_type: <output_type_specific_to_input_type>
```

Adding a block similar to the following:

```yaml
parameters:
  kapitan:
    compile:
    - output_path: <output_path_in_target_dir>
      input_type: jinja | jsonnet | kadet | helm
      input_paths:
        - path/to/input/dir/or/file
      output_type: <output_type_specific_to_input_type>
      metadata:
        name: my_special_name
        version: v1.1.0
```

The metadata could then be referenced in each component:

```
local kap = import "lib/kapitan.libjsonnet";
local inventory = kap.inventory();
local metadata = kap.metadata();

{
    "data_java_opts": inventory.parameters[metadata.name].roles.data.java_opts,
}
```

uberspot:
How does this approach compare with a basic one like:
```
parameters:
  metadata:
    name: my_special_name
    version: v1.1.0
```
and then referencing it as:
`    "data_java_opts": inventory.parameters[parameters.metadata.name].roles.data.java_opts,` ? 

The usual usage patterns we've seen with kapitan most of the time define a "common.yml" class which all targets inherit and it makes it easy to globally distribute any type of generic configuration, but still gives you the flexibility of overwriting anything you want in specific targets. I'm not sure it that would be possible with the above approach.

taharah:
The main difference would that this would allow the metadata to be set per compile input rather than for the entire target. This would allow someone to reference the same component input(s), but end up compiling entirely different outputs by being able to flex the inventory data referenced via the metadata.

uberspot:
This is basically polymorphism but with the inventory. You can also do that per component by just inheriting a common class in all your component yml's in the inventory and overwrite it only when necessary in some targets.
We do something similar in our infrastructure to change args in k8s deployments per target, but also have a set of "base" global args that we ship to all targets.

I'm not against the metadata keyword either. I'm just clarifying because it seems that Kapitan would become slightly opinionated to make polymorphism easier (usability-wise). 

[...]

PR #560, opened 2020-07-12
Adds kadet_params for compile time parameter injection into kadet modules

sebradloff:
Adds kadet_params as a way to create compile time parameters for kadet modules. This should enable a reusable plugin like system within a single target, that shows the contract cleary through injection rather than relying on the 'parameters' concept in inventory.

The compile process currently works in a blocking sequential manner. Only targets are multi threaded. This enables a the concept of "post processing" or "overlays" as mentioned in #527 and #365.

## Proposed Changes

  - 'kadet_params' as attribute for kadet input type

### DIFF
--- a/tests/test_resources/kadet_functions/add_team_label/__init__.py
+++ b/tests/test_resources/kadet_functions/add_team_label/__init__.py
@@ -18,12 +18,14 @@ def set_namespace(obj, namespace):
     obj.root.metadata["namespace"] = namespace
     return obj
 
+
 def get_root_path(path, target_name):
     drive, tail = os.path.split(path)
     if tail == target_name:
         return drive
     else:
         return get_root_path(drive, target_name)
+
 
 def main(kadet_params):
     team_name = kadet_params.get("team_name", "no-owner")
--- a/tests/test_resources/kadet_functions/add_team_label/__init__.py
+++ b/tests/test_resources/kadet_functions/add_team_label/__init__.py
@@ -1,0 +1,68 @@
+import os
+import json
+from kapitan.resources import jinja2_render_file, yaml_load
+from kapitan.inputs import kadet
+import glob
+import yaml
+import time
+
+inventory = kadet.inventory()
+
+def set_team_name(obj, team_name):
+    obj.root.metadata["labels"]["team_name"] = team_name
+    return obj
+
+def set_namespace(obj, namespace):
+    obj.root.metadata["namespace"] = namespace
+    return obj
+
+def main(kadet_params):
+    team_name = kadet_params.get("team_name", "no-owner")
+    if "namespace" in kadet_params:
+        namespace = kadet_params.get("namespace")
+    else:
+        raise ValueError("'namespace' key not found in 'kadet_params'")
+
+    if "input_paths" in kadet_params:
+        input_paths = kadet_params.get("input_paths")
+    else:
+        raise ValueError("'input_paths' key not found in 'kadet_params'")
+    
+    # get path where files have been compiled on this run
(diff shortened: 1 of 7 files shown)


tests/test_resources/kadet_functions/add_team_label/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+    if "input_paths" in kadet_params:
+        input_paths = kadet_params.get("input_paths")
+    else:
+        raise ValueError("'input_paths' key not found in 'kadet_params'")
+
+    # get path where files have been compiled on this run
+    target_name = inventory.parameters.kapitan.vars.target
+    compile_path = kadet_params.get("compile_path")
+    root_path = ""
+    for s in compile_path.split("/"):
```
uberspot, line 38: Don't hardcode path delimiters. This might not work the same on all operating systems in the future. 
https://docs.python.org/2/library/os.path.html#os.path.split use os.path.split() or smth similar to get all path components. 
Maybe os.path.basename (?).

tests/test_resources/kadet_functions/add_team_label/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+        else:
+            root_path += "/" + s
+
+    target_compiled_dir = root_path + "/" + target_name
+
+    all_objects = {}
+    for ip in input_paths:
+        output = kadet.BaseObj()
+
+        ip_file_path = os.path.abspath(target_compiled_dir + "/" + ip)
```
uberspot, line 52: Same here, use os.path.join for joining all paths.

uberspot:
Hey @sebradloff thanks for the PR :) 
Could you also add a couple sentences in the docs on how/why people would use this? Maybe in docs/compile.md(?).

ramaro:
Hey @sebradloff thanks for the contribution! Do I understand correctly that this is somewhat related to https://github.com/deepmind/kapitan/issues/527?

ramaro:
> This should enable a reusable plugin like system within a single target, that shows the contract clearly through injection rather than relying on the 'parameters' concept in inventory.

Hey @sebradloff thanks for that. I'm not sure I understand this completely. I'm a bit confused because `kadet_params` is set in the inventory and missing the point. Can you elaborate with an example?

sebradloff:
@ramaro yea I elaborated a bit in the docs that @uberspot requested. The idea is that sometimes there is a need for "post processing" or "layering" that can't be controlled in just one compile block. For instance, using the helm input type, I would like template manifests for a helm chart but maybe there's no parameter for something need to change or add to kubernetes resources. Instead of modifying the helm chart and now needing to be responsible for a custom helm chart, we can write a kadet module that takes parameters `kadet_params` and make the necessary changes there. 

A common issue I've run into, is that some helm charts do not include the namespace for which you would like to deploy resources during templating and leave that up to deploy time. With that in mind, I would like to iterate over those manifests and add the namespace. I could do this without needing `kadet_params` but I would need to 1) setup an implicit contract with parameters that kadet reads in and 2) can only run the module once and would need to ensure it is the last compile block to run in order to transform all the necessary manifests. With `kadet_params` I can clearly specify the module to run and inject arguments for the post processing. This makes the module usage clear and the contract becomes explicit rather than implicit.

sebradloff:
We could also use the example in the kadet proposal. Imagine that you wanted to reuse that module to create the same kubernetes objects for multiple classes or components in the same target. The only way to specify that right now, would be to modify the python code of the kadet module. With `kadet_params` you could inject the parameters instead of having to pull them from the kadet inventory.

PR #562, opened 2020-07-14
Allow reclass override none feature

srueg:
Enable the reclass `allow_none_override` feature [1] to be able to
remove dicts and lists further down in the hierarchy.

[1] https://github.com/salt-formulas/reclass/blob/develop/README-extensions.rst#allow-override-list-and-dicts-by-empty-entitynone-instead-of-merge

## Proposed Changes

Set `allow_none_override=true` in reclass config.

### DIFF
--- a/kapitan/resources.py
+++ b/kapitan/resources.py
@@ -307,6 +307,7 @@ def inventory_reclass(inventory_path):
             "nodes_uri": os.path.join(inventory_path, "targets"),
             "classes_uri": os.path.join(inventory_path, "classes"),
             "compose_node_name": False,
+            "allow_none_override": True,
         }
 
         try:


uberspot:
This would change the behavior for inventory merging in Kapitan by default which would break kapitan for users. 
Could you make this configurable with a flag? Something like `--inventory-allow-none-override` ? 
And set the default to false as it is now?

ramaro:
I don't even think that's necessary. Users can have their own reclass config file inside inventory/ and enable this, right?

srueg:
> This might change the behavior for inventory merging in Kapitan by default which could break kapitan for users?

Since it's currently not allowed to merge `None` over dicts/lists it wouldn't break any setup (since they'd be broken anyway), right?


> I don't even think that's necessary. Users can have their own reclass config file inside inventory/ and enable this, right?

Good point. I'd then propose to merge the config found in the file together with the defaults set by Kapitan?

I was just really confused by the [reclass docs](http://reclass.pantsfullofunix.net/operations.html#data-merging) which suggest that this is enabled by default anyway, hence this PR.
> While the scalar could be appended to an existing list, there is no sane default assumption in the context of a dictionary, so this behaviour seems the most logical. Plus, it allows for a dictionary to be erased by overwriting it with the null value.

uberspot:
So if it doesn't break any use case, I'm fine with adding it as a default option to match reclass docs.
Could you also rebase your branch so it can be merged?

ramaro:
lgtm

Issue #437, opened 2019-12-18
Modularise Kapitan


uberspot:
Currently kapitan is packaged in pypi (or as a binary) along with all its dependencies.
Adding more input types or secret backends means adding extra dependencies which might not be useful for certain users. 
This issue is to track modularisation of kapitan to a set of core dependencies (cryptography,pyyaml,jsonschema etc) and then extra modules (e.g. boto3,google-api-python-client) that can be loaded by users that actually use those specific features.

AlphaRoy14:
Hi! Just looking for a bit of clarification. Along with modularization into main module and extra modules that is packages in pypi, are we also interested in something like lazy loading of dependencies?

mrsupiri:
In my opinion, it's better to ship these different kinds of secret management systems as standalone plugins, So if someone has their own way of storing secrets they can just write a custom plugin and load it without interfering with Kapitan's core functionality.

adrianchifor:
The idea here is simple:
```
pip3 install --user kapitan
```
Will just install core kapitan functionality (reclass inventory, jinja2/jsonnet/kadet templates, only ref secrets).
```
 pip3 install --user kapitan[gkms]
```
Core functionality + `google-api-python-client` via [pypi extras](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies), in case you need to use secrets backed by Google KMS.

And so on for the other secret backends (gpg, awskms).

PR #569, opened 2020-07-20
Add proposal to modularize kapitan

AlphaRoy14:
Utilize pip extras to specify dependencies used with
specific modules such as secret backends.

Related to issue #437


docs/kap_proposals/kap_8_modularize_kapitan.md
```diff
@@ -0,0 +1,15 @@
...
+```sh
+$ pip3 install --user kapitan # to install only core dependencies
+$ Pip3 install --user kapitan[gkms] ​# gkms is the module
+```
+
+## Implementation
+
+- The main module includes the essential kapitan dependencies and reclass dependencies, which will be included in the ​requirement.txt​ file.
+- The extra module [pypi extras](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) will be defined in the `s​etup.py`​ file.
+- The extra dependencies are of the secret backend like (AWS Key backend, Google KMS Key backend, Vault Key backend etc.).
```
uberspot, line 15: besides these it might make sense to also add helm in the use cases where people might want it as a module.

Issue #572, opened 2020-07-21
Bug: Using remote inventory files (still) requires a separate fetch command due to reclass failing on missing classes.


skinny:
**this is just a quick write up and I'll try to expand on this tomorrow.**

Using the latest 0.29-rc version :

I would really like to see that running a `compile --fetch` once would download the remote files and be able to use them afterwards in the same run as an inventory class.

Downloading the inventory file works fine but referencing the just downloaded file in other places like :

```
classes:
  - stages.dev
  - clusters.k8s-euw-039
  - tentants.customer-x
```
breaks the compile process because the file 'customer-x' is not present (yet). After running the `inventory` command the file exists and the compilation succeeds.

Another idea that I had was that it would be great to also be able to fetch remote 'targets' where the target file is generated/served by an external system instead.

uberspot:
>> Another idea that I had was that it would be great to also be able to fetch remote 'targets' where the target file is generated/served by an external system instead.
This is definitely interesting, but might be better suited in a separate issue. :) 

The rest does seem like a valid bug where fetched inventories aren't loaded yet into kapitan (?). cc @AlphaRoy14 

@skinny could you also add a snippet with your kapitan config for the remote inventories if they're public urls? To help debug.

AlphaRoy14:
If the configuration is something like:
```yaml
classes:
  - stages.dev
  - clusters.k8s-euw-039
  - tentants.customer-x # not currently in the classes directory but will be created if inventory fetch is successful 
parameters:
    kapitan:
        inventory:
            - type: git
               output_path: xyz
               source: xyz
```
Then kapitan would throw an error since we are using reclass in the load_target_inventory() function. All the classes must be present for reclass to work without any error.
https://github.com/deepmind/kapitan/blob/bbe0979391a23a311c506209ca8d899d8864c18b/kapitan/targets.py#L100
https://github.com/deepmind/kapitan/blob/bbe0979391a23a311c506209ca8d899d8864c18b/kapitan/targets.py#L334

It does make sense to have classes declared before they are fetched. So for that, we can do something like:
- Define a function in `/kapitan/kapitan/resources.py` similar to inventory_reclass() https://github.com/deepmind/kapitan/blob/bbe0979391a23a311c506209ca8d899d8864c18b/kapitan/resources.py#L294
[...]

PR #574, opened 2020-07-23
Fix issue #572 Enable adding classes before they are available

AlphaRoy14:
Enable adding classes that are are not yet available
but are fetched at a later time.

Fixes issue #572


kapitan/targets.py
```diff
@@ -97,10 +103,14 @@ def compile_targets(
...
-                target_objs = load_target_inventory(inventory_path, updated_targets)
+                target_objs = load_target_inventory(
+                    inventory_path, updated_targets, ignore_class_notfound=True
+                )
                 cached.inv_sources.update(new_sources)
                 new_sources = list(set(list_sources(target_objs)) - cached.inv_sources)
-
+            # reset inventory cache and load target objs to check for missin classes
+            cached.reset_inv()
+            target_objs = load_target_inventory(inventory_path, updated_targets, ignore_class_notfound=True)
```
uberspot, line 113: I believer this line should be `ignore_class_notfound=False` ?
also typo in comment above `missin classes`

AlphaRoy14: Yes, that should be False. A few more issues have come up during testing and I'm trying to resolve it asap. Will make push the code and ping you as soon as it is ready for review :)

AlphaRoy14:
Hi @uberspot, I'm facing a few issues while trying to fix this.

1. Unit tests work fine but when the tests are run together, it shows the `http2: Transport received unsolicited DATA frame; closing connection` message and stalls the execution indefinitely 
I suspect it is caused by Go trying to fetch the helm dependencies. 

2. When using the `--targets` flag to specify targets, Kapitan still throws an error if any target exists in the target directory which does not have its classes present locally. This is because of the last function call to load_target_inventory() with ignore_class_notfound=False `target_objs = load_target_inventory(inventory_path, updated_targets, ignore_class_notfound=False)`. 
Since the entire inventory is mapped by reclass before looping through selected targets, I could not think of a solution without changing reclass. 

3. Could not find a way to mute reclass from printing warning about class not found 
https://github.com/kapicorp/reclass/blob/fb50054d6332167c34dd2ae8b486943dd9c28746/reclass/core.py#L140

uberspot:
> Hi @uberspot, I'm facing a few issues while trying to fix this.
> 
> 1. Unit tests work fine but when the tests are run together, it shows the `http2: Transport received unsolicited DATA frame; closing connection` message and stalls the execution indefinitely
>    I suspect it is caused by Go trying to fetch the helm dependencies.
> 2. When using the `--targets` flag to specify targets, Kapitan still throws an error if any target exists in the target directory which does not have its classes present locally. This is because of the last function call to load_target_inventory() with ignore_class_notfound=False `target_objs = load_target_inventory(inventory_path, updated_targets, ignore_class_notfound=False)`.
>    Since the entire inventory is mapped by reclass before looping through selected targets, I could not think of a solution without changing reclass.
> 3. Could not find a way to mute reclass from printing warning about class not found
>    https://github.com/kapicorp/reclass/blob/fb50054d6332167c34dd2ae8b486943dd9c28746/reclass/core.py#L140

1. That does happen when it fetches the http helm dependencies and it seems it can't find the chart that's being asked in code? 
I guess this is happening for prometheus 11.3.0 as defined in monitoring.yml. I think this is where we get those charts from https://github.com/helm/charts/blob/master/stable/prometheus/Chart.yaml Maybe it just needs a version bump?

2. Shouldn't it have downloaded all missing inventories in the first two calls to load_target_inventory (in the while loop that also fetches the remotes?). My assumption was that by the third call to load_target_inventory it would have already fetched all missing inventory dependencies. Are you sure they're downloaded to the correct output folder? Does this happen on local `python -m unittest -v run  tests.test_remote_inventory.RemoteInventoryTest` runs too? 

3. A warning is not a huge issue at this point. Let's get it working first. :) We could send a commit later on upstream to reclass to make the warning output configurable.

AlphaRoy14:
> > Hi @uberspot, I'm facing a few issues while trying to fix this.
> > 
> > 1. Unit tests work fine but when the tests are run together, it shows the `http2: Transport received unsolicited DATA frame; closing connection` message and stalls the execution indefinitely
> >    I suspect it is caused by Go trying to fetch the helm dependencies.
> > 2. When using the `--targets` flag to specify targets, Kapitan still throws an error if any target exists in the target directory which does not have its classes present locally. This is because of the last function call to load_target_inventory() with ignore_class_notfound=False `target_objs = load_target_inventory(inventory_path, updated_targets, ignore_class_notfound=False)`.
> >    Since the entire inventory is mapped by reclass before looping through selected targets, I could not think of a solution without changing reclass.
> > 3. Could not find a way to mute reclass from printing warning about class not found
> >    https://github.com/kapicorp/reclass/blob/fb50054d6332167c34dd2ae8b486943dd9c28746/reclass/core.py#L140
> 
> 1. That does happen when it fetches the http helm dependencies and it seems it can't find the chart that's being asked in code?
>    I guess this is happening for prometheus 11.3.0 as defined in monitoring.yml. I think this is where we get those charts from https://github.com/helm/charts/blob/master/stable/prometheus/Chart.yaml Maybe it just needs a version bump?
> 2. Shouldn't it have downloaded all missing inventories in the first two calls to load_target_inventory (in the while loop that also fetches the remotes?). My assumption was that by the third call to load_target_inventory it would have already fetched all missing inventory dependencies. Are you sure they're downloaded to the correct output folder? Does this happen on local `python -m unittest -v run  tests.test_remote_inventory.RemoteInventoryTest` runs too?
> 3. A warning is not a huge issue at this point. Let's get it working first. :) We could send a commit later on upstream to reclass to make the warning output configurable.

2. The issue arises if suppose there are three targets (`a`, ` b`,`  c`) and we specify selected targets with ` $ kapitan compile --fetch -t a b`. `load_target_inventory` function maps the entire inventory so all three targets are checked for missing classes. This is fine as long as `ignore_class_notfound=True`,  but in the last call we set `ignore_class_notfound=False` which raises the error of class not found for `c`. 

For this reason, I made a separate environment with just one target which fetches its class, the unit test `python -m unittest -v run tests.test_remote_inventory.RemoteInventoryTest` worked fine.

tests/test_remote_inventory/environment_three/inventory/targets/nginx.yml
```diff
@@ -0,0 +1,31 @@
...
+  kapitan:
+    compile:
+      - output_path: manifests
+        input_type: jinja2
+        input_paths:
+         - templates/nginx.conf
+    inventory:
+      - type: https
+        output_path: classes/test-objects.yml
+        source: https://raw.githubusercontent.com/AlphaRoy14/kapitan/master/tests/test_resources/inventory/classes/test-objects.yml
```
uberspot, line 22: Can we point these source urls in this file to kapitan repo files? Otherwise any change in your repo in the future would break tests. :)

Issue #467, opened 2020-03-04
Add support for GCP Secret Manager 


ademariag:
**Describe the bug/feature**
Add support for GCP Secret Manager (https://cloud.google.com/secret-manager/)

While we support `gkms` when storing local secrets/references, it would be useful to also support GCP Secret Manager.

With Secret Manager (SM) Kapitan would be able to share secrets with scripts/processes that do not have access to the Kapitan repository.

Example use case: Allow a script in a developer repository to have access to the same secrets already setup/available in Kapitan (i.e. a shared API key)

PR #575, opened 2020-07-28
Proposal to add support for Google Secret Manager backend

AlphaRoy14:
With [Google Secret Manager](https://cloud.google.com/secret-manager/), Kapitan would be able to share secrets with scripts/processes
that do not have access to the Kapitan repository.

Related to issue #467

### DIFF
--- a/docs/kap_proposals/kap_8_google_secret_management.md
+++ b/docs/kap_proposals/kap_8_google_secret_management.md
@@ -1,6 +1,6 @@
-# Support for Google Secret Management
+# Support for [Google Secret Manager](https://cloud.google.com/secret-manager/)
 
-This feature will enable users to retrieve secrets from Google Secret Management API using the `gsm` keyword.
+This feature will enable users to retrieve secrets from Google Secret Manager API using the `gsm` keyword.
 
 ## Specification
 
--- a/docs/kap_proposals/kap_8_google_secret_management.md
+++ b/docs/kap_proposals/kap_8_google_secret_management.md
@@ -26,17 +26,18 @@ export PROJECT_ID=<Project_Id>
 ```
 
 ## Using a secret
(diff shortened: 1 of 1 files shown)


docs/kap_proposals/kap_8_google_secret_management.md
```diff
@@ -0,0 +1,83 @@
...
+        image: mysql:latest
+        users:
+            root:
+                password: ?{gsm:path/to/secret_id:version_id}
+```
+Here, `version_id` will be an optional argument. By default it will point to `latest`.
+
+## Revealing a secret
+
+After compilation, the secret reference will be prefixed with 8 characters from the sha256 hash of the retrieved password
```
uberspot, line 63: Postfixed instead of prefixed *

docs/kap_proposals/kap_8_google_secret_management.md
```diff
@@ -0,0 +1,83 @@
...
+type: Opaque
+```
+To reveal the secret, the following command will be used
+`$ kapitan ref --reveal -f compiled/file/containing/secret`
+
+## Dependencies
+
+* [google-cloud-secret-manager](https://github.com/googleapis/python-secret-manager)
+
+*note* Kapitan will not be responsible for authentication or access management
```
uberspot, line 83: to GCP*

docs/kap_proposals/kap_8_google_secret_management.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+- As an environment variable
+```shell
+export PROJECT_ID=<Project_Id>
+```
+
+## Using a secret
+
+`secret_id` is used to point to the secret data. The following command will be used to add a `secret_id` to kapitan.
```
ramaro, line 30: Can you add a small explanation of what secret_id is in the context of GSM? Just to make things clearer as this is a bit different from other secret backends. :)

Issue #583, opened 2020-08-06
Compile all targets with fetch option will copy over http dependencies that get cached to compiled directory


sebradloff:
**Describe the bug/feature**
As a part of my CI pipeline, I run `kapitan compile --fetch` to ensure that all upstream dependencies download and that they haven't changed. Right now, using the `http` external dependency manager, it will fetch the dependency and cache it in the tmp directory (working directory) created by kapitan during compile. [One of the last steps of a kapitan compile command](https://github.com/deepmind/kapitan/blob/master/kapitan/targets.py#L140), if no targets are specified, is to copy EVERYTHING from the working directory into the compiled directory. I think the solve here is to have a "cache" that is separate from the working directory for dependencies that are downloaded.

I believe this is the code in `fetch_http_source` that does the caching of the file to the tmp dir: https://github.com/deepmind/kapitan/blob/master/kapitan/dependency_manager/base.py#L213

This is the code calling the function feeding it a `cached_source_path`: https://github.com/deepmind/kapitan/blob/41345d34a6b9ee97641fb86088721e1b762a4182/kapitan/dependency_manager/base.py#L163

**To Reproduce**
inventory/targets/dev.yml
```inventory/targets/dev.yml
classes:
  - components.core-dns.v1

parameters:
  kapitan:
    vars:
      target: dev
```

inventory/targets/stg.yml
```inventory/targets/stg.yml
classes:
  - components.core-dns.v1

parameters:
  kapitan:
    vars:
      target: stg
```

inventory/classes/components/core-dns/v1.yml
```inventory/classes/components/core-dns/v1.yml
# documentation for CoreDNS come from EKS guides: https://docs.aws.amazon.com/eks/latest/userguide/coredns.html
parameters:
  kapitan:
    dependencies:
      - type: https
        output_path: components/core-dns/v1/v1.yml
        source: https://amazon-eks.s3.us-west-2.amazonaws.com/cloudformation/2020-07-23/dns.yaml
     compile:
        - input_type: copy
          output_path: core-dns/v1
          input_paths:
           - components/core-dns/v1
```

```
root@db0cd51a81f2:/src# tree compiled/
compiled/
├── 72eafc31dns.yaml
├── dev
│   └── core-dns
│       └── v1
│           └── v1.yaml
└── stg
    └── core-dns
        └── v1
            └── v1.yaml

6 directories, 3 files
```

logs from kapitan compile all targets with fetch option
```
[...]

PR #584, opened 2020-08-06
Add a dedicated directory in the temp path

AlphaRoy14:
Add a dedicated directory in the `temp_path` to
store the compiled items and finally copy the items
to the compile path.

This way, when `$ kapitan compile --fetch` is used, fetched dependencies won't be copied to the compile path.

Fixes issue #583

### DIFF
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -47,6 +47,7 @@ def compile_targets(
     """
     # temp_path will hold compiled items
     temp_path = tempfile.mkdtemp(suffix=".kapitan")
+    temp_compile_path = os.path.join(temp_path, "compiled")
     dep_cache_dir = temp_path
 
     updated_targets = targets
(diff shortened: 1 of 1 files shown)


AlphaRoy14:
As I understand the `temp_path` in `kapitan/targets.py` was initially designed to just temporally store compiled targets and then copy the entire contents of the `temp_path` to the `compiled_path` when no targets were explicitly specified.
https://github.com/deepmind/kapitan/blob/41345d34a6b9ee97641fb86088721e1b762a4182/kapitan/targets.py#L89
https://github.com/deepmind/kapitan/blob/41345d34a6b9ee97641fb86088721e1b762a4182/kapitan/targets.py#L139

Now the `temp_path` is also used to temporally store fetched dependencies and inventories before they are copied to their respective places. This causes the fetched dependency items to also be copied to the compile path i.e `./compiled` along with the compiled targets when  `shutil.copytree()` is used in https://github.com/deepmind/kapitan/blob/41345d34a6b9ee97641fb86088721e1b762a4182/kapitan/targets.py#L139

So, having a dedicated directory in the `temp_path` e.g `temp_path/compiled`  for storing the compiled targets temporarily before copying them to `./compiled` should prevent copying unwanted dependency items to `./compiled` (?)

sebradloff:
anything I can do to help get this approved?

PR #587, opened 2020-08-16
Create External Input

sebradloff:


sebradloff:
@AlphaRoy14 @yoshi-1224 @ramaro I think this might interest all of you, since we're all people who have taken a crack at the helm binding for Mac OSX. I'm curious to hear your thoughts on whether you think this would make sense high level. If it's hard to imagine, I might be able to pull off a quick and dirty spike next week to demonstrate proof of concept.

AlphaRoy14:
Hey @sebradloff Nice idea. However, there are some advantages of using binding over using subprocess commands as mentioned in this [StackOverflow answer](https://stackoverflow.com/questions/16786183/writing-a-python-binding-vs-using-command-line-directly). Since the binding works fine for Linux, maybe we can try to implement the BYO binary idea on darwin for now?

ramaro:
I think this is a good idea. While we figure out the binding issue, we can have this implemented as long as it doesn't change the way the helm input type works. Also, this could be a fallback to the lack of installed binding or even as feature to allow running different versions of helm without having the change kapitan itself.

ramaro:
@sebradloff  and as you say, maybe it even makes more sense to invest on a  "bring your own binary" input type!

sebradloff:
@AlphaRoy14 that makes sense. It depends on how tightly we want to control the interface. I’m starting to think we want it more open, given we want kapitan to be a general solution. I messed around with the Binary input type last night. I’ll clean it up a bit and push it up hopefully tonight. It’s def more of a POC but I think it’ll be interesting to get your takes on it.

sebradloff:
I just pushed up a draft of what a "Bring Your Own Binary" input type would look like. A target file is included: tests/test_resources/inventory/targets/binary-test.yml

What I focused on in the interface is using `input_paths` as a way to specify where the binary is you wish to execute, then supplemented additional args `binary_env_vars` and `binary_args` that you can inject to use with your binary. The last big piece is making `\${compiled_target_dir}` this specifically something that will be transpiled before the binary command is executed. This allows users to manipulate files that already exist in the compiled target directory or create files and direct them to the compiled target directory. Essentially as long as the binary returns with a non zero exit code, it is valid. This keeps the interface wide. All of this is achievable today through kadet but forces users to write python. This lowers the barrier to entry for making decisions but also keeps in line with the spirit of kapitan, which is to have a consistent, reliable, and general way to generate file changes.

Excited to know what y'all think about this interface.

kapitan/inputs/base.py
```diff
@@ -42,6 +42,8 @@ def compile_obj(self, comp_obj, ext_vars, **kwargs):
         for input_path in comp_obj["input_paths"]:
             globbed_paths = [glob.glob(os.path.join(path, input_path)) for path in self.search_paths]
             inputs = list(itertools.chain.from_iterable(globbed_paths))
+            # remove duplicate inputs
+            inputs = list(dict.fromkeys(inputs))
```
ramaro, line 46: I think it's better to use set() instead

kapitan/inputs/binary.py
```diff
@@ -0,0 +1,76 @@
...
+import os
+import shutil
+from distutils.dir_util import copy_tree
+
+from kapitan.inputs.base import InputType
+
+logger = logging.getLogger(__name__)
+
+
+class Binary(InputType):
```
ramaro, line 20: I know this is a bit late. But what do you think of calling this "External" instead? I think it's more accurate than "binary", since it runs an external executable, which could be a script or a binary.

kapitan/inputs/binary.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self, compile_path, search_paths, ref_controller):
+        super().__init__("binary", compile_path, search_paths, ref_controller)
+        self.binary_env_vars = {}
+        self.binary_args = []
+        self.target_name = None
+
+    def set_binary_env_vars(self, binary_env_vars):
```
ramaro, line 27: `set_env_vars()` seems clearer to me. Same for `binary_env_vars`, `env_vars` is cleaner?

kapitan/targets.py
```diff
@@ -414,14 +415,16 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, inven
...
 
-    jinja2_compiler = Jinja2(compile_path, search_paths, ref_controller, inventory_path)
-    jsonnet_compiler = Jsonnet(compile_path, search_paths, ref_controller)
-    kadet_compiler = Kadet(compile_path, search_paths, ref_controller)
-    helm_compiler = Helm(compile_path, search_paths, ref_controller)
-    copy_compiler = Copy(compile_path, search_paths, ref_controller)
-    remove_compiler = Remove(compile_path, search_paths, ref_controller)
-
     for comp_obj in compile_objs:
+        # these need to be inside this for loop so default values are reinitialized for each compile object
```
ramaro, line 419: OK. I feel that at some point, these should be instantiated on demand.

kapitan/targets.py
```diff
@@ -444,10 +447,14 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, inven
             input_compiler = copy_compiler
         elif input_type == "remove":
             input_compiler = remove_compiler
+        elif input_type == "binary":
+            input_compiler = binary_compiler
+            if "binary_args" in comp_obj:
```
ramaro, line 452: why just not `args`?

kapitan/targets.py
```diff
@@ -444,10 +447,14 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, inven
...
+                binary_compiler.set_binary_args(comp_obj["binary_args"])
+            if "binary_env_vars" in comp_obj:
```
ramaro, line 454: same, `env_vars` instead?

Dockerfile
```diff
@@ -17,6 +17,12 @@ COPY ./MANIFEST.in ./MANIFEST.in
 COPY ./requirements.txt ./requirements.txt
 COPY ./setup.py ./setup.py
 
+ARG YTT_VERSION="0.30.0"
```
ramaro, line 20: if this is just for the test, I don't think we need it permanently, do we?

ramaro (changes requested):
Great idea! Can we update the title? Left a few comments, mainly OCD nits... :D

Dockerfile
sebradloff: Correct. Just for the test. Def something we don’t need permanently.

kapitan/inputs/binary.py
```diff
@@ -0,0 +1,76 @@
...
+                    "Executing binary input with command '{}' and env vars '{}' failed: {}".format(
+                        args, self.binary_env_vars, binary_result.stderr
+                    )
+                )
+
+        except OSError as e:
+            logger.exception(f"Binary failed to run. Error: {e}")
+
+    def default_output_type(self):
+        # no output_type options for copy
```
uberspot, line 75: Is this comment supposed to be "for binary" instead of "for copy"?

sebradloff: yep. changing that

kapitan/targets.py
sebradloff: I was doing this to stick with the convention established in the helm input since no other input_type uses that arg but I have no problem changing it.

kapitan/targets.py
sebradloff: done in this PR

sebradloff:
I like the idea of adding a checksum but would like to make it a part of a new PR. I think I still need to add some documentation here for External and specify it's in alpha/experimental. @uberspot I'm having some issues running the linter locally `make format_codestyle` seems to produce different results than the github action

AlphaRoy14:
> I like the idea of adding a checksum but would like to make it a part of a new PR. I think I still need to add some documentation here for External and specify it's in alpha/experimental. @uberspot I'm having some issues running the linter locally `make format_codestyle` seems to produce different results than the github action

Upgrading `black` using `$ pip3 install --upgrade --user black` worked for me. You can try it out maybe? :)

ramaro (approved):

PR #589, opened 2020-08-16
Remove input type

sebradloff:
## Proposed Changes

  - Adds new input type `Remove`. This allows users to remove any compiled files.
  - Adds `temp_path` to `search_paths`. This allows users to reference files that have already been compiled for the `Remove` input to work but also opens up the ability to use the outputs of previous compile blocks to be the inputs for preceding compile blocks.

### DIFF
--- a/tests/test_remove.py
+++ b/tests/test_remove.py
@@ -64,7 +64,7 @@ class RemoveTest(unittest.TestCase):
     def test_remove_file_folder(self):
         test_dirs_bootstrap_helper()
         self.copy_compiler.compile_file(test_file_path, compile_path, None)
-        
+
         self.assertTrue(os.path.exists(test_file_compiled_path))
         self.remove_compiler.compile_file(test_file_compiled_path, compile_path, None)
         self.assertFalse(os.path.exists(test_file_compiled_path))
@@ -72,7 +72,7 @@ class RemoveTest(unittest.TestCase):
     def test_remove_folder_folder(self):
         test_dirs_bootstrap_helper()
         self.copy_compiler.compile_file(file_path, compile_path, None)
-        
+
         self.assertTrue(os.path.exists(compile_path))
         self.remove_compiler.compile_file(compile_path, compile_path, None)
         self.assertFalse(os.path.exists(compile_path))
--- a/tests/test_remove.py
+++ b/tests/test_remove.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+
+# Copyright 2019 The Kapitan Authors
+# SPDX-FileCopyrightText: 2020 The Kapitan Authors <kapitan-admins@googlegroups.com>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"remove tests"
+import filecmp
+import hashlib
+import os
+import shutil
+import sys
+import unittest
+
+from kapitan.cli import main
+from kapitan.inputs.copy import Copy
+from kapitan.inputs.remove import Remove
+from kapitan.utils import directory_hash
+
+search_path = ""
+ref_controller = ""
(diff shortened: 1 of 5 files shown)


tests/test_remove.py
```diff
@@ -0,0 +1,103 @@
...
+from kapitan.utils import directory_hash
+
+search_path = ""
+ref_controller = ""
+test_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "test_remove")
+compile_path = os.path.join(test_path, "output")
+file_path = os.path.join(test_path, "input")
+test_file_path = os.path.join(file_path, "test_copy_input")
+test_file_compiled_path = os.path.join(compile_path, "test_copy_input")
+test_file_content = """
```
uberspot, line 28: It might be slightly cleaner to keep this yaml as a normal file next to other test resources?  For easier maintenance/updates to it?

sebradloff: the file content doesn't actually matter here. It could easily just be one line. I lifted it from the `test_copy.py` file.

sebradloff:
@ramaro @uberspot I wanted to call out not just the input type I'm introducing but also introducing the `temp_path` to search paths: https://github.com/deepmind/kapitan/pull/589/files#diff-280d4028c0e294072ab70d24452f7b8bR52 This would allow all input blocks to use any files that exist in the `temp_path`, enabling the ability for users to use input blocks in a sequentially transformative manner. This I believe opens the door for "overlaying" as called out in this feature request: https://github.com/deepmind/kapitan/issues/527. While we can currently do that with kadet today, we can't do it with any of the other input types.

tests/test_remove.py
uberspot: Gotcha, no worries then. :)

sebradloff:
> Pretty cool work. Is my understanding correct that if you specify both copy and remove inputs the copy one will always run first, and then the remove one?

@uberspot it's less about the fact that it's a copy and remove inputs, it's that when reclass runs, the compile block for the target determines the order in which inputs run. These are run sequentially in kapitan right now.

For example, if your compile block looks like this:
```yaml
parameters:
  target_name: removal
  kapitan:
    vars:
      target: ${target_name}
    compile:
      - input_type: copy
        input_paths:
          - copy_target
        output_path: .
      # test removal of a file
      - input_type: remove
        input_paths:
          - compiled/${kapitan:vars:target}/copy_target
        output_path: .
```
The copy input will run first then the remove input, working no problem.

If your compile block looks like this:
```yaml
parameters:
  target_name: removal
  kapitan:
    vars:
      target: ${target_name}
    compile:
      # test removal of a file
      - input_type: remove
        input_paths:
          - compiled/${kapitan:vars:target}/copy_target
        output_path: .
      - input_type: copy
        input_paths:
          - copy_target
        output_path: .
```
The remove input will run first and fail the entire compile because the file it's trying to remove does not exist.

Does that make sense?

uberspot:
Nice, that's what i expected. But I wasm't sure if reclass maintains/guarantees the order or not. I can definitely imagine users could fall into scenarios where the order of these causes weird behavior when compiling. Would be great to add a note in documentation explaining this behavior explicitly. Everything else looks good. :)

sebradloff:
@uberspot totally agreed on the docs. I want to maybe see if we can start creating an index for different terminology. I defined this idea of "phases" during the compile command and I think that helps frame the discussion about what's going on here: https://github.com/deepmind/kapitan/pull/588/files#diff-bf1b6d5bfe119bf7d3ecaf4dc0636334R13

Specifically that the commands run sequentially. I do think I need to add something about the search path in the documentation but not sure where.

uberspot:
Given the context I would add it as a note right after all input related documentation (i think compile.md?). We can always move it to a more appropriate place in the future. :)

Issue #366, opened 2019-09-12
Query: are there any plans to support Azure vaults ?


anirban1c:
love kapitan, any plans to support azure vaults ?

ramaro:
Thanks! No plans at the moment but we are always happy to help and review PRs 😄

anirban1c:
sure, will do

anirban1c:
would it be possible to get an invite for slack channel ? have been going through the code and have a few questions about what az vault features  to use/implement, ask some silly questions to check if I understood the concepts, thanks

ramaro:
@anirban1c does https://slack.k8s.io/ work for you?

anirban1c:
yes, thanks @ramaro , got it

uberspot:
Done by https://github.com/kapicorp/kapitan/pull/590

PR #590, opened 2020-08-19
Support for Azure KMS backend

AlphaRoy14:
Fixes issue #366 

Todo:
- ~Add test cases~
- ~Documentation~


1arrow:
This is awesome

kapitan/refs/secrets/azkms.py
```diff
@@ -0,0 +1,179 @@
...
+            if isinstance(data, str):
+                _data = data.encode()
+        try:
+            ciphertext = ""
+            # Mocking encrypted response for tests
+            if key == "mock":
+                ciphertext = base64.b64encode(_data)
+            else:
+
+                request = azkms_obj(key).encrypt(EncryptionAlgorithm.rsa_oaep, _data)
```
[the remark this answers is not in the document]
uberspot: This seems to be "rsa_oaep" still?

uberspot (changes requested):
A tiny change left. Also could you rebase to latest master so that we can directly merge? :1st_place_medal:

kapitan/refs/secrets/azkms.py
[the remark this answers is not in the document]
AlphaRoy14: Hmm I'm not sure why. In the commit (https://github.com/deepmind/kapitan/pull/590/commits/db39a67fb218d7a5724eda3a9c0187100687f064) It shows the change .

uberspot (approved):

AlphaRoy14:
Hi @uberspot, I was running some local tests and there seems to be a problem when decrypting with `rsa_oaep_256` in the decrypt function.
```sh
kapitan.refs.secrets.azkms.AzureKMSError: (BadParameter) The parameter is incorrect.
```
However, when using just `rsa_oaep` as the encryption algorithm in the decrypt function, it works fine even though `rsa_oaep_256` is used for encryption. `rsa_oaep_256` should work fine according to the [docs](https://azuresdkdocs.blob.core.windows.net/$web/python/azure-keyvault-keys/4.0.0/azure.keyvault.keys.crypto.html). I'm still investigating as to why this is happening, might be a bug in the azure API. Will update as soon as I find it. In the meanwhile should I change the decrypt algorithm to `rsa_oaep` with a comment?

uberspot:
> Hi @uberspot, I was running some local tests and there seems to be a problem when decrypting with `rsa_oaep_256` in the decrypt function.
> 
> ```shell
> kapitan.refs.secrets.azkms.AzureKMSError: (BadParameter) The parameter is incorrect.
> ```
> 
> However, when using just `rsa_oaep` as the encryption algorithm in the decrypt function, it works fine even though `rsa_oaep_256` is used for encryption. `rsa_oaep_256` should work fine according to the [docs](https://azuresdkdocs.blob.core.windows.net/$web/python/azure-keyvault-keys/4.0.0/azure.keyvault.keys.crypto.html). I'm still investigating as to why this is happening, might be a bug in the azure API. Will update as soon as I find it. In the meanwhile should I change the decrypt algorithm to `rsa_oaep` with a comment?

Hmm, odd. I don't see any particular settings needed for this to work.

Did you try regenerating all azure secrets you had locally too? 
And also did you run this with debug? To get some more verbose output from the azure library about it?

AlphaRoy14:
It was my bad, I was trying to reveal an older secret encrypted with `rsa_oaep`. Thanks for your help!

kapitan/refs/cmd_parser.py
```diff
@@ -115,6 +116,28 @@ def ref_write(args, ref_controller):
...
+            inv = inventory_reclass(args.inventory_path)
+            kap_inv_params = inv["nodes"][args.target_name]["parameters"]["kapitan"]
+            if "secrets" not in kap_inv_params:
+                raise KapitanError(
+                    "parameters.kapitan.secrets not defined in inventory of target {}".format(
+                        args.target_name
+                    )
+                )
+
+            key = kap_inv_params["secrets"]["azkms"]["key"]
```
ramaro, line 132: This will raise TypeError/KeyError if either `parameters.kapitan.secrets.azkms` or `parameters.kapitan.secrets.azkms.key` 
We should either implement jsonschema validation for this parameter structure or just turn it into something like 
```python
try:
  key = kap_inv_params["secrets"]["azkms"]
  except (KeyError, TypeError):
  raise KapitanError(
                    "parameters.kapitan.secrets.azkms.key not defined in inventory of target {}".format(
                        args.target_name
                    )
```
And I just realised we have the same issue for gpg/awskms/gkms...

kapitan/refs/cmd_parser.py
```diff
@@ -238,6 +261,25 @@ def secret_update(args, ref_controller):
...
 
+    elif token_name.startswith("azkms:"):
+        key = args.key
+        if args.target_name:
+            inv = inventory_reclass(args.inventory_path)
+            kap_inv_params = inv["nodes"][args.target_name]["parameters"]["kapitan"]
+            if "secrets" not in kap_inv_params:
+                raise KapitanError("parameters.kapitan.secrets not defined in {}".format(args.target_name))
+
+            key = kap_inv_params["secrets"]["azkms"]["key"]
```
ramaro, line 272: Same issue here (this actually needs refactoring, later, not now)

kapitan/refs/secrets/azkms.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def azkms_obj(key_id):
+    """
+    Return Azure Key Vault Object
+    """
+    # e.g of key_id https://kapitanbackend.vault.azure.net/keys/myKey/deadbeef
+    if not cached.azkms_obj:
+
+        attrs = key_id.split("/")
```
ramaro, line 34: Wouldn't it be cleaner to parse this with `urllib.parse.urlparse()`?

ramaro (changes requested):
Looks good, just some minor things! :+1:

kapitan/refs/secrets/azkms.py
AlphaRoy14: Even even with `urllib.parse.urlparse()` we'll have to `slpit()` to get the `key_name` and `key_version`. Plus an added dependency of `urllib`. Besides this, the URL must be introduced by '//' for `urllib` to correctly parse it.

ramaro: Yes, we'd still need split, but urllib is part of the standard library and this works just fine:

```python
In [2]: urllib.parse.urlparse("https://kapitanbackend.vault.azure.net/keys/myKey/deadbeef").path.split("/")
Out[2]: ['', 'keys', 'myKey', 'deadbeef']

In [3]: urllib.parse.urlparse("kapitanbackend.vault.azure.net/keys/myKey/deadbeef").path.split("/")
Out[3]: ['kapitanbackend.vault.azure.net', 'keys', 'myKey', 'deadbeef']
```

uberspot:
Nice, one last rebase to master remaining? So it runs test against latest changes too.

ramaro:
@AlphaRoy14 I apologise for missing this! Do you have some handy documentation we can stick in the kapitan.dev refs section for this?

AlphaRoy14:
Hi @ramaro, Thanks for the review! I have already added a few lines about `azkms`  in `docs/secrets.md` which is homogeneous to the way `gkms` and `awskms` are written. Would you like me to add a new subsection about `azkms` too?

ramaro:
@AlphaRoy14 that would be great. Yes please 👍

ramaro (approved):

PR #610, opened 2020-09-25
Update reclass dependency

srueg:
## Proposed Changes

Update reclass dependency to enable parameter overrides in interpolating class names https://github.com/kapicorp/reclass/pull/2


ramaro:
Thanks @srueg !

ramaro (approved):

Issue #622, opened 2020-10-08
VAULT_SKIP_VERIFY breaks with yaml boolean defined in parameters


haggishunk:
**Describe the bug/feature**
Setting `VAULT_SKIP_VERIFY` in `parameters.kapitan.secrets.vaultkv` results in an exception when using yaml boolean `false` or `true`.  Setting to a string value throws no exception and functions as expected.

The vaultkv secrets backend supports several settings, some of which can be set via environment variable or in kapitan parameters and others which can only be set by env vars.  The `VAULT_SKIP_VERIFY` setting can be set by either.  The current implementation pulls from the `parameters.kapitan.secrets.vaultkv` dict first then falls back to a defined env var.  The logic to determine the vault skip verify relies on a `str.lower()` operation and subsequent comparison against `"false"`.  The `.lower()` throws an exception when the `VAULT_SKIP_VERIFY` is a yaml boolean interpreted as a python `bool` type.

**To Reproduce**
1. Set a vaultkv secret backend (any auth should suffice):
```
parameters:                                                                                                                                                   
  kapitan:                                                                                                                                                    
    vars:                                                                                                                                                     
      target: ${target_name}                                                                                                                                  
      managed_by: kapitan                                                                                                                                     
    secrets:                                                                                                                                                  
      vaultkv:                                                                                                                                                
        auth: github                                                                                                                                          
        engine: kv-v2                                                                                                                                         
        mount: kv                                                                                                                                             
        VAULT_SKIP_VERIFY: false                                                                                                                              
```
2. Write a kapitan ref referencing a vault kv store secret:
```
echo "gosh/darn:this" | kapitan refs --write vaultkv:secrets/test --target test-target -f -
```
3. Include inventory referencing the kapitan ref
```
parameters:                                                                                                                                                                                                                                       
  test_param: ?{vaultkv:secrets/test}    
```
4. Compile and reveal secrets
```
kapitan compile --reveal --target test-target
```
5. See error
```
Unknown (Non-Kapitan) Error occurred                                                                                                                          
multiprocessing.pool.RemoteTraceback: 
"""                                                                            
Traceback (most recent call last):                                                                                                                                                                                                
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 125, in worker                                                                                                                                                          
    result = (True, func(*args, **kwds))                                                                                                                                                                                          
  File "<install_location>/kapitan/kapitan/targets.py", line 453, in compile_target                                                                                                                                               
[...]

PR #623, opened 2020-10-08
VAULT_SKIP_VERIFY supports with yaml boolean defined in parameters

haggishunk:
vaultkv `skip_verify` converges on str type to support subsequent `.lower()` string operation for determining `client_parameters["verify"]`.  as `skip_verify` can be sourced from yaml kapitan parameters or an environment variable, it is decidedly simpler to convert any python bool type to string (ie "True" or "False") rather than evaluate truthiness of a mixed bag of bool types and strings.

Fixes issue #622


haggishunk:
@googlebot I signed it!

ramaro (approved):

uberspot (approved):

PR #628, opened 2020-10-09
add libssl for static(x) binary

ramaro:


uberspot:
Tests failed

ramaro:
`FileNotFoundError: [Errno 2] No such file or directory: '/usr/lib/x86_64-linux-gnu/libssl.so'` OK so we can't assume a symlink to `libssl.so` is always present. Will need to write a function to expand to the canonical path and include in staticx linking then.

ademariag (approved):

uberspot (approved):

PR #631, opened 2020-10-22
[RFC] Python packages for helm binaries

chrisglass:
This pull requests proposes to move the helm bindings for kapitan to their own (python) package, [currently available here](https://github.com/chrisglass/kapitan-helm-bindings).

While this branch and the related repository do work today, some polish would still be needed before a full release. In particular, documentation and tests have not been updated yet.

Please use this pull request to discuss the approach taken, keeping in mind that it is not yet "merge ready". I will fix the tests and documentation etc... once the approach is validated.

Caveats: for this proof of concept only linux is supported, but work to publish macos binaries is planned.

## Testing this PR manually

Checkout this branch, then in a pristine virtualenv:

```
pip install -r requirements.txt
pip install -e .  # This installs kapitan from the current branch
pip install --index-url https://test.pypi.org/simple/ kapitan_helm  # Notice: this is *test* pypi
kapitan compile ...
```

This should allow you to use helm charts as inputs.

## Running automated tests

From a pristine virualenv
```
pip install -r requirements.txt
make test
```


chrisglass:
@googlebot I signed it!

ramaro:
Hey @chrisglass thanks very for this! I think it makes perfect sense to make it a binary, separate dependency. Would you mind if we fork your [kapitan-helm-bindings](https://github.com/chrisglass/kapitan-helm-bindings) repo under [kapicorp](github.com/kapicorp) to make an official pypi package?

chrisglass:
@ramaro I don't mind at all, on the contrary I think that's the right approach. You should also publish packages to pypi ASAP in my humble opinion (at least to claim the name, if it is suitable).

Building the binaries from that repository should be straightforward, hopefully, but don't hesitate to reach out if you need to.

chrisglass:
The last failing test seems to be related to the binary installer (it assumes built binaries). I'm no sure if the binary installer really has a reason to exist should this PR land - but if it does, then I will need guidance on how to solve the problem with the tests (I'm not really sure what pyinstaller is doing).

ramaro:
@chrisglass regarding the binary, have you trying adding kapitan_helm as a `--hidden-import`, like in https://github.com/deepmind/kapitan/blob/master/scripts/pyinstaller.sh#L19 ?

chrisglass:
That doesn't seem to help unfortunately.

I guess I'll try and spend some time during the end of year to dig further into the failure.

ramaro:
closing in favor or #701

PR #648, opened 2020-11-06
Update to Helm 3

### LINKED ISSUE #647: Support for Helm 3
**Describe the bug/feature**
Are there any plans for supporting Helm 3 for rendering charts?

I could imagine to support both versions by introducing a new input type `helm3` for example. Fully switching to Helm 3 might have some compatibility issues as well. For example charts which use the deprecated `crd-install` hook won't render correctly with Helm 3.

**Additional context**
Some Helm charts will only work correctly when rendered with Helm version 3. For example whenever CRDs are included in the new [crds folder](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) they won't be picked up by Helm 2. For example the [Crossplane chart](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) won't render correctly (CRDs missing) due to this.

srueg:
Closes #647 

## Proposed Changes

  - Update to Helm 3


srueg:
PR to get rid of the debug output of Helm: https://github.com/helm/helm/pull/8992

ramaro:
Hey @srueg thanks for this! Would it make sense perhaps to join forces with @chrisglass on #631?

tobru:
Would it make sense to continue with this PR without waiting for #631? Or are there any dependencies on it?

srueg:
Superseded by https://github.com/kapicorp/kapitan/pull/701

PR #670, opened 2020-12-23
Fix searchvar not finding vars with false values

roman8422:
Summary:  currently, searchvar doesn't find variables if the value of the variable is something considered False in python (like False itself, 0, empty string, list or dictionary). This change fixes that.


ramaro:
@roman8422 good find and thanks for the fix!

Issue #543, opened 2020-06-26
Enable Setting `KubeVersion` for Helm Template


srueg:
**Describe the bug/feature**
When rendering a Helm chart, the used kube version is fixed to `v1.9`. This leads to issues if this version is used in Helm charts (as it is in [cert-manager](https://github.com/jetstack/cert-manager/blob/f0231cc5d5a7f74fa0d77395546cef2cd0b80efd/deploy/charts/cert-manager/templates/webhook-mutating-webhook.yaml#L29) for example).

**To Reproduce**
1. Use an input type `helm`
2. Use the `{{ .Capabilities.KubeVersion.GitVersion }}` variable in the chart
3. See fixed version

**Expected behavior**
I'd like to have an option to override this default version, ideally via the inventory.

** If it's a bug (please complete the following information):**
 - `python --version`: `Python 3.8.3`
 - `pip3 --version`: `pip 20.0.2`
 - `Are you using pyenv or virtualenv?` virtualenv

**Additional context**
The fixed version is set [here](https://github.com/deepmind/kapitan/blob/master/kapitan/inputs/helm/template.go#L207) and defined [here](https://github.com/helm/helm/blob/release-2.14/pkg/chartutil/capabilities.go#L31).

ramaro:
Fixed in #671

PR #671, opened 2020-12-24
Add parameter to set kube version for Helm charts rendering

dverbeir:
Fixes issue #543

## Proposed Changes

Some charts generate manifests slightly differently depending on the target Kubernetes version (e.g. targeting different APIs). In order to obtain the appropriate rendered manifests,  the Kubernetes version can be set using the kapitan `kube_version` parameter:

    parameters:
      kapitan:
        kube_version: 1.16

- target_obj is provided to the Helm "compiler" which extracts the `kube_version` parameter and passes it to the go libtemplate to be set when rendering the chart.


ramaro:
Hey @dverbeir thanks for this! I think we could make it more standard with the rest of the helm options if implementing `kube_version` instead as a parameter for the helm compile item, e.g.

```yaml
parameters:
   compile:
      - input_type: helm
        output_path: .
        input_paths:
          - charts/prometheus
        kube_version: 1.16
        ....
```

That way we can still have a global value passed to it and have targets define their own e.g.:
```yaml
parameters:
  my_kube_version: 1.20

  kapitan:
   compile:
      - input_type: helm
        output_path: .
        input_paths:
          - charts/prometheus
        kube_version: ${my_kube_version}
        ....
```

Also this means that `target_obj` wouldn't need to be passed into `Helm()`, as this value will be accessible via `comp_obj`.
What do you think?

dverbeir:
> Also this means that `target_obj` wouldn't need to be passed into `Helm()`, as this value will be accessible via `comp_obj`.
> What do you think?

Unfortunately, as `input:compile_file()` doesn't receive the `comp_obj`, this doesn't really simplify anything. But I agree that it looks more standard to have `kube_version` under the compile item. I pushed a revised changeset.

dverbeir:
@googlebot I signed it!

dverbeir:
@googlebot I signed it!

dverbeir:
@googlebot What a pain!

Could somebody check what's wrong with Google CLA? We set up a corporate CLA 4 days ago and it's still not active... Don't you want contributions???

uberspot:
> @googlebot What a pain!
> 
> Could somebody check what's wrong with Google CLA? We set up a corporate CLA 4 days ago and it's still not active... Don't you want contributions???

Hmm, we don't have full control over it. Are you using the exact same email for your git commits as the one you used for the CLA?  Afaik that's the main thing that has to match (?).

dverbeir:
> > @googlebot What a pain!
> > Could somebody check what's wrong with Google CLA? We set up a corporate CLA 4 days ago and it's still not active... Don't you want contributions???
> 
> Hmm, we don't have full control over it. Are you using the exact same email for your git commits as the one you used for the CLA? Afaik that's the main thing that has to match (?).

I understand it's not under you control, of course. Sorry for letting my frustration out...
I was just hoping you could help accelerate this.

Exact same email is used. Problem is likely with corporate CLA itself. We contacted cla-submissions@google.com but no response so far...

docs/compile.md
```diff
@@ -355,6 +355,7 @@ parameters:
       	namespace: <substitutes_.Release.Namespace>
       	name_template: <namespace_template>
       	release_name: <chart_release_name>
+      kube_version: "1.16"
```
ramaro, line 358: nit: needs an extra space

ramaro (changes requested):
Looking good! Just the small nit and we can merge. Don't worry about the CLA issue as you signed it, sometimes the bot is flaky and we can still merge.

docs/compile.md
dverbeir: I think it was correct. You were probably mislead by the fact that the above lines (but not all) included tab characters. I replaced those by spaces. Hope you don't mind.

Also, I changed the "1.16" to <target_kubernetes_version_string> to harmonize with other params, and moved the example into the description.

ademariag:
@dverbeir as a side effect of moving the repository to kapicorp, the CLA problem should go away :)

PR #681, opened 2021-01-24
Handle KeyError in kapitan secrets

AlphaRoy14:
- add json schema validation for kapitan secrets
- add KeyError exception handling for kapitan secrets


ramaro:
thanks @AlphaRoy14 !

PR #682, opened 2021-01-26
Update references to GitHub project

srueg:
Since the Kapitan project moved to the kapicorp GitHub org.


ramaro:
Thanks @srueg !

PR #687, opened 2021-01-29
Ignore copy file on missing source

ademariag:
Fixes issue #688

## Proposed Changes

Allows the "copy" input type to continue even if the source path is missing. 

The main motivation behins this change is to enable a workaround to handle terraform lock files which are now being produced when running terraform init, see https://github.com/hashicorp/terraform/issues/27241 

The solution would be to allow adding the following copy settings:
```
 kapitan:
    compile:
      - input_type: copy
        ignore_missing: true
        input_paths:
          - resources/state/${target_name}/.terraform.lock.hcl
        output_path: terraform/
```

Without this feature, the compile process would fail for all new environments where the state file has not already been placed in the state folder.

### DIFF
--- a/kapitan/inputs/base.py
+++ b/kapitan/inputs/base.py
@@ -44,7 +44,7 @@ class InputType(object):
             # remove duplicate inputs
             inputs = set(inputs)
             ignore_missing = comp_obj.get("ignore_missing", False)
-            if len(inputs) == 0 and ignore_missing == False:
+            if len(inputs) == 0 and not ignore_missing:
                 raise CompileError(
                     "Compile error: {} for target: {} not found in "
                     "search_paths: {}".format(input_path, ext_vars["target"], self.search_paths)
--- a/kapitan/inputs/base.py
+++ b/kapitan/inputs/base.py
@@ -43,7 +43,8 @@ class InputType(object):
             inputs = list(itertools.chain.from_iterable(globbed_paths))
             # remove duplicate inputs
             inputs = set(inputs)
-            if len(inputs) == 0:
+            ignore_missing = comp_obj.get("ignore_missing", False)
+            if len(inputs) == 0 and ignore_missing == False:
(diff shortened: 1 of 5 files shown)


kapitan/inputs/base.py
```diff
@@ -43,7 +43,8 @@ def compile_obj(self, comp_obj, ext_vars, **kwargs):
             inputs = list(itertools.chain.from_iterable(globbed_paths))
             # remove duplicate inputs
             inputs = set(inputs)
-            if len(inputs) == 0:
+            ignore_missing = comp_obj.get("ignore_missing", False)
+            if len(inputs) == 0 and ignore_missing == False:
```
uberspot, line 47: "and not ignore_missing" is slightly more pythonic(?)

uberspot (approved):

kapitan/inputs/copy.py
```diff
@@ -25,17 +26,22 @@ def compile_file(self, file_path, compile_path, ext_vars, **kwargs):
...
                 else:
-                    os.makedirs(compile_path, exist_ok=True)
-                    shutil.copy2(file_path, os.path.join(compile_path, os.path.basename(file_path)))
-            else:
-                compile_path = os.path.abspath(compile_path)  # Resolve relative paths
-                copy_tree(file_path, compile_path)
+                    compile_path = os.path.abspath(compile_path)  # Resolve relative paths
+                    copy_tree(file_path, compile_path)
+            elif ignore_missing == False:
+                raise OSError(f"Path {file_path} does not exist and `ignore_missing` is {ignore_missing}")
```
ramaro, line 44: I guess it's better to raise CompileError to keep kapitan error handling/printing consistent

ademariag: @ramaro I used the same error that was already raised in line 39. Shall I also change that?

PR #701, opened 2021-03-03
Kap 09

Jean-Daniel:
Implementation of kap-9: Bring your own Helm.

fixes: #368 #647 

## Proposed Changes

  - Replace helm cffi by Helm CLI invocation. This greatly simplify the helm logic and make adoption of new capabilities easier (just add a flag to the command line invocation)

This PR takes care of both helm_fetch and helm input. It completely removes the need for golang and cffi to build and install Kapitan.

It is based on and tested Helm 3 only as Helm 2 [support ended the November 13, 2020](https://helm.sh/blog/helm-v2-deprecation-timeline/).


ademariag:
Thank you @Jean-Daniel, this is a great idea. 

I know that @simu and @srueg are also interested in this and would also like to get their opinion. 

Personally I like the idea of moving away from helm cffi, although we need to consider other angles.

srueg:
That's great, thanks!
I don't think there's any downside to this (at least from our side) and it would solve a lot of problems indeed 👍 

I'll do some tests and review this PR 🎉

kapitan/inputs/helm.py
```diff
@@ -0,0 +1,123 @@
...
+        if kwargs.get("name_template", None):
+            args.append("--name-template")
+            args.append(kwargs["name_template"])
+        else:
+            args.append(kwargs.get("release_name", "--generate-name"))
+
+        # uses absolute path to make sure helm interpret it as a
+        # local dir and not a chart_name that it should download.
+        args.append(os.path.abspath(chart_dir))
+
```
srueg, line 122: Should we add `--include-crds` or make it configurable?

srueg: And probably `--skip-tests`?

srueg (approved):
LGTM, I ran tests with some of our setups and apart from the `app.kubernetes.io/managed-by=Helm` label change and some reordered RBAC, all compiled output stays the same 👍

srueg:
I tried to get rid of some of the debug output in this PR: https://github.com/helm/helm/pull/8992
Maybe if you add your vote there?

kapitan/inputs/helm.py
Jean-Daniel: Yes, this is both 2 reasonable flags to add by defaults. 
Should we add `skip_crds` and `include_tests` configuration flags then.

The only drawback I see with this names, is that helm already has an unrelated `--skip-crds` flags.

Jean-Daniel:
> I tried to get rid of some of the debug output in this PR: [helm/helm#8992](https://github.com/helm/helm/pull/8992)
> Maybe if you add your vote there?
> 
> Or should we try to filter the stdout of the `helm` cli to remove the `wrote [...]` debug logs?

As the Kapitan debug output already print the generated files names, we can probably disable helm stdout completely (`stdout=subprocess.DEVNULL`).

Or enable it only when debug log is enabled.

kapitan/inputs/helm.py
srueg: I'd suggest to add `include_crds` and `skip_tests` just with a default of `true` to name them after the flags.

srueg (approved):

ramaro:
Btw @sebradloff you might want to have a look at this 😄

kapitan/inputs/helm.py
```diff
@@ -0,0 +1,129 @@
...
+            args.append(kwargs["helm_values_file"])
+
+        if kwargs.get("namespace", None):
+            args.append("-n")
+            args.append(kwargs["namespace"])
+
+        args.append("--output-dir")
+        args.append(output_path)
+
+        if self.kube_version:
```
sebradloff, line 106: This is an amazing PR! I would recommend however to take `helm_params` as an input for all the additional flags like `api-versions`, `validate`, etc. I think it makes it a little more flexible.

Jean-Daniel: I pushed a new version that supports most helm flags. There is just a bunch of flags that are disabled because they either affect the output, or are useless for templating and only relevant on install.

Note that `helm template` has a `--release-name` flag that conflicts with the `release_name` helm_params. This version now recommend to use `name` instead of `release_name`. To supports inventory that keep using `release_name`, the helm input check if the `release_name` value is a  `bool`. If not, it prints a deprecation warning and keep using the `release_name` value as `name`.

kapitan/inputs/helm.py
```diff
@@ -84,46 +99,80 @@ def compile_file(self, file_path, compile_path, ext_vars, **kwargs):
...
+            if param == "values":
+                logger.warning(
+                    "helm '%s' flag is not supported. Use 'helm_values_files' to specify template values files",
+                    param,
+                )
+                continue
+
+            if param in HELM_DENIED_FLAGS:
+                logger.warning("helm flag '%s' is not supported and will be ignored.", param)
+                continue
```
srueg, line 134: Should we maybe abort here? Warning output might be missed and lead to unexpected behavior if the flags are just dropped.

srueg (changes requested):

kapitan/inputs/helm.py
sebradloff: In regards to the conflict of `name` and `release_name`, I think your solution make sense of printing the warning makes sense.

.travis.yml
```diff
@@ -24,18 +24,17 @@ before_install:
...
   # Loop until update succeeds (timeouts can occur)
   - travis_retry $(! sudo apt-get -qq update 2>&1 | grep Failed)
-  - sudo apt-get install -y gnupg2 git
-  - sudo apt-get install -y golang-1.14
-  - export PATH="/usr/lib/go-1.14/bin:$PATH"
-  - export GOROOT=/usr/lib/go-1.14
-  - go version
+  - sudo apt-get install -y gnupg2 git curl
+  # Install Helm Binary
+  - curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
```
sebradloff, line 29: should we be pinning the version of helm3 that gets downloaded? I believe if I'm reading the `get_helm.sh` script correctly, we could technically set a `DESIRED_VERSION` to control the helm3 version being downloaded instead of defaulting to latest.

Dockerfile
```diff
@@ -31,10 +17,15 @@ RUN apt-get update \
     && python -m venv /opt/venv \
     && pip install --upgrade pip yq wheel \
     && pip install -r requirements.txt \
-    && ./kapitan/inputs/helm/build.sh \
-    && ./kapitan/dependency_manager/helm/build.sh \
     && pip install .
 
+# Install Helm
+RUN apt-get install --no-install-recommends -y curl \
+    && curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 \
```
sebradloff, line 24: should we be pinning the version of helm3 that gets downloaded? I believe if I'm reading the get_helm.sh script correctly, we could technically set a DESIRED_VERSION to control the helm3 version being downloaded instead of defaulting to latest.

Jean-Daniel: get_helm.sh supports `--version` flag. What would be the benefit of pinning the version ?

sebradloff: I feel like in general there are bug fixes here and there that happen with any software. I feel like since we directly depend on the helm binary and we have an "open" interface with `helm_params` it might be worth being methodical on upgrading versions. Just a thought, don't think it's required.

Jean-Daniel: The get_helm.sh script is design to fetch only v3 version. So, all updates should be backward compatible.

Just like this image is using latest python3.7, git or gnupg binaries, I don't think it should cause issues using the latest helm 3 release.

ramaro:
@Jean-Daniel @srueg @sebradloff are we happy with the state of this?

Jean-Daniel:
I'm using it for some times now and haven't encounter any issue yet. So I'm fine with it.

kapitan/helm_cli.py
```diff
@@ -0,0 +1,20 @@
...
+from subprocess import PIPE, DEVNULL
+
+logger = logging.getLogger(__name__)
+
+
+def helm_cli(args, stdout=None, verbose=False):
+    try:
+        logger.debug("launching helm with arguments: %s", args)
+        res = subprocess.run(
+            args=["helm"] + args, stderr=PIPE, stdout=stdout or (PIPE if verbose else DEVNULL)
```
ccremer, line 12: I think it would make sense if we can somehow configure which helm binary to use instead of relying on a hardcoded value. Some users might have a dual-helm setup with helm v2 and v3 and their helm binary might be called `helm3` or so. Maybe try to use the binary in an env var like "HELM_BIN" (or whatever appropriate) and default to `helm` if not defined?

e.g. https://github.com/roboll/helmfile allows to set the "helmBinary" config.

ramaro: For that, I think it would be best to support a `helm_path` option along the other ones

sebradloff:
> @Jean-Daniel @srueg @sebradloff are we happy with the state of this?

I think this PR is great. Helm2 binary as an option sounds like a good idea, since some people might be stuck with old open source charts.

kapitan/helm_cli.py
Jean-Daniel: Both use cases are valid IMHO. Some people may need to set the path globally (if they want to use a specific helm for everything), and some will need to set it per compile action (to compile a `helm2` chart for instance).

ccremer: yeah, sure. being able to specify it globally or per compilation should give enough flexibility

zstoth:
@Jean-Daniel is anything blocking this?

ramaro:
Hey @Jean-Daniel do you wanna master rebase this? It should fix the failed binary tests as we discontinued the binary in #740

Jean-Daniel:
> Hey @Jean-Daniel do you wanna master rebase this? It should fix the failed binary tests as we discontinued the binary in #740

Done. Note that #740 erroneously remove the `build_helm_binding` target in the Makefile, which breaks helm support (until the PR is merged).

ademariag:
I did a quick test with my repo, and this pretty much doubles the compile speed (for prometeus chart) in my case

### before
```
time kapitan compile -t intranet-cluster-infra
Compiled intranet-cluster-infra (8.97s)
real    0m12.099s
user    0m11.563s
sys     0m1.508s
```
### after
```
time kapitan compile -t intranet-cluster-infra
Compiled intranet-cluster-infra (3.41s)
real    0m6.199s
user    0m8.038s
sys     0m1.285s
```

ademariag:
A tiny issue I have found, is that the produced files have a different formatting (which we already experienced in https://github.com/kapicorp/kapitan/issues/360) so that replacing the old helm support for the new one will make the review process tedious (because we will have lots of formatting diffs)

Do you know whether it is possible to still prettify the helm output even with this new input type?

ramaro:
We can do the same we've done before here https://github.com/kapicorp/kapitan/pull/385/files @ademariag

ramaro (approved):

PR #707, opened 2021-03-09
Fix usage of undefined variable.

Jean-Daniel:
parser is not defined in generate_inventory

so call to `parser.error("--pattern requires --target_name")` fails.


ramaro (approved):
Thank you! :D

Issue #144, opened 2018-09-03
replace jsonnet with go-jsonnet


ramaro:

mithil2311:
can you explain about this problem statement?

adrianchifor:
@mithil2311 We are currently using the direct Python library which uses the C++ version of jsonnet. To increase the speed of jsonnet compilation, we need to move to the Go version https://github.com/google/go-jsonnet and create Python bindings for it.

sbarzowski:
(Jsonnet dev here)

FYI basic C bindings for go-jsonnet are probably going to be merged soon (PR: https://github.com/google/go-jsonnet/pull/257). The main missing part is the support for native functions. Other than that it's a drop-in replacement.

My understanding is that the Python bindings just use C bindings internally, so no changes outside of `setup.py` should be necessary.

yoshi-1224:
HI @sbarzowski,

Please excuse my lack of understanding. Could you kindly explain what you mean by
> the main missing part is the support for native functions

sbarzowski:
Sure. I meant that this function from the libjsonnet.h is not implemented yet in the bindings:
```
/** Register a native extension.
 *
 * This will appear in Jsonnet as a function type and can be accessed from std.nativeExt("foo").
 *
 * DO NOT register native callbacks with side-effects!  Jsonnet is a lazy functional language and
 * will call your function when you least expect it, more times than you expect, or not at all.
 *
 * \param vm The vm.
 * \param name The name of the function as visible to Jsonnet code, e.g. "foo".
 * \param cb The PURE function that implements the behavior you want.
 * \param ctx User pointer, stash non-global state you need here.
 * \param params NULL-terminated array of the names of the params.  Must be valid identifiers.
 */
void jsonnet_native_callback(struct JsonnetVm *vm, const char *name, JsonnetNativeCallback *cb,
                             void *ctx, const char *const *params);
```

Native extensions are a feature which allows calling your own C function from Jsonnet - I can see that Kapitan uses it for a few functions (https://github.com/deepmind/kapitan/blob/master/kapitan/lib/kapitan.libjsonnet). 

The `jsonnet_json_*` family of functions (which is only useful for native functions) is also not implemented in the bindings. 

The other missing parts unrelated to native functions are evaluation variants `*_stream` and `*_multi` and import callbacks (`jsonnet_import_callback` function).

[...]

Issue #704, opened 2021-03-07
crash in macOS when forking


Jean-Daniel:
**Describe the bug/feature**
Most of macOS frameworks and runtime libraries are not fork safe. While using only low-level libraries may work, in practice, even Python relies on some fork unsafe code, which make it globally fork unsafe as can be shown by this crash when trying to fetch a HTTP file.

```
Dependency https://github.com/…/controller.yaml: fetching now
objc[27302]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[27302]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

**To Reproduce**
1. try to fetch/compile a simple inventory containing a HTTP dependency on macOS 11/python 3.9 (maybe crash with other versions too).

See above.

**Expected behavior**
Do not crash.

** If it's a bug (please complete the following information):**
 - `python --version`: Python 3.9.2 (brew)
 - `Are you using pyenv or virtualenv?`: python3 -m venv

**Additional context**
macOS 11.2.2

Jean-Daniel:
For the record:

https://docs.python.org/3/library/multiprocessing.html

> Changed in version 3.8: On macOS, the spawn start method is now the default. The fork start method should be considered unsafe as it can lead to crashes of the subprocess. See [bpo-33725](https://bugs.python.org/issue33725).

ramaro:
Thanks for reporting @Jean-Daniel ! It's unfortunately a very frequent problem. I guess going forward, we should remove multiprocessing from Kapitan when go-jsonnet is good and async http fetching instead for dependencies.
This also reminds me that we need to setup a macOS pipeline to run tests as we currently only do that on Linux.

MatteoVoges:
resolved by #753

PR #728, opened 2021-04-02
Gojsonnet

pvanderlinden:
Fixes issue #144 
Might also solve #704

## Proposed Changes

  - optional use of gojsonnet
  - use spawn method instead of fork for multiprocessing (as gojsonnet will not work)
  - copy `cached` as a parameter. I have not looked into what it is actually used for, but this will make sure using spawn will actually work.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -29,7 +29,7 @@ COPY ./setup.py ./setup.py
 # Build final image
 FROM python:3.7-buster
 
-ARG CLOUD_SDK_VERSION=276.0.0
+ARG CLOUD_SDK_VERSION=334.0.0
 
 ENV CLOUD_SDK_VERSION=$CLOUD_SDK_VERSION
 ENV KAPP_URL=https://github.com/k14s/kapp/releases/download/v0.19.0/kapp-linux-amd64
--- a/Dockerfile.pyinstaller
+++ b/Dockerfile.pyinstaller
@@ -24,7 +24,7 @@ FROM python:3.7-slim-stretch
 
 SHELL ["/bin/bash", "-i", "-c"]
 
-ARG PYINSTALLER_VERSION=3.5
+ARG PYINSTALLER_VERSION=4.2
 
 COPY --from=helm-builder /kapitan /kapitan
 WORKDIR /kapitan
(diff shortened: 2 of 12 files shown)


ademariag:
Thank you @pvanderlinden. Join our #kapitan channel on the kubenetes slack if you haven't already.

pvanderlinden:
Mmh, changing to spawn causes issue with pyinstaller.
It seems like newer versions of pyinstaller solve these issues. Unfortunately the newer version of pyinstaller doesn't work with staticx. The issue has been resolved, but no new release has been yet.
See:
https://github.com/JonathonReinhart/staticx/issues/71
https://github.com/pyinstaller/pyinstaller/pull/5511

ramaro:
Hey @pvanderlinden these look great and work great on macOS! I'm gonna give it a try on Linux just to validate things but looking great so far - thank you 😄

ramaro:
We can also now remove this block https://github.com/kapicorp/kapitan/blob/master/kapitan/cli.py#L99

pvanderlinden:
It looks like pyinstaller has a new release which should fix the issue using spawn and building a binary. I should have time to remove the code you linked and test that :crossed_fingers:

pvanderlinden:
I'm having troubles getting the tests for the binary up and running, even from the master branch. So no idea if my local fix actually fixes anything

pvanderlinden:
@ramaro I'm trying to fix this PR, I might have a solution, but I can't test it. The issue is that I can't get the binary test to work at all, not even on master. I tried to contact you on slack twice in the last month, but unfortunately no response. Any idea how to get this PR moving?

ramaro:
Hey @pvanderlinden apologies, I often miss slack notifications! As previously discussed in a slack thread that I didn't miss, let's discontinue the binary - it's not worth it. I created #740 for this and will be pushing it to master soon. Hopefully this will unblock the Gojsonnet work :) Thanks for the patience.

ramaro:
Removal PR at #741

pvanderlinden:
THanks Ramaro, I have subscribed to the other PR, and will revisit this PR after it's merged

janeklb:
Hi @pvanderlinden, just in case you've missed it #741 was merged a couple of weeks ago.

If you're too busy I'd be more than happy to continue where you (and @ramaro!) left off with this work. Looking forward to having this in the main trunk!

janeklb:
Have opened up a new PR based on your work just to see if travis-ci tests pass: #753
(removed the duplicate 'add cache' & 'enable cache' commits since they were just "merge artefacts", also removed the last two commits which, I assume, were meant to address the binary build failures).

The performance improvement is fantastic.

Also, worth noting that @pvanderlinden's branch fixes an issue that I was seeing @ramaro's original `gojsonnet` branch.
```
Fatal Python error: deallocating None
Python runtime state: finalizing (tstate=0x7fab37405b30)

Current thread 0x00000001134a1e00 (most recent call first):
<no Python frame>
[2]    54358 abort      kapitan compile --verbose
```

The compilation output (that I was seeing) was correct; however, the message is a worrying one.

pvanderlinden:
@janeklb I have been busy so far, I was planning to pick this up soonish. Thanks for rebasing, I will close this one.

PR #747, opened 2021-07-05
fix "Phases of the compile command" markdown 

gabriel-dantas98:
image


ramaro:
@gabriel-dantas98 obrigado 👍🏼 😄

ramaro (approved):

PR #752, opened 2021-07-10
Jinja2 input_parameters

sebradloff:
This PR introduces `input_params` to the `jinja2` input type similar to `input_params` used for the `kadet` input type introduced in #560. This is useful when you want to create a template that can be reused across multiple components and targets. 

Currently with jinja2 templates, we need know the component(s) the template is going to be used for given the `inventory` or `inventory_global` blocks as the way to inject context. On top of that, the compile block is invoked for the template, but the template variables aren't necessarily clear. With `input_params`, it is clear when the compile block is written, what template variables are to be injected.

## Proposed Changes
- `input_params` as attribute for `jinja2` input type

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,0 +1,36 @@
+name: Build and deploy documentation on merge to master
+on:
+  push:
+    branches:
+      - master
+jobs:
+  build:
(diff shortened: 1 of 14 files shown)


ramaro:
Nice and clean! Thanks @sebradloff

PR #753, opened 2021-07-11
Gojsonnet

janeklb:
This is just a copy of #728, rebased from master's HEAD.

Creating a new PR just to see if tests pass in travis-ci


janeklb:
🤔  travis-ci integration seems to be broken -- last few PRs haven't triggered builds either

ramaro:
@janeklb thanks for this. Yes, travis is broken at the moment and we're aware and working on it and will get back to you soon.

kapitan/targets.py
```diff
@@ -86,7 +84,7 @@ def compile_targets(
                 logger.info("No changes since last compilation.")
                 return
 
-    pool = multiprocessing.Pool(parallel)
+    pool = multiprocessing.get_context("spawn").Pool(parallel)
```
pvanderlinden, line 87: It's better to change it here: https://github.com/kapicorp/kapitan/blob/1c1a00b5266e9cc57c95e4430240f325af9ec088/kapitan/cli.py#L102

pvanderlinden (changes requested):
Thanks @janeklb I added a comment to fix a mistake I made in the original PR.

kapitan/targets.py
janeklb: Thanks @pvanderlinden - updated as requested

pvanderlinden (approved):

ramaro:
@janeklb @pvanderlinden thanks both! LGTM, merging.

PR #766, opened 2021-08-10
Allow external compile type to use PATH executable as input

ofen:
Fix for https://github.com/kapicorp/kapitan/issues/695

### DIFF
--- a/docs/compile.md
+++ b/docs/compile.md
@@ -552,15 +552,15 @@ parameters:
     compile:
       - input_type: external
         input_paths:
-          - /usr/local/bin/ytt # path to ytt on system
-        output_path: .
+          - ytt # path to ytt on system
+        output_path: ingresses/ingresses.yaml
         args:
           - -f
           - ingresses/ # directory with ingresses
           - -f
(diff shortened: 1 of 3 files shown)


ofen:
@ramaro could you please take a look

docs/compile.md
```diff
@@ -552,15 +552,15 @@ parameters:
     compile:
       - input_type: external
         input_paths:
-          - /usr/local/bin/ytt # path to ytt on system
-        output_path: .
+          - ytt # path to ytt on system
+        output_path: ingresses/ingresses.yaml
```
ramaro, line 556: Why are we changing this?

ramaro:
Hey @ofen sorry for the delay and thanks for this. I have a question - wouldn't this break compatibility with the current behavior? If so, I think a better option would be to have a new  parameter, e.g. `in_path: true` to enable this.

simu:
I think this should be addresssed in #864

ramaro:
> I think this should be (at least partially) addressed in #864

Yes it is. Thanks. closing.

PR #781, opened 2021-08-23
Raise an error if helm input is compiled twice

laserb:
If an input path is found more than once in the search paths.
Then the second compilation will be wrong,
because the helm values were reset after the first compilation.
See: https://github.com/kapicorp/kapitan/blob/master/kapitan/inputs/helm.py#L87-L89

At least raise an exception if this happens.


ramaro:
@laserb good point! Thank you.

PR #783, opened 2021-09-02
improve jinja2 templating error report for TemplateError

dverbeir:
jinja2.TemplateError does not include the template line number which can be of great help when troubleshooting errors like
   'foo' has no attribute 'bar'

The line number can however be inferred from the stack trace. This was inspired by
https://github.com/saltstack/salt/blob/master/salt/utils/templates.py

Example output:

   Jsonnet error: failed to compile test/main.jsonnet:
    RUNTIME ERROR: Jsonnet jinja2 failed to render templates/abc.j2: Jinja2 TemplateError: 'foo' has no attribute 'bar', at templates/abc.j2:240

### DIFF
--- a/kapitan/utils.py
+++ b/kapitan/utils.py
@@ -16,6 +16,7 @@ import stat
 import sys
 import re
 import tarfile
+import traceback
 from zipfile import ZipFile
 from collections import Counter, defaultdict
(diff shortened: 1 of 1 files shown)


ramaro:
Thank you @dverbeir !

PR #798, opened 2021-11-27
Initial add randomint

Moep90:
## Proposed Changes

- Additional random generator Type: `randomint`

### DIFF
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -67,6 +67,7 @@ Kapitan will inherit the secrets configuration for the specified target, and enc
 When referencing your secret in the inventory during compile, you can use the following functions to automatically generate, encrypt and save your secret:
 
 - `randomstr` - Generates a random string. You can optionally pass the length you want i.e. `||randomstr:32`
+- `randomint` - Generates a random string. You can optionally pass the length you want i.e. `||randomint:32`
 - `base64` - base64 encodes your secret; to be used as a secondary function i.e. `||randomstr|base64`
 - `sha256` - sha256 hashes your secret; to be used as a secondary function i.e. `||randomstr|sha256`. You can optionally pass a salt i.e `||randomstr|sha256:salt` -> becomes `sha256("salt:<generated random string>")`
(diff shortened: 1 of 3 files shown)


docs/secrets.md
```diff
@@ -67,6 +67,7 @@ Kapitan will inherit the secrets configuration for the specified target, and enc
 When referencing your secret in the inventory during compile, you can use the following functions to automatically generate, encrypt and save your secret:
 
 - `randomstr` - Generates a random string. You can optionally pass the length you want i.e. `||randomstr:32`
+- `randomint` - Generates a random string. You can optionally pass the length you want i.e. `||randomint:32`
```
ramaro, line 70: Fix description to `Generates a random int`

kapitan/refs/functions.py
```diff
@@ -48,6 +49,20 @@ def randomstr(ctx, nbytes=""):
         ctx.data = secrets.token_urlsafe()
 
 
+def randomint(ctx, nbytes=""):
```
ramaro, line 52: I see what you're doing here, but just a thought - to me `randomint` should have a behaviour similar to Python `randint()` where it returns a number between its 2 parameters .
Python's `secrets.randbits()` is a bit a different, so maybe call this function `randbits` instead?

kapitan/refs/functions.py
```diff
@@ -48,6 +49,20 @@ def randomstr(ctx, nbytes=""):
...
+    generates a URL-safe text string, containing nbytes random bits
+    sets it to ctx.data
+    """
+    if nbytes:
+        nbytes = int(nbytes)
+        # Generate twice the amount of integer asked for
+        # and then trim them to nbytes length if it's longer
+        ctx.data = secrets.randbits(2 * nbytes)[:nbytes]
+    else:
+        ctx.data = secrets.randbits()
```
ramaro, line 63: `randbits()` requires a parameter, so this won't work here.

kapitan/refs/functions.py
ramaro: (or `randbytes`)

PR #825, opened 2022-02-25
Feature/new_secret_backend_vault_transit

Moep90:
## Proposed Changes

 - add vaulttransit as new secret engine
 - add tests to vaulttransit secret engine for encryption, decryption
 - add documentation for the new secret engine
 - this will allow to use the Hashicorp Vaults Secret Engine called `Transit`
   - usage is similar to GKMS but with vault
   - you can encrypt, decrypt secrets
   - you can use different policies like engineers can encrypt secrets but not decrypt them
   - a key rotation using the same key is allowed and would reencrypt the secrets
   - changing the key will reencrypt the secrets using the kapitan reencrypt way


Moep90:
> How come this was closed? No longer needed?

github usage confusion :-)

see here: https://github.com/kapicorp/kapitan/pull/826

PR #828, opened 2022-03-05
kapitan: fix python tests.

lion24:
This change needs to be merged first: https://github.com/kapicorp/reclass/pull/7

* On the latest version, to render helm template it now uses the helm
  cli directly. However it seems using helm 3.8.0 (the latest stable),
  the default behavior if to release name is specify is to populate with
  the string "release-name", not "RELEASE-NAME" hence all lowercase.
  There's a simple way to reproduce:
  ```
  helm create test-chart
  helm template --include-crds --skip-tests
  ````
  And check how {{ .Release.Name }} gets populated when there's none
  specified.

Signed-off-by: Lionel H <me@nullbyte.be>

Fixes issue #

## Proposed Changes

  -
  -
  -

PR #826, opened 2022-03-01
Feature/new secret backend vault transit

Moep90:
## Proposed Changes

 - add vaulttransit as new secret engine
 - add tests to vaulttransit secret engine for encryption, decryption
 - add documentation for the new secret engine
 - this will allow to use the Hashicorp Vaults Secret Engine called `Transit`
   - usage is similar to GKMS but with vault
   - you can encrypt, decrypt secrets
   - you can use different policies like engineers can encrypt secrets but not decrypt them
   - a key rotation using the same key is allowed and would reencrypt the secrets
   - changing the key will reencrypt the secrets using the kapitan reencrypt way

PR #829, opened 2022-03-07
Open binary files in binary mode and do not encode them

### LINKED ISSUE #791: kapitan refs fail to handle binary files
**Describe the bug/feature**
Kapitan refs fail to correctly encode binary files, while it works ok with text files

**To Reproduce**
```
kapitan refs --write base64:binary/binary -t echo-server --base64 -f /usr/bin/true 
  File "/usr/local/bin/kapitan", line 8, in <module>
    sys.exit(main())
  File "/Users/ademaria/Library/Python/3.9/lib/python/site-packages/kapitan/cli.py", line 562, in main
    args.func(args)
  File "/Users/ademaria/Library/Python/3.9/lib/python/site-packages/kapitan/refs/cmd_parser.py", line 27, in handle_refs_command
    ref_write(args, ref_controller)
  File "/Users/ademaria/Library/Python/3.9/lib/python/site-packages/kapitan/refs/cmd_parser.py", line 50, in ref_write
    data = fp.read()
  File "/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xca in position 0: invalid continuation byte
```

**Expected behavior**
Binary is correctly encoded.

**Screenshots**
If applicable, add screenshots to help explain your problem/idea.

** If it's a bug (please complete the following information):**
 - `python3 --version`: Python 3.9.7
 - `pip3 --version`: pip 21.2.4 from /usr/local/lib/python3.9/site-packages/pip (python 3.9)
 - `Are you using pyenv or virtualenv?` no

**Additional context**
Related to work for https://github.com/kapicorp/kapitan-reference/issues/106

**WORKAROUND**
Simply encode the file before importing. i.e.:

```
base64 /usr/bin/true  | kapitan refs --write base64:binary/binary -t echo-server -f -
```

Tymolc:
Fixes #791

The MIME type of files is checked for the "text" keyword. If it is present, the file will be opened in standard text mode (modifier `r`), if not it is assumed to be binary data and opened in binary mode (modifier `rb`).
In case of a binary file, the UTF-8 encoding will be skipped.

Related to https://github.com/kapicorp/kapitan-reference/issues/106

### DIFF
--- a/kapitan/refs/cmd_parser.py
+++ b/kapitan/refs/cmd_parser.py
@@ -38,8 +38,8 @@ def ref_write(args, ref_controller):
     "Write ref to ref_controller based on cli args"
     token_name = args.write
     file_name = args.file
+    is_binary = args.binary
     data = None
-    is_binary = False
 
     if file_name is None:
         fatal_error("--file is required with --write")
@@ -49,11 +49,12 @@ def ref_write(args, ref_controller):
             data += line
     else:
         mime_type = mimetypes.guess_type(file_name)
-        modifier = "r" if "text" in mime_type or  mime_type[0] is None else "rb"
+        modifier = "rb" if is_binary else "r"
         with open(file_name, modifier) as fp:
-            data = fp.read()
-            if "b" in fp.mode:
-                is_binary = True
+            try:
+                data = fp.read()
+            except UnicodeDecodeError as e:
+                raise KapitanError("Could not read file. Please add '--binary' if the file contains binary data. ({})".format(e))
 
     if token_name.startswith("gpg:"):
         type_name, token_path = token_name.split(":")
--- a/kapitan/refs/cmd_parser.py
+++ b/kapitan/refs/cmd_parser.py
@@ -4,6 +4,7 @@ import base64
 import logging
 import os
 import sys
+import mimetypes
 
 from kapitan.errors import KapitanError, RefHashMismatchError
 from kapitan.refs.base import PlainRef, RefController, Revealer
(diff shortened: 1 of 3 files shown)


Moep90:
tests seams to be fixed with #828

kapitan/refs/cmd_parser.py
```diff
@@ -46,8 +48,12 @@ def ref_write(args, ref_controller):
         for line in sys.stdin:
             data += line
     else:
-        with open(file_name) as fp:
+        mime_type = mimetypes.guess_type(file_name)
+        modifier = "r" if "text" in mime_type else "rb"
+        with open(file_name, modifier) as fp:
             data = fp.read()
+            if 'b' in fp.mode:
+                is_binary = True
```
Tymolc, line 56: Right now this only works when passing a file to `--write`. Using `-` as the filename and using data from `stdin` currently requires the data to be utf-8 encodable (i.e. some kind of text). I think it might come down to a design decision wether data from stdin should always be treated as text or always as binary.
Alternatively, I could imagine adding a `--binary` flag that will always skip UTF-8 encoding

Tymolc:
TODOS: 
* CLI Tests
* PR #826 will probably need adaptations

kapitan/refs/cmd_parser.py
uberspot: I think defaulting to text and having a --binary flag would cover most cases and give people enough flexibility to use this. Most use cases I can think of would work fine with text so it's a good default.

ramaro:
Thank you @Tymolc @Moep90  Merging :D

PR #831, opened 2022-03-08
feature: add new lower_alpha_num ref function

serapch:
## Proposed Changes

  - add lower_alpha_num ref function which generates string without any special characters or capital letters to be compliant with accountName of Azure StorageAccounts - https://docs.microsoft.com/en-us/rest/api/storagerp/storage-accounts/create


ramaro:
Hey @serapch thanks for this!

I know it's really just a nit, but in the interest of uniformity, can we rename the ref function to `loweralphanum`?

ramaro:
Btw tests are failing but this is something being look at #828 and not related to this PR

Cyanopus:
@serapch I would also remove the "sets it to ctx.data" in the secrets.md as these are implementation details that do not concern the users.

serapch:
@ramaro I renamed the function and trimmed the unnecessary description :)

ramaro (approved):

ramaro:
All good now, merging. Thank you @serapch

PR #834, opened 2022-03-24
Add support for Helm charts from OCI registries

graillus:
Since version 3.8.0 Helm supports OCI registries for chart storage [https://helm.sh/docs/topics/registries](https://helm.sh/docs/topics/registries/), but the `helm pull` command used to fetch a chart from an OCI registry slightly changes, the `--repo` flag have to be omitted because it only works with http(s) chart repositories. Helm expects the fully-qualified URL to the chart instead as argument.

## Proposed Changes

This PR adds support for Helm chart dependencies from OCI registries by removing the `--repo` flag passed to the `helm` command when the `oci://` prefix is detected in the `source` property, and pass it as argument instead.

For instance, this configuration:
```yaml
  kapitan:
    dependencies:
    - type: helm
      output_path: charts/example
      source: oci://europe-west1-docker.pkg.dev/my-project/helm-charts/example
      version: 1.2.3
      chart_name: example
```

 will generate the following Helm command:
```shell
helm pull --destination charts/example --untar --version 1.2.3 oci://europe-west1-docker.pkg.dev/my-project/helm-charts/example
```

### DIFF
--- a/docs/external_dependencies.md
+++ b/docs/external_dependencies.md
@@ -176,6 +176,8 @@ Fetches helm charts and any specific subcharts in the `requirements.yaml` file.
 It defaults to the value of the `KAPITAN_HELM_PATH` environment var or simply to `helm` if neither is set.
 You should specify only if you don't want the default behavior.
 
+`source` can be either the URL to a chart repository, or the URL to a chart on an OCI registry (supported since Helm 3.8.0).
+
 ### Usage
 
 ```yaml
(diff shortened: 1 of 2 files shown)


ademariag:
Sweet @graillus ! Thank you, we'll review it!

If you haven't already, join our #kapitan channel on https://kubernetes.slack.com

uberspot (approved):

ramaro (approved):

ramaro:
Thanks @graillus! LGTM :D

Issue #776, opened 2021-08-13
jinja2 inventory_path missing


sebradloff:
**Describe the bug/feature**
I believe jinja2 templating might have issues with inventory similar to this issue with Kadet a few months ago: https://github.com/kapicorp/kapitan/pull/664

**To Reproduce**
I'm not sure how to reproduce at this time

**Expected behavior**
Inventory path to not be an issue. Last worked for me on `v0.29.2`

**Screenshots**
```
root@8ed7e8a1e9cf:/src# kapitan c -t box-0
Unknown (Non-Kapitan) Error occurred
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/opt/venv/lib/python3.7/site-packages/kapitan/targets.py", line 464, in compile_target
    input_compiler.compile_obj(comp_obj, ext_vars, **kwargs)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/inputs/base.py", line 55, in compile_obj
    self.compile_input_path(input_path, comp_obj, ext_vars, **kwargs)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/inputs/base.py", line 77, in compile_input_path
    **kwargs,
  File "/opt/venv/lib/python3.7/site-packages/kapitan/inputs/jinja2.py", line 39, in compile_file
    context["inventory"] = inventory(self.search_paths, target_name)
  File "/opt/venv/lib/python3.7/site-packages/kapitan/resources.py", line 264, in inventory
    inventory_path = inventory_path_arg.inventory_path
AttributeError: 'NoneType' object has no attribute 'inventory_path'
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/venv/lib/python3.7/site-packages/kapitan/targets.py", line 136, in compile_targets
    [p.get() for p in pool.imap_unordered(worker, target_objs) if p]
  File "/opt/venv/lib/python3.7/site-packages/kapitan/targets.py", line 136, in <listcomp>
    [p.get() for p in pool.imap_unordered(worker, target_objs) if p]
  File "/usr/local/lib/python3.7/multiprocessing/pool.py", line 748, in next
    raise value
AttributeError: 'NoneType' object has no attribute 'inventory_path'


'NoneType' object has no attribute 'inventory_path'
```

[...]

Issue #737, opened 2021-05-26
Kapitan 0.29.3 not working with Python 3.9.5


xccatana:
**Describe the bug/feature**
Kapitan 0.29.3 doesn't work with Python 3.9.5. but it works with Python 3.8.5

**To Reproduce**
1. Create virtualenv and source to it
2. pip install kapitan==0.29.3
3. kapitan compile
```
(kapitan-dev) ccatana monitoring (master) >> kapitan compile
Unknown (Non-Kapitan) Error occurred
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/Users/ccatana/Documents/workspace/git_corp_com/monitoring/kapitan-dev/lib/python3.9/site-packages/kapitan/targets.py", line 453, in compile_target
    input_compiler.compile_obj(comp_obj, ext_vars, **kwargs)
  File "/Users/ccatana/Documents/workspace/git_corp_com/monitoring/kapitan-dev/lib/python3.9/site-packages/kapitan/inputs/base.py", line 54, in compile_obj
    self.compile_input_path(input_path, comp_obj, ext_vars, **kwargs)
  File "/Users/ccatana/Documents/workspace/git_corp_com/monitoring/kapitan-dev/lib/python3.9/site-packages/kapitan/inputs/base.py", line 69, in compile_input_path
    self.compile_file(
  File "/Users/ccatana/Documents/workspace/git_corp_com/monitoring/kapitan-dev/lib/python3.9/site-packages/kapitan/inputs/jinja2.py", line 35, in compile_file
    context["inventory"] = inventory(self.search_paths, target_name)
  File "/Users/ccatana/Documents/workspace/git_corp_com/monitoring/kapitan-dev/lib/python3.9/site-packages/kapitan/resources.py", line 260, in inventory
    inventory_path = inventory_path_arg.inventory_path
AttributeError: 'NoneType' object has no attribute 'inventory_path'
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/ccatana/Documents/workspace/git_corp_com/monitoring/kapitan-dev/lib/python3.9/site-packages/kapitan/targets.py", line 126, in compile_targets
[...]

PR #837, opened 2022-05-08
fix cli alias handling (#776, #737)

dhorbach:
Fixes issue #776, #737

## Proposed Changes

  - Caching of args by key should use full command name and not alias
  - Inventory alias is duplicated with init alias - changed "i" to "inv"

### DIFF
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -96,22 +96,14 @@ def trigger_compile(args):
     )
 
 
-def main():
-    """main function for command line usage"""
-    try:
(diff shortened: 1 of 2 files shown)


dhorbach:
@ramaro could you please review this PR?

ramaro:
@dhorbach apologies, missed this! Having a look.

ramaro (approved):

ramaro:
@dhorbach this is great, thank you! Had to review the associated issues and think about it once again, but this makes sense 😃

PR #842, opened 2022-07-01
Conditional dependency download and compile

megian:
## Proposed Changes

Depending on features turned on or off, this allows to individual
download or skip certain dependencies and compile them just if they
turned on.

The default behavior of download and compile everything doesn't
change with it. Existing use cases must not have any behavior changes.

### DIFF
--- a/kapitan/dependency_manager/base.py
+++ b/kapitan/dependency_manager/base.py
@@ -48,6 +48,12 @@ def fetch_dependencies(output_path, target_objs, save_dir, force, pool):
         try:
             dependencies = target_obj["dependencies"]
             for item in dependencies:
+                if not item.get("condition", True):
+                    logger.warning(
+                        "Skipping dependency download for uri %s because condition is false", item["source"]
+                    )
+                    continue
+
                 dependency_type = item["type"]
                 source_uri = item["source"]
 
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -472,6 +472,12 @@ def compile_target(target_obj, search_paths, compile_path, ref_controller, globa
         input_type = comp_obj["input_type"]
         output_path = comp_obj["output_path"]
 
+        if not comp_obj.get("condition", True):
+            logger.warning(
+                "Skipping compile for type %s and output_path %s because condition is false", input_type, output_path
+            )
+            continue
+
         if input_type == "jinja2":
             input_compiler = Jinja2(compile_path, search_paths, ref_controller)
             if "input_params" in comp_obj:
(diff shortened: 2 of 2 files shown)


megian:
Documentation and tests are pending.

simu:
The broader context here: Conditional dependency fetching and compile steps would be very helpful for us in Project Syn, e.g. for stuff like https://github.com/projectsyn/component-keycloak/pull/135/commits/95d3bc6029bfc7f2bf518ec8ea07ba5dbd0347ee where we now have a postprocessing step to delete the rendered Postgres helm chart if Keycloak is configured to use an external database. @ramaro @ademariag what do you think?

ramaro:
@megian @simu how would this valued be managed? Would it be an external process (e.g. projectsyn) updating the value of `condition`? Also thoughts on renaming `condition` into `skip` ? (I find in it more obvious)

megian:
Initially inspired by the helm chart dependency [condition](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags) flag `condition: somechart.enabled`. As Simon @simu already mentioned, this would result in a inverse logic.

In a real example with `condition`:
```
=_enable_pg_chart:
  builtin: true         # PostgreSQL Helm Chart has to be rendered
  external: false    # PostgreSQL Helm Chart has not to be rendered

parameters:
  kapitan:
    dependencies:
      - type: helm
        condition: ${keycloak:_enable_pg_chart:${keycloak:database:provider}}
```
Compared to `skip`:
```
=_skip_pg_chart:
  builtin: false      # PostgreSQL Helm Chart has to be rendered
  external: true    # PostgreSQL Helm Chart has not to be rendered

parameters:
  kapitan:
    dependencies:
      - type: helm
        skip: ${keycloak:_skip_pg_chart:${keycloak:database:provider}}
```
For my understanding `_enable_pg_chart` condition on true makes it obvious it is executed (Feature enabled, feature is downloaded and compiled). `_skip_pg_chart` causes me questions while reading the code, when should I skip what. This will even become more complex, when multiple features have to be skipped in a inverse logic.

Summarized: As set skip manually to `yes` is not really a use case, except of debugging, which we can ignore here I think. The main use case is to switch features on or off. Usually it is turned on by default for the reason as it is there (if it wouldn't, it would got dropped). If the variable is called `condition` or `enabled` doesn't matter. I still think that from the main usage perspective a positive logic makes sense.

kapitan/dependency_manager/base.py
```diff
@@ -48,6 +48,12 @@ def fetch_dependencies(output_path, target_objs, save_dir, force, pool):
         try:
             dependencies = target_obj["dependencies"]
             for item in dependencies:
+                if not item.get("condition", True):
+                    logger.warning(
```
MatteoVoges, line 52: please change to `debug`, to make it less verbose

MatteoVoges (changes requested):
Do you want to only be able to skip dependencies or do you want this behavior for entire compile definitions as well?

Issue #843, opened 2022-07-06
[Helm] invalid Yaml document separator: null


Moep90:
# Describe the bug/feature
When rendering the `kube-prometheus-stack` chart, I get the following issue while applying the manifest:

invalid Yaml document separator: null

# To Reproduce

1. Define the chart with the values below
2. `./kapitan compile --fetch`
3. `kubectl apply --recursive -f compiled/..../kube-prometheus-stack.yml`

## Values
```yaml
parameters:
  kube-prometheus-stack:
    application_version: 0.57.0
    chart_name: kube-prometheus-stack
    chart_version: 36.2.1
    fullname: kube-prometheus-stack
    helm_params:
      name: kube-prometheus-stack
    helm_values:
      alertmanager:
        enabled: false
        podAntiAffinityTopologyKey: kubernetes.io/hostname
        securityContext:
          fsGroup: 2000
          runAsGroup: 2000
          runAsNonRoot: true
          runAsUser: 1000
      coreDns:
        enabled: true
      defaultRules:
        create: false
        rules:
          alertmanager: true
          configReloaders: true
          etcd: true
          general: true
          k8s: true
          kubeApiserverAvailability: true
          kubeApiserverBurnrate: true
          kubeApiserverHistogram: true
          kubeApiserverSlos: true
          kubePrometheusGeneral: true
          kubePrometheusNodeRecording: true
          kubeProxy: true
          kubeScheduler: true
          kubeStateMetrics: true
          kubelet: true
          kubernetesApps: true
          kubernetesResources: true
          kubernetesStorage: true
          kubernetesSystem: true
          network: true
          node: true
          nodeExporterAlerting: true
          nodeExporterRecording: true
          prometheus: true
          prometheusOperator: true
      global:
        rbac:
          create: true
          pspEnabled: true
      grafana:
        enabled: false
        serviceMonitor:
          interval: 30s
          scheme: http
          scrapeTimeout: 30s
          tlsConfig: {}
      kubeApiServer:
        enabled: true
        tlsConfig:
          insecureSkipVerify: false
          serverName: kubernetes
      kubeControllerManager:
        enabled: true
      kubeEtcd:
        enabled: true
      kubeProxy:
        enabled: true
      kubeScheduler:
        enabled: true
      kubelet:
        enabled: true
        namespace: kube-system
      nameOverride: kube-prometheus-stack
      namespaceOverride: monitoring
      prometheus:
        enabled: true
        prometheusSpec:
[...]

PR #844, opened 2022-07-06
Feature/fix yaml document seperator null

Moep90:
Fixes issue #843

## Proposed Changes

  - Introduced SafeDumper as proposed [here](https://stackoverflow.com/questions/30134110/how-can-i-output-blank-value-in-python-yaml-file)
  - https://pyyaml.docsforge.com/master/api/yaml/representer/


ramaro:
Hey @Moep90 sorry for the delay. 

I see the intention here, but I personally make use of `null` in YAML when setting values to `None` in Kadet and honestly think that's the correct, expected behavior.

However I understand why this could be useful, so maybe we can add this as a feature behind a flag?

MatteoVoges:
Hey @ramaro,

I'm working on this feature for the last hours and it seems like there is a problem with the yaml-representer.
I introduced the feature as a flag but the representer is loaded before the program starts  and the flags/ arguments are read just after that.

So replacing the none-type-representer at runtime won't work. I have thought of some constructor overloading for the PrettyDumper class, so that it can be instantiated with different parameters for different representers etc.

What do you suggest?

Moep90:
fixed with #861

PR #861, opened 2022-10-04
Feature: Cli-flags for controlling the dumping behavior

MatteoVoges:
**Fixed Issues**
* https://github.com/kapicorp/kapitan/issues/852
* https://github.com/kapicorp/kapitan/issues/843
* https://github.com/kapicorp/kapitan/issues/859

**Added two new CLI-flags:**

1. `--multiline-string-style [STYLE]`
* Set representation of multiline string to given style (default: `"`)
* Available options:
  * `literal`: Using `|` as representation (turns every newline within the string into a literal newline)
  * `folded`: Using `>` as representation (removes single newlines within the string)
  * `double-quotes`: Using `"` as representation (dumps newlines as `\n`-character in one string)
  * Find out more about yaml-strings [here](https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines)

2. `--dump-null-as-empty`
* Set representation of none-type-entries to ` ` (default: `null`)

The default behavior of  the PrettyDumper does not change.

### DIFF
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -271,19 +271,20 @@ def build_parser():
     )
 
     compile_parser.add_argument(
-        "--multiline-string-rep",
+        "--multiline-string-style",
         type=str,
-        choices=('|', '>', '"'),
-        metavar="ARG",
-        default=from_dot_kapitan("compile", "multiline-string-rep", '"'),
-        help="set multiline string style to '|'",
+        choices=["literal", "folded", "double-quotes"],
+        metavar="STYLE",
+        action="store",
+        default=from_dot_kapitan("compile", "multiline-string-style", False),
+        help="set multiline string style to STYLE, default is 'double-quotes'",
     )
     
     compile_parser.add_argument(
-        "--dump-null",
-        default=from_dot_kapitan("compile", "dump-null", False),
+        "--dump-null-as-empty",
+        default=from_dot_kapitan("compile", "dump-null-as-empty", False),
         action="store_true",
-        help="dumps all none-type entries"
+        help="dumps all none-type entries as empty"
     )
 
     compile_selector_parser = compile_parser.add_mutually_exclusive_group()
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -279,12 +279,12 @@ def build_parser():
         default=from_dot_kapitan("compile", "multiline-string-style", "double-quotes"),
         help="set multiline string style to STYLE, default is 'double-quotes'",
     )
-    
+
     compile_parser.add_argument(
         "--dump-null-as-empty",
         default=from_dot_kapitan("compile", "dump-null-as-empty", False),
         action="store_true",
-        help="dumps all none-type entries as empty"
+        help="dumps all none-type entries as empty",
     )
 
     compile_selector_parser = compile_parser.add_mutually_exclusive_group()
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -271,20 +271,20 @@ def build_parser():
     )
 
     compile_parser.add_argument(
-        "--multiline-string-style",
+        "--yaml-multiline-string-style",
         type=str,
         choices=["literal", "folded", "double-quotes"],
         metavar="STYLE",
         action="store",
-        default=from_dot_kapitan("compile", "multiline-string-style", "double-quotes"),
+        default=from_dot_kapitan("compile", "yaml-multiline-string-style", "double-quotes"),
         help="set multiline string style to STYLE, default is 'double-quotes'",
     )
 
     compile_parser.add_argument(
-        "--dump-null-as-empty",
-        default=from_dot_kapitan("compile", "dump-null-as-empty", False),
+        "--yaml-dump-null-as-empty",
+        default=from_dot_kapitan("compile", "yaml-dump-null-as-empty", False),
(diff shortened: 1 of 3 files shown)


kapitan/cli.py
```diff
@@ -270,6 +270,23 @@ def build_parser():
         help='set schema cache path, default is "./schemas"',
     )
 
+    compile_parser.add_argument(
+        "--multiline-string-style",
```
ramaro, line 274: can we rename this to `--yaml-multiline-string-style`?

kapitan/cli.py
```diff
@@ -270,6 +270,23 @@ def build_parser():
...
+        type=str,
+        choices=["literal", "folded", "double-quotes"],
+        metavar="STYLE",
+        action="store",
+        default=from_dot_kapitan("compile", "multiline-string-style", "double-quotes"),
+        help="set multiline string style to STYLE, default is 'double-quotes'",
+    )
+
+    compile_parser.add_argument(
+        "--dump-null-as-empty",
```
ramaro, line 284: also this to `--yaml-dump-null-as-emtpy`?

kapitan/utils.py
```diff
@@ -212,6 +212,54 @@ def increase_indent(self, flow=False, indentless=False):
...
+    Ref: https://github.com/yaml/pyyaml/issues/240#issuecomment-1018712495
+    """
+    # get parsed args from cached.py
+    try:
+        cached_dict = cached.as_dict()
+        args = vars(cached_dict["args"]["compile"])
+
+        style = args["multiline_string_style"]
+    except KeyError:
+        style = None
```
ramaro, line 228: you should be able to use cached instead like this:

```
compile_args = cached.args
style = args.get("multiline_string_style", None)
```

kapitan/utils.py
```diff
@@ -212,6 +212,54 @@ def increase_indent(self, flow=False, indentless=False):
...
+    return dumper.represent_scalar("tag:yaml.org,2002:str", data)
+
+
+PrettyDumper.add_representer(str, multiline_str_presenter)
+
+
+def null_presenter(dumper, data):
+    """Configures yaml for omitting value from null-datatype"""
+    # get parsed args from cached.py
+    try:
```
ramaro, line 247: same cached tip here

kapitan/utils.py
MatteoVoges: Hi @ramaro,
I don't understand how the args in cached.py are stored and especially why so.
Your suggestion doesn't work, because the `compile_args`-variable is just a dictionary with a single `compile`-key. So the wanted arg-key cannot get found and the default-value gets taken. 
Am I doing something wrong or not seeing anything?

If we print the `compile_args`-variable and the received value for style we get:
Running with `$ kapitan compile --multiline-string-style literal`
```
{'compile': Namespace(subparser_name='compile', [...], multiline_string_style='literal', [...], name='compile')}
Received value for style: None
```
The mentioned `compile`-key has a Namespace object as its value. You can resolve this object by using `vars()` to get the actual dictionary with the args and their values.

I would suggest something like this: 
```python
compile_args = vars(cached.args.get("compile"))
style = compile_args.get("multiline_string_style", None)
```  
This works fine and is a solution without using `try-catches`.  
Again, please let me know, if I'm doing something not as intended.

kapitan/cli.py
MatteoVoges: Of course!

kapitan/cli.py
MatteoVoges: same here

kapitan/utils.py
ramaro: Hmm sorry, my bad for commenting late at night. I meant to suggest that the following instead will grab either the default or flag value in argparse without any conversions

```
style = cached.args["compile"].multiline_string_style
if style bla:
  foobar
```

MatteoVoges: **Update:**
The unittest `test_cli_ref_reveal_recursive_dir` in `test_cli.py` fails with my suggestion. I had this issue before and used `try-except` to make this test not fail.

I know, that if a test fails, we shouldn't wrap everything in a `try-except`-block to make it work.

Maybe the test isn't working correctly, because it throws
```
File "/home/matteo/kapitan_nexenio/kapitan/utils.py", line 222, in multiline_str_presenter
    compile_args = vars(cached.args.get("compile"))
TypeError: vars() argument must have __dict__ attribute
```
This is weird, because in manual use the error doesn't come up.

What are your ideas on that?

ramaro: Ah yes, good point. This is not a big deal though! Both your code and the test are working as expected.
The `test_cli_ref_reveal_recursive_dir` is testing the cli, therefore command line arguments are hardcoded within the test via sys.argv.

So I think the safest way is to adapt your code to check if `compile` is passed (which is always then when not running kapitan from a test). Something like this should work:

```py
compile = cached.args.get("compile", None)
style = None
if compile:
  style = cached.args["compile"].multiline_string_style
if style bla:
  foobar
```

MatteoVoges: This is a good idea. Only change: I renamed `compile` to `compile_args`.

ramaro: thanks!

ramaro:
LGTM! Thanks for this, merging.

Issue #858, opened 2022-09-19
Support force fetch per compile target


Moep90:
Support the ability to force fetch selected external dependencies.

The documentation says: "_If you simply want the latest chart available, either don't include the version key or specify an empty string._"

But to have the latest to compile, I need to always run `kapitan compile --fetch --force`.
This applies to all external dependencies.

## Helm
This would fetch the latest version of the chart if there is an update.
Make sure to run helm `helm repo update` prior to fetching the latest chart.
```yaml
  kapitan:
    dependencies:
    - type: helm
[...]
      version: "" <----- to fetch always latest
      fetch_always: true <--- force download this dependency always
```

## Git
This would apply if we update the master between kapitan runs.

```yaml
- type: git output_path: source/kapitan
  source: git@github.com:kapicorp/kapitan.git
  subdir: kapitan
  ref: master
  fetch_always: true <--- force download this dependency always
```
## HTTP
```yaml
- type: http | https
  output_path: path/to/file
  source: http[s]://<url>
  unpack: True | False
  fetch_always: true <--- force download this dependency always
```

## Full example with helm
```yaml
---
parameters:
  filebeat:
    chart_name: filebeat
    fullname: filebeat
    chart_version: "7.17.3"
    application_version: ${filebeat:chart_version}
    namespace: ${namespace}
  kapitan:
    dependencies:
    - type: helm
      output_path: components/charts/${filebeat:chart_name}/${filebeat:chart_version}/${filebeat:application_version}
      source: https://helm.elastic.co
      version: "" <----- to fetch always latest
      fetch_always: true <--- force download this dependency always
      chart_name: ${filebeat:chart_name}
    compile:
    - input_type: helm
      input_paths:
        - components/charts/${filebeat:chart_name}/${filebeat:chart_version}/${filebeat:application_version}
      output_path: manifests
      helm_params:
        namespace: ${filebeat:namespace}
        name: ${filebeat:chart_name}
        output_file: ${filebeat:chart_name}.yml
      helm_values: ${filebeat:helm_values}
```

### Example compile run
```
$ kapitan compile --fetch
Not updating lib/terraform/__init__.py (file already exists)
[...]
Dependency helm chart filebeat and version 7.17.3: fetching now
Dependency helm chart filebeat and version 7.17.3: successfully fetched
```

Moep90:
closed by #862

PR #862, opened 2022-10-04
Feature: Added force-fetch functionality (#858)

MatteoVoges:
Fixes issue: #858 
## Proposed Changes
* Added a option to have a `fetch_always`-property in your external dependencies.
    If `fetch_always` is set to `true`, this dependency will always get fetched and overwrite existing files (force-fetch).

* Renamed `--force`-flag to `--force-fetch`, so you have to run
    * ` $ kapitan compile --fetch` to fetch (existing files won't get overwritten) or
    * ` $ kapitan compile --force-fetch` to fetch the dependencies and overwrite existing files.

If the property is false or not given, the default-behavior of the fetch-functionality doesn't change.
Also the behavior of `--fetch` and (now:)`--force-fetch` doesn't change either.

**NOTE:** Using `--force` will still work with its default behavior, but is now a _deprecated_ flag and will be removed soon.

## Documentation and testing
Updated flag-name in the docs and in the tests.

### DIFF
--- a/docs/kap_proposals/kap_1_external_dependencies.md
+++ b/docs/kap_proposals/kap_1_external_dependencies.md
@@ -27,6 +27,7 @@ Git type may also include `ref` and `subdir` parameters as illustrated below:
   source: <git_url>
   subdir: relative/path/in/repository
   ref: <commit_hash/branch/tag>
+  fetch_always: <bool>
 ```
 
 If the file already exists at `output_path`, the fetch will be skipped. For fresh fetch of the dependencies, users may add `--fetch` option as follows:
@@ -35,7 +36,7 @@ If the file already exists at `output_path`, the fetch will be skipped. For fres
 $ kapitan compile --fetch
 ```
 
-Users can also add the `fetch_always: true` option to the `kapitan.compile` in the inventory in order to force fresh fetch of the dependencies every time.
+Users can also add the `fetch_always: true` option to the `kapitan.dependencies` in the inventory in order to force fetch of the dependencies of the target every time.
 
 ## Implementation details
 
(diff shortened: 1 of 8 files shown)


MatteoVoges:
Hi @ramaro @ademariag ,

I applied the changes we discussed, so the PR is ready to review.

But there is one more thing I noticed:
I don't like the way the code for the fetch-functionality is structured. There are multiple if-statements checking the same, and its hard to follow and unterstand the code in general. I would suggest something like this:
```python
[...]
# getting the flag values
if fetch:
    # fetch inventories
    # fetch dependencies
else:
    # fetch targets with fetch_always tag
[...]
```
So with that we would have one if-statement, which indicates if we are fetching or not.

Should I dump my suggestion into this PR or wait until this PR is merged and then create a new one?

ramaro:
@MatteoVoges thanks for getting this done! 

>Should I dump my suggestion into this PR or wait until this PR is merged and then create a new one?
I think doing this in a new PR makes sense to keep things simpler!

ramaro:
This lgtm @MatteoVoges ! Would you mind updating the doc changes and title to force_fetch instead of fetch_always? We can merge straight after that,

ramaro:
@MatteoVoges there's some `fetch_always` usage in docs and code. Particularly because during our discussion, we've agreed to rename to `force_fetch` and this is not reflected for example in `valid_target_obj()`. Unless I'm missing something here 🤔

MatteoVoges:
@ramaro , you're right. I updated it now. Please let me know if there is still something to do here.

docs/kap_proposals/kap_1_external_dependencies.md
```diff
@@ -27,6 +27,7 @@ Git type may also include `ref` and `subdir` parameters as illustrated below:
   source: <git_url>
   subdir: relative/path/in/repository
   ref: <commit_hash/branch/tag>
+  fetch_always: <bool>
```
ramaro, line 30: @MatteoVoges last update here needed

MatteoVoges: @ramaro ahh I missed this one. Done with 5b632c26316309f51764a369284d3edb4415867e

ramaro: now that was quick! Thanks 👍🏼

PR #863, opened 2022-10-11
Substitute `${compiled_target_dir}` in env vars for `input_type: external`

simu:
## Summary

Kapitan already substitutes `${compiled_target_dir}` in the command line arguments for the command executed for compile objects with `input_type: external`.

This commit adds logic to also substitue `${compiled_target_dir}` in the environment variables provided to the external command. This allows users to specify the compilation target directory in an environment variable, which can be useful for example to simplify wrapper scripts which do some work before calling the external tool.

See https://github.com/projectsyn/commodore/pull/646 for an example PR which implements a wrapper script for `kustomize`. This wrapper script could benefit greatly from being able to pass the output directory as an environment variable as it would remove the hard requirement of the wrapper script that users provide the output path as the first entry in `args`.

## Proposed Changes

  - Substitute `${compiled_target_dir}` in all environment variables passed in compile objects with `input_type: external`.

### DIFF
--- a/kapitan/inputs/external.py
+++ b/kapitan/inputs/external.py
@@ -45,13 +45,16 @@ class External(InputType):
             args = " ".join(args)
 
             # compile_path (str): Path to current target compiled directory
-            args = re.sub(r"(\${compiled_target_dir})", compile_path, args)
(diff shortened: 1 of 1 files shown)


ramaro:
Looks good @simu . Thanks!

PR #864, opened 2022-10-11
Propagate env vars `$PATH` and `$HOME` for `input_type: external`

simu:
## Summary

For compile objects with `input_type: external`, the external command is executed with `subprocess.run()` with keyword parameter `env`. By providing this keyword parameter, the command will not inherit Kapitan's `os.environ`, as documented in
https://docs.python.org/3/library/subprocess.html:

> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment. It is passed directly to Popen.

This commit ensures that we pass `$PATH` and `$HOME` from Kapitan's `os.environ` to the external command unless those environment variables were explicitly set in the compile object's `env_vars`.

## Proposed Changes

  - Propagate environment variables `$PATH` and `$HOME` to external command for `input_type: external` unless explicitly set in the compile object's `env_vars`

### DIFF
--- a/docs/compile.md
+++ b/docs/compile.md
@@ -564,3 +564,10 @@ parameters:
 ```
 
 *Supported output types*: N/A (no need to specify `output_type`)
+
+Additionally, the input type supports field `env_vars`, which can be used to set environment variables for the external command.
+By default, the external command doesn't inherit any environment variables from Kapitan's environment.
+However, if environment variables `$PATH` or `$HOME` aren't set in `env_vars`, they will be propagated from Kapitan's environment to the external command's environment.
+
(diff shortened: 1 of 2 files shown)


Moep90:
@simu would you please mind to change the docs as well?
See: https://github.com/kapicorp/kapitan/blob/master/docs/compile.md#external-alphaexperimental

ramaro:
@simu agree with @Moep90 ! Once we get the doc updated, this is good to merge 👍🏼

ramaro:
Perfect, thank you @simu ! Merging.

Issue #694, opened 2021-02-10
[Feature] Please add more automatic generators


Moep90:
# Describe the feature
Currently only few [automatically](https://kapitan.dev/secrets/#automatically) generators are available.
## Automatically generate random integer
I can think of a `randomint` generator like the `randomstr`
In my case I'd like to generate random high userIDs for the security context.
For example:: `runAsUser: 1000620000 fsGroup: 1000620000`
Example:
```yaml
securityContext:
  runAsUser: ?{base64:targets/${target_name}/runAsUser||randomint:12}
```

## Ability to in/exclude special characters and/or 
If a tool only allows certain special characters I'd like to specify them.
I can think of something for example regex
Example:
```yaml
database_credentials:
  mysql_password_certain_special_chars: ?{base64:targets/${target_name}/runAsUser||randomstr("^[a-zA-Z0-9 ,:/\\-]*$"):6}
```

MatteoVoges:
Actually, the issue is resolved with , but there is still a tiny problem:

The tags 
* `?{ref:path/to/secret||alphanumspec:16: }` - Error: whitespace as argument for special_chars (expected Non-Whitespace-Character)
* `?{ref:path/to/secret|randomstr}` - Error: using single `|` instead of `||`

get dumped in the output file and no secret was created.

I wan't at least that the error gets detected and a RefError gets raised, so that the user knows what he did wrong.

The new regex would detect both of it, but - i think - doesn't get evaluated.

Where do you think is the right place to check the tag for syntax first?

Moep90:
closed by #866

PR #866, opened 2022-10-13
Added more automatic generator functions (#694)

MatteoVoges:
Fixes issue #694

## Proposed Changes
**New generic `random` function:**
Generates a random string based on given parameters.
### Usage
```yaml
secret: ?{ref:path/to/secret||random:type:nchars:special}
```
The first parameter `type` has to be one of:
* `str` : generator function for alphanumeric characters, will be url-token-safe
* `int` : generator function for digits (0-9)
* `loweralpha` : generator function for lowercase letters (a-z)
* `upperalpha` : generator function for uppercase letters (A-Z)
* `loweralphanum` : generator function for lowercase letters and numbers (a-z and 0-9)
* `upperalphanum` : generator function for uppercase letters and numbers (A-Z and 0-9)
* `special` : generator function for alphanumeric characters and given special characters

I'm always open for name suggestions.

All these types have in common, that the second parameter `nchars` indicates how many characters the generated secret contains.
The default value for this is
  * `random:str` : 43 characters
  * `random:int` : 16 characters
  * other : 8 characters

The `special` type has a third argument that indicates which special characters are allowed for the generation.
The default for that is `string.punctuation` (``!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~``).

**NOTE:** in manual usage the characters `|` and `:` are used for function and argument parsing, so they can't be specified.

This might be a 
## Breaking change
To pass special characters into the generator pool, we have to change the **REF_TOKEN_TAG_PATTERN**-regex.
The new regex `(\?{(\w+:[\w\-\.\@\=\/\:]+)(\|(?:(?:\|\w+)(?::\S*)*)+)?\=*})` matches all existing use cases, but is more restricted.
So be beware of unexpected errors.

Furthermore the existing generator functions `randomstr` and `lowernumalpha` are still working and still behave the same.
But they will be removed soon.

## Examples
Here are some example cases with expected output:
* Tag: `?{ref:path/to/secret||random:int}` generates a secret containing 16 (default) digits, i.e. `1254872907909324`
* Tag: `?{ref:path/to/secret||random:upperalphanum:12}` generates a secret containing 12 chars from uppercase letters and numbers, i.e. `VUPA0GIO47C2`
* Tag: `?{ref:path/to/secret||random:special:16:!$%&}` generates a secret containing 16 chars from alphanumeric letters and given special characters `!$%&`, i.e. `!kQT&m%csSDsdaBy`

You could leave the argument empty, so `?{ref:path/to/secret||random:special:12:}` generates a alphanumeric string without any special characters.

## Documentation and testing
* Updated Tests
* **TBD:** Update docs for latest iteration.

## Minor Issue
tl;dr: Some errors don't get caught and could produce some unexpected behavior.
See detailed: https://github.com/kapicorp/kapitan/issues/694#issuecomment-1277919468

### DIFF
--- a/kapitan/refs/functions.py
+++ b/kapitan/refs/functions.py
@@ -10,7 +10,6 @@ import hashlib
 import logging
 import secrets  # python secrets module
 import string
-import re
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
@@ -192,15 +191,15 @@ def alphanumspec(ctx, nchars="8", special_chars=string.punctuation):
     default is !"#$%&'()*+,-./:;<=>?@[\]^_`{|}~
     NOTE: '|' and ':' are used for arg parsing and aren't allowed manually, in default they work
     """
-    # make sure that each character is include only once or not at all
+    # make sure that each allowed character is include only once or not at all
     special_chars = "".join(set(special_chars).intersection(string.punctuation))
-    pool = string.ascii_letters + special_chars
+    pool = string.ascii_letters + string.digits + special_chars
     generic_alphanum(ctx, nchars, pool)
 
 
 def generic_alphanum(ctx, nchars, pool):
     """
-    generates a DNS-compliant text string, containing nchars from pool
+    generates a text string, containing nchars from pool
     default for nchars is 8 chars
     sets it to ctx.data
     """
@@ -209,6 +208,10 @@ def generic_alphanum(ctx, nchars, pool):
         nchars = int(nchars)
     except ValueError:
         raise RefError(f"Ref error: eval_func: {nchars} cannot be converted into integer.")
+
+    allowed_pool = string.ascii_letters + string.digits + string.punctuation
+    if not set(pool).issubset(allowed_pool):
+        raise RefError("{}: invalid elements in pool".format(set(pool).difference(set(allowed_pool))))
 
     # generate string based on given pool
     generated_str = "".join(secrets.choice(pool) for i in range(nchars))
--- a/kapitan/refs/functions.py
+++ b/kapitan/refs/functions.py
@@ -47,7 +47,7 @@ def randomstr(ctx, nbytes=""):
     sets it to ctx.data
     """
     # deprecated function
-    logger.info("")
+    logger.info("DeprecationWarning: randomstr is deprecated. Use random:str instead")
     random(ctx, "str", nbytes)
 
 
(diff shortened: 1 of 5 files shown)


kapitan/refs/functions.py
```diff
@@ -20,20 +20,30 @@
...
+        "loweralphanum": lower_alpha_num,
+        "upperalpha": upper_alpha,
+        "upperalphanum": upper_alpha_num,
+        "alphanumspec": alphanumspec,
     }
 
-    return func_lookup[func_name](ctx, *func_params)
-
 
 def randomstr(ctx, nbytes=""):
```
MatteoVoges, line 48: I could update `randomstr` to a more generic solution as well.
It would look like:
```python
def randomstr(ctx, nchars="43"):
    pool = string.ascii_letters + string.digits + "-_"
    generic_alphanum(ctx, nchars, pool)
```
Also we could use `alphanumspec` for that:
```python
def randomstr(ctx, nchars="43"):
    alphanumspec(ctx, nchars, "-_")
```

tests/test_refs.py
```diff
@@ -440,3 +441,113 @@ def test_ref_function_loweralphanum(self):
...
+        # Test with every allowed special char, nchars=32
+        allowed_special_chars = set(string.punctuation).difference(":|")
+        for special_char in allowed_special_chars:
+            tag = "?{plain:ref/alphanumspec||alphanumspec:32:" + special_char + "}"
+            REF_CONTROLLER[tag] = RefParams()
+            REVEALER._reveal_tag_without_subvar.cache_clear()
+            revealed = REVEALER.reveal_raw_file(file_with_tags)
+            self.assertEqual(len(revealed), 32)
+            intersection = set(string.punctuation).intersection(revealed)
+            self.assertTrue(intersection.issubset(set(special_char)))
```
MatteoVoges, line 553: I think the for loop is unneccessary here. We could test it directly with `allowed_special_chars` as one string as input.

ademariag:
suggestion from Kapitan Workgroup:

introduce one new function `random` with the following parameters

`random:<class>:<length>`

where class is one of:

`str`: same as currently using `randomstr`
`int` : generator function for digits (0-9)
`loweralpha` : generator function for lowercase letters (a-z)
`upperalpha` : generator function for uppercase letters (A-Z)
`upperalphanum` : generator function for uppercase letters and numbers (A-Z and 0-9)
`alphanumspec` : generator function for alphanumeric characters and given special characters

i.e. 
`random:str` produces same results as current `randomstr`
`random:int` produces a random int
`random:loweralpha:15` produces a 15 character `loweralpha` string

kapitan/refs/functions.py
MatteoVoges: Fixed with latest iteration

tests/test_refs.py
MatteoVoges: Fixed with latest iteration

MatteoVoges:
Do you think in `test_refs.py` every single option for `type` should be checked? Or is it enough to take some samples?

Issue #867, opened 2022-10-14
Add Markdown linter to GitHub Actions


ramaro:
Add Markdown linter to GitHub Actions

ademariag:
Resolved by https://github.com/kapicorp/kapitan/pull/872

PR #872, opened 2022-10-28
Add markdown linter to pre-commit

MatteoVoges:
Fixes issue #867 

## Proposed Changes
  - Added markdown linter from [markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli) to pre-commit
  - Added `.markdownlintignore` file : specifies which directories/ files get ignored
  - Added `.markdownlint.yml` file : configuration for markdown linter
  - Applied markdown rules automatically (`markdownlint-fix`) and manually
 
### Explanation of non-default rules in `markdownlint.yml`
- `MD013` sets the maximum line length by default to 80 characters. In the docs are some lines containing up to 550 characters, so I set the maximum to 600 characters. Fixing these long lines down to 80 characters, will be a change for the future.
- `MD024` won't allow the same header name in one file. Thus we have several headers like `Usage`, `Breaking`, `Updates`, `Examples` and so on, I deactivated the rule.
- `MD033` won't allow `html`-tags in the markdown files. To allow the `img` tag in `README.md`, we have to add it to the parameter `allowed_elements`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue_template.md
+++ b/.github/ISSUE_TEMPLATE/issue_template.md
@@ -12,6 +12,7 @@ A clear and concise description of what the bug/feature request is.
 
 **To Reproduce**
 Steps to reproduce the behavior:
+
 1. Go to '...'
 2. Click on '....'
 3. Scroll down to '....'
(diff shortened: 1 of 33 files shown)


ramaro:
This is awesome. thanks @MatteoVoges !

ramaro:
Merging!

PR #873, opened 2022-10-28
Add basic auth generator function

MatteoVoges:
## Proposed Changes
  - Added a generator function for basic auth secrets

### Usage
The generator function has two optional arguments: `?{ref:path/to/secret||basicauth:username:password}`
If the arguments are not specified, a text-string containing 8 characters will be generated.

### Examples
- `||basicauth` could generate something like `dddkfrgl:DDwlFByq` as username:password - token
**NOTE:** The secret itself is base64 encoded so it will look like `ZGRka2ZyZ2w6RER3bEZCeXE=`
- `|basicauth:user123:key123` would exactly generate `user123:key123` as token, so the secret (base64) would be `dXNlcjEyMzprZXkxMjM=`

### Testing and docs
Added test for basic auth generator function.
Updated `secret.md` as well.


ramaro:
lgtm @MatteoVoges ! Can you please rebase before we can merge?

ramaro (approved):

ramaro:
merging! 👍🏼

PR #874, opened 2022-10-28
update jinja yaml filter to enfoce the string type

arielly-parussulo:
Fix
There is no issue created for this fix. We figured it out in one of our organization's projects. 
We use kapitan with jinja2 to transform a JSON file that we use in our deployment system into a YAML file. One of the values that is rendered into our deployment template is the commit id of the application. We figure an issue during this rendering that when the commit has a pattern of a number, interpreters like YTT and Helm3 transforms those values. E.g.: If the commit id is **357e009**, the YTT transforms that value into **3.57e+11**. 
To solve that issue, we ended up creating a custom jinja filter for kapitan that forces the quotes into string values to guarantee that YTT will maintain the value that was informed in the object that was informed to the YAML filter. 

## Proposed Changes
Update the **yaml** jinja2 filter library to guarantee that the output YAML guarantee that the string values are interpreted correctly.


ramaro:
Hey @arielly-parussulo thanks for this. I am a bit confused about it though. What does ruamel does differently that pyyaml doesn't? Do you have an example to reproduce this?

thiagoluiznunes (approved):

arielly-parussulo:
> Hey @arielly-parussulo thanks for this. I am a bit confused about it though. What does ruamel does differently that pyyaml doesn't? Do you have an example to reproduce this?

Hello, @ramaro,
I used ruamel because I saw that we could configure the preserve_quotes config using it. If you know a way of doing that using pyyaml, let me know that I can change this PR. 
The diference of the output of those libs:

### input
```json
{
  "app": {
    "name": "app-1",
    "port": 8080,
    "tag": "357e009"
  }
}
```

### pyyaml
```yaml
app:
  name: app-1
  port: 8080
  tag: 357e009
```

### rumel
```yaml
parameters:
  app:
    name: app-1
    port: 8080
    tag: "357e009"
```

It basically guarantees that the output YAML file will keep the property type. In the case of the **tag** parameter, the value output without quotes will be a number. So in order to keep the property type string (the same as the input), the lib add the quotes.

Issue #852, opened 2022-09-10
Support formatted YAML outputs for multiline yamls


Moep90:
**Support formatted YAML outputs for multiline yamls**

In this examole I used the thanos [helmchart](https://github.com/bitnami/charts/tree/master/bitnami/thanos/)

```yaml
apiVersion: v1
kind: Secret
metadata:
  labels:
    name: thanos-objstore
  name: thanos-objstore
stringData:
  objstore.yml: "type: S3\nconfig:\n  bucket: my-bucket\n  endpoint: s3.example.com\n\
    \  region: eu-de\n  access_key: ?{vaultkv:ey......:embedded}\n\
    \  secret_key: ?{vaultkv:ey.......19:embedded}\n\
    \  http_config:\n    insecure_skip_verify: false\n"
type: Opaque
```
Should look like this:

```yaml
apiVersion: v1
kind: Secret
metadata:
  labels:
    name: thanos-objstore
  name: thanos-objstore
stringData:
  objstore.yml: |
    type: S3
    config:
      bucket: my-bucket
      endpoint: s3.example.com
      region: eu-de
      access_key: ?{vaultkv:..............:embedded}    
      secret_key: ?{vaultkv:..............:embedded}    
      http_config:
        insecure_skip_verify: false
type: Opaque

```

PR #877, opened 2022-11-04
Add formatting flags for `$ kapitan inventory`

MatteoVoges:
## Proposed Changes
  - Added flag `--indent` (`-i`) for `$ kapitan inventory`: same behavior as in `$ kapitan compile`
    > set the indentation spaces in the inventory output to given argument, default is 2
  - Added flag `--multiline-string-style [STYLE]` for `$ kapitan inventory`: same behavior as in `$ kapitan compile`
    > set multiline string style to STYLE, default is `double quotes`
   
This is a extension to #852 and #861 which introduced the flag `--yaml-multiline-string-style [STYLE]`. We have the prefix `yaml` here because it was just a feature for the yaml-output. Now the `inventory` command only prints out yaml, so we don't need the prefix here. But if you (@ramaro) want to name the flag `--yaml-multiline-string-style` for consistency, I would be okay with that. 

Follow-up question: Do we need a short flag name for that (i.e. `-literal-style` or even `-ymss` :slightly_smiling_face: ) ? What do you think and how would you name it?

### Documentation
I didn't find any spot in the docs where `$ kapitan inventory` is documented. It was used in some examples but is nowhere fully explained. I think the users have to figure out for themselves with running `$ kapitan inventory --help`.


ramaro:
@MatteoVoges this is a good idea! I think you're right, as the inventory only emits yaml, it's ok to call it `--multiline-string-style` :D
For the short flag, how about `-L`?

MatteoVoges:
@ramaro Nice suggestion. I applied it now. Should I add the short flag for `$ kapitan compile` as well?

ramaro:
> @ramaro Nice suggestion. I applied it now. Should I add the short flag for `$ kapitan compile` as well?

Yep let's do that! You will also need to rebase from master

MatteoVoges:
@ramaro Done!

Issue #738, opened 2021-05-28
[Feature] Allow to decide when Kapitan renders the secrets


Moep90:
# Describe the feature
When using the helm "provider" of kapitan, the secrets cannot be injected. The reason seams to be that kapitan doesn't inject the secrets into helm rendering.

# To Reproduce

```yaml
$ cat inventory/class/artifactory.yml
parameters:
  artifactory:
    secret:
      accessAdmin:
        secret: ?{base64:targets/${target_name}/accessAdmin_secret||randomstr}

  artifactory:
    chart_name: artifactory
    chart_version: 8.4.0
    namespace: ${namespace}
    helm_values:
      name: artifactory
      artifactory:
        accessAdmin:
          secret: ${artifactory:secret:accessAdmin:secret]

  kapitan:
    dependencies:
    - type: https
      output_path: components/charts/
      source: https://charts.jfrog.io/artifactory/helm/artifactory-${artifactory:chart_version}.tgz
      unpack: True

    compile:
    - input_type: helm
      input_paths:
        - components/charts/artifactory
      output_path: .
      helm_params:
        namespace: ${artifactory:namespace}
        release_name: ${artifactory:chart_name}
      helm_values: ${artifactory:helm_values} 
```
```yaml
$ cat compiled/artifactory/artifactory/templates/secret.yml
apiVersion: v1
kind: Secret
metadata:
  labels:
    app: artifactory
    chart: artifactory-11.9.4
    heritage: Tiller
    release: artifactory
  name: artifactory-artifactory-access-config
stringData:
  access.config.patch.yml: "security:\n  tls: false\n secret: P3tiYXNlNjQ6dGFyZ2V0cy8vYWNjZXNzQWRtaW5fc2VjcmV0fHxyYW5kb21zdHJ9Cg=\n"
type: Opaque
```
```bash
$ echo "P3tiYXNlNjQ6dGFyZ2V0cy8ke3RhcmdldF9uYW1lfS9hY2Nlc3NBZG1pbl9zZWNyZXR8fHJhbmRvbXN0cn0=" | base64 -d
?{base64:targets/${target_name}/accessAdmin_secret||randomstr}
```

`./kapitan compile`
# Expected behavior
1. Kapitan generates the secrets
2. injects them into the helm building process

Moep90:
**Whats the issue?**

[...]

PR #880, opened 2022-11-18
Add support for secrets in helm (#738)

MatteoVoges:
Fixes issue #738 

## Proposed Changes
- New CLI and inventory flag: `--helm-refs` / `helm_refs: true`:
  - decodes the `.data` key in the kinds _Secret_ and _ConfigMap_
  - this allows the kapitan secret engine to get active and create, manage, store or reveal these secrets

--- 


ademariag:
Hello @MatteoVoges, this is a very nice idea, thank you for this.

I think @ramaro's work goes beyond the simple use case of allowing Kapitan secrets, as it opens up to combining generators and charts and much more.

However I think you are onto something here. Things I would suggest:
* put this feature behind a flag (in the inventory, at `compile.input_type: helm` level)
* rather than going through all the objects, limit your scope to some resources only. Although this is not as cool as a recursion, it prevents things from going wild if we have big resources to traverse. I'd limit it for now to:
  * Secrets (under `.data` only)
  * ConfigMap (under `.data` only)

* if we see this is useful, we can expand to other types/paths (or define them in config)

@MatteoVoges @ramaro what do you think?

MatteoVoges:
Hi @ademariag,
thank you for your ideas. Based on them I introduced the new flags (CLI and inventory)
* `--helm-secrets` / `helm_secrets: true`: 
    If activated the `.data`-values of `kind: Secret`'s and `kind: ConfigMap`'s will be searched after base64-encoded strings. Strings found will then be decoded, so that the RefController can handle them.
* `--encode_base64` / `encode_base64: true`:
  If activated the same `.data`-values as above get base64 encoded again, but after the RefController could do its stuff.

The second flag is part of some post-processing. However I wrote my code directly into the `base.py` module. I'm a bit unhappy about that, because the flag only really works with helm as inputtype.

### TODO:
* Find good names for flags and methods
* Add documentation
* Add tests?

I think, if these things are done, the PR is ready for review.

Do we want a "official" Post-Compile step, where we can manipulate directly the output? Or am I missing something, and we have that already?

kapitan/cli.py
```diff
@@ -286,13 +284,25 @@ def build_parser():
...
     )
-
     compile_parser.add_argument(
         "--yaml-dump-null-as-empty",
         default=from_dot_kapitan("compile", "yaml-dump-null-as-empty", False),
         action="store_true",
         help="dumps all none-type entries as empty, default is dumping as 'null'",
     )
+    compile_parser.add_argument(
+        "--helm-secrets",
```
ramaro, line 294: Can I suggest that we rename this into `--helm-refs`? Reason being that not only secrets could be used here but also other non encrypted refs if the user wishes to.

kapitan/cli.py
```diff
@@ -286,13 +284,25 @@ def build_parser():
...
+        action="store_true",
+        default=from_dot_kapitan("compile", "helm-secrets", False),
+        help="enable kapitan secret engine on helm refs",
+    )
+    compile_parser.add_argument(
+        "--encode-base64",
```
ramaro, line 300: Since this is helm specific, maybe call it instead `--helm-refs-base64`?

kapitan/inputs/base.py
```diff
@@ -178,3 +184,35 @@ def __enter__(self):
...
+
+
+def check_data_for_b64(yml_obj):
+    """
+    check for .data in kind: Secret / ConfigMap
+    replace b64 encoded strings
+    """
+    for item in yml_obj:
+        kind = item.get("kind", None)
+        if kind and (kind == "Secret" or kind == "ConfigMap"):
```
ramaro, line 199: I think it's cleaner to do instead `if kind in ('Secret, 'ConfigMap'):`

kapitan/inputs/helm.py
```diff
@@ -272,3 +282,44 @@ def load_chart(self):
...
+    """
+    for item in yml_obj:
+        kind = item.get("kind", None)
+        if kind and (kind == "Secret" or kind == "ConfigMap"):
+            if item.get("data", None):
+                # Option 1:
+                # item["data"] = replace_b64_refs(item["data"])
+
+                # Option 2:
+                item["stringData"] = replace_b64_refs(item["data"])
```
ramaro, line 300: Yeah I think option 2 is best. We are mangling data anyway, so best to make it into valid content in stringData, as data requires its content to be base64 and a ref is certainly not valid base64.

ramaro, line 300: Do we need to recurse through .data? Since we're only support Secrets and ConfigMaps and data is a dictionary, wouldn't it be enough to iterate through its items and replace the values? I might be missing why recursion is needed

kapitan/cli.py
MatteoVoges: Sure. It's done in the latest commit

kapitan/cli.py
MatteoVoges: Done in the latest commit

kapitan/inputs/base.py
MatteoVoges: You're right. I restructured the code a bit and applied your suggestion

kapitan/inputs/helm.py
MatteoVoges: I am not sure. I thought that any yaml object could be very deeply nested, e.g. a dictionary with lists in it....
I talked with Danny about this, and we both agreed that recursion is needed here. But we can discuss this later in the workgroup meeting.

MatteoVoges: I wish it would solve the original problem. It turned out that the key `.stringData` is not allowed in `ConfigMap`s. So now we have a mixed solution: 
* For `Secret`s `.data` is replaced with `.stringData` (originally option 2). 
* For `ConfigMap`s after the compile-step the `.data` key is base64 - encoded again (originally option 1).

But again, we can clarify this later in the workgroup meeting.

PR #891, opened 2022-11-24
improve helm documentation

ademariag:
Improve helm documentations

* adds example helm target and class
* add dependency download of helm chart
* add charts to repo @ramaro WDYT are you ok with that?
* fix auto-expand for other examples
* makes use of common class import in examples for clarity

### DIFF
--- a/docs/kapitan_overview.md
+++ b/docs/kapitan_overview.md
@@ -182,30 +182,7 @@ Kapitan can also be used to manage **Helm**, giving you access to its enourmous 
         Please use the `kapitan compile --fetch` flag if the chart has not been downloaded already
 
     ```yaml
-    parameters:
-        namespace:
-        nginx:
-            version: 4.4.0
-            replicas: 2
-            name: ${target_name}
-            namespace: ${target_name}
-        kapitan:
-            dependencies:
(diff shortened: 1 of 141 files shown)


ademariag:
@ramaro I gave up on this because I can't fix the tests. Would you be able to understand how to fix it when you have time?

MatteoVoges:
Hey @ademariag 
the tests use directory hashes to prove that the compiled test output is exactly the expected (human-checked) output. You are changing these directories, so the hashes are no longer the same. You could fix this by also updating the hash-test to your new hash-value `5582798e46a03cffc96588c83b989bd6a6f9a9cb54cc34aba86407789dfd7231`. I hope that it will fix your tests. Personally I don't like your PR #901 , because I think its important to check if the whole output is correct, no matter what changes have been made. In example the next PR-submitter would have these issues, because the hash have changed, and he thinks that he is wrong.

ademariag:
HI @MatteoVoges , thank you for your suggestion. Unfortunately there is no hash to update, because the directory is checked at runtime  https://github.com/kapicorp/kapitan/blob/master/tests/test_compile.py#L119

PR #907, opened 2022-12-07
fix #906: Jinja input_params gets reset

### LINKED ISSUE #906: Kapitan fails to render Jinja template with `input_params`
**Describe the bug/feature**
Kapitan fails to render Jinja with `input_params`

while defining a simple `input_params` for a`jinja` template I realised that it was not working

Will add more soon
**To Reproduce**

```jinja
gcloud --project {{input_params.gcp_project_id}}
```

```yaml
parameters:
  scripts:
  - templates/scripts/gcloud
  scripts_params:
    gcp_project_id: ${gcp_project_id}
  kapitan:
    compile:
    ...
      - output_path: scripts
        input_type: jinja2
        input_paths: ${scripts}
        input_params: ${scripts_params}
```

```
Jinja2 error: failed to render /src/templates/scripts/gcloud: Jinja2 TemplateError: 'dict object' has no attribute 'gcp_project_id', at /src/templates/scripts/gcloud:5
Compile error: failed to compile target: dev
```

**Expected behavior**

```
gcloud --project isolabs-platform
```

This is possibly caused by this loop which would allow for input_params to only work for the first file being compiled

https://github.com/kapicorp/kapitan/blob/master/kapitan/inputs/base.py#L55

ademariag:
Fixes issue #906

## Proposed Changes

- do not reset input_params in between compile
- clean up the way input_params are passed to Kadet and Jinja
- also exposes compiled path to jinja (same as Kadet)
- adds test

## Tests before change:

```
kapitan compile -t jinja2-input-params
Rendered inventory (0.04s)

Jinja2 error: failed to render /usr/local/google/home/ademaria/isolabs/kapitan/tests/test_resources/templates/pod.yml: Jinja2 TemplateError: 'dict object' has no attribute 'name', at /usr/local/google/home/ademaria/isolabs/kapitan/tests/test_resources/templates/pod.yml:4
Compile error: failed to compile target: jinja2-input-params
```

## Test after change
```
kapitan compile -t jinja2-input-params
Rendered inventory (0.04s)
Compiled jinja2-input-params (0.01s)
```

### DIFF
--- a/kapitan/inputs/jinja2.py
+++ b/kapitan/inputs/jinja2.py
@@ -30,7 +30,7 @@ class Jinja2(InputType):
         """
         reveal = kwargs.get("reveal", False)
         target_name = kwargs.get("target_name", None)
-        
+
         input_params = self.input_params
         # set compile_path allowing jsonnet to have context on where files
         # are being compiled on the current kapitan run
@@ -42,7 +42,7 @@ class Jinja2(InputType):
         context["inventory"] = inventory(self.search_paths, target_name)
         context["inventory_global"] = inventory(self.search_paths, None)
         context["input_params"] = input_params
-        
+
         jinja2_filters = kwargs.get("jinja2_filters")
 
         for item_key, item_value in render_jinja2(
--- a/kapitan/inputs/jinja2.py
+++ b/kapitan/inputs/jinja2.py
@@ -16,11 +16,8 @@ logger = logging.getLogger(__name__)
 
 
 class Jinja2(InputType):
-    def __init__(self, compile_path, search_paths, ref_controller):
+    def __init__(self, compile_path, search_paths, ref_controller, input_params={}):
         super().__init__("jinja2", compile_path, search_paths, ref_controller)
-        self.input_params = {}
-
-    def set_input_params(self, input_params):
         self.input_params = input_params
 
     def compile_file(self, file_path, compile_path, ext_vars, **kwargs):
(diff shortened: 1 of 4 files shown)


kapitan/inputs/jinja2.py
```diff
@@ -34,13 +31,18 @@ def compile_file(self, file_path, compile_path, ext_vars, **kwargs):
...
+        # are being compiled on the current kapitan run
+        # we only do this if user didn't pass its own value
+        input_params.setdefault("compile_path", compile_path)
+
         # set ext_vars and inventory for jinja2 context
         context = ext_vars.copy()
         context["inventory"] = inventory(self.search_paths, target_name)
         context["inventory_global"] = inventory(self.search_paths, None)
-        context["input_params"] = self.input_params
-        # reset between each compile if jinja2 component is used multiple times
```
ramaro, line 42: I'm confused. Why is this not desired?

ademariag: As explained offline, resetting the input_params causes the bug. Example:

```yaml
    compile:
      - output_path: test-1/      <---   We only set input_params at creation of each compile obj kapitan/targets.py:L477
        input_type: jinja2
        input_params:
          name: test1
          namespace: ns1
        input_paths:
          - templates/pod.yml     <---   reset happens after here
          - templates/service.yml <---   here input_params is empty (changed name to give a possible use case)
 ```

ademariag:
@sebradloff just getting a final approval from you in case you know of any reason why it was implemented this way?

sebradloff:
@ademariag sorry I seemed to have lost this notification. I don't remember why I specifically implemented it this way, but the goal was to be able to reuse the same jinja template within a single target. If that still works with this code, I don't see any issues.

ramaro (approved):

PR #913, opened 2022-12-18
improve docs and examples

ademariag:
Fixes issue #

## Proposed Changes

-
-
-


examples/kubernetes/compiled/helm/manifests/helm-namespace.yml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
ademariag, line 1: this is not needed

Issue #805, opened 2021-12-06
Ability to strip .j2 extension when performing jinja2 compilation and potentially fully rename


KostasKgr:
**Describe the feature**
Ability to strip .j2 extension when performing jinja2 compilation.
It is useful to have .j2 extension on files for at least two reasons:
- IDEs provide syntax highlighting
- People know it is a jinja2 template file without having to read through it

If a file path is specified for output_path then this gets interpreted as a folder name.
https://kapitan.dev/compile/#specifying-inputs-and-outputs

Perhaps general renaming would cover more use cases, such as compiling a template to multiple outputs in the same node or providing a better semantic name to the template.

Raising this after discussing stripping .j2 in kapitan slack channel.

I don't have some good recommendation on how to approach it, as the current interface of compile expects multiple input files and one output directory.
Perhaps ansible template module can be used for inspiration: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/template_module.html

**To Reproduce**
```
parameters:
  kapitan:
    compile:
    - output_path: app/my_app1/config.cfg
      input_type: jinja2
      input_paths:
        - templates/generic_python_config.cfg.j2
      input_params:
      	name: my_app1
      	# ...
```

Output of the above would be compiled to app/my_app1/config.cfg/generic_python_config.cfg.j2

**Expected behavior**
Able to drop .j2 extension and rename.
Ouput would be compiled to app/my_app1/config.cfg

**Additional context**
Using kapicorp/kapitan:v0.30.0-rc.0

beddari:
Valid problem! I have not seen the slack discussion, but my best idea to solve this would be to simply strip any `.j2` extension by default.

That would break any current users wanting to write e.g a `example.j2` file to output_path, but that is maybe more of an edge case, and could be fixed by naming the input file `example.j2.j2`?

beddari:
The referenced PR implements stripping and so I guess what remains is more a decision about what would be the best default behavior - and again, changing it to strip `.j2` by default seems sensible.

Moep90:
suffix_remove and suffix_stripped config options are now available in the jinja2 input type.

PR #916, opened 2022-12-31
feat/strip-j2-jinja: implement rstrip for jinja2 input

jkrzemin:
Added options "strip_postfix" and "stripped_postfix" to jinja2 input. This allows to remove output files extensions which is often the case for jinja2 template files (.j2).

Stripping is disabled by default and default value to strip is ".j2".

Closes issue #805

## Proposed Changes
- add option "strip_postfix" to jinja2 input
- add option "stripped_postfix" to jinja2 input
- optional removal of file postfixes (i.e. extensions) on output files

### DIFF
--- a/kapitan/inputs/jinja2.py
+++ b/kapitan/inputs/jinja2.py
@@ -16,9 +16,11 @@ logger = logging.getLogger(__name__)
 
 
 class Jinja2(InputType):
-    def __init__(self, compile_path, search_paths, ref_controller):
+    def __init__(self, compile_path, search_paths, ref_controller, args):
         super().__init__("jinja2", compile_path, search_paths, ref_controller)
         self.input_params = {}
+        self.strip_postfix = args.get("strip_postfix", False)
+        self.stripped_postfix = args.get("stripped_postfix", ".j2")
 
     def set_input_params(self, input_params):
         self.input_params = input_params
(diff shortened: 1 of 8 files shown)


beddari:
Would it improve clarity a bit to change the variable names used here?

```
suffix_remove: true
suffix_list:
  - .j2
```

Additionally, we would like this feature to become a new default for the jinja2 input type, I added a comment about that to #805

jkrzemin:
Fair point, this naming makes more sense. I'll update the PR tomorrow

ademariag:
IMHO passing the list of suffix to strip might overcomplicate the problem, but curious to see what @ramaro thinks.

I'd stick to expect, if the flag is enabled, that all files have a `.j2` extension, and remove it present.
what happens if the file doesn't have a j2 extension? we silently ignore it? we don't process it?

the one benefit of expecting the .j2 extension is that it would be easier to exclude from tests and linting files that are templates, which at the moment is not possible.

jkrzemin:
Currently if file doesn't have an expected suffix (.j2 is a default, can be overridden) it will simply pass it over to rendering.

I don't think that skipping files without suffix is a good idea - you still have to provide them explicitly in compile step so I think that we can assume that if file ends up here, it should be rendered.

ademariag:
> you still have to provide them explicitly in compile step so I think that we can assume that if file ends up here, it should be rendered.

Kapitan also supports passing a directory, in which case you won't have explicitly all files declared in the inventory.

But yes I also think we can just make it compile anyway

PR #930, opened 2023-01-23
Update requirements.txt to more recent versions

gburiola:


uberspot (approved):
lgtm, but worth getting another pair of eyes/testing more.

Issue #939, opened 2023-02-08
Build multi-arch docker image


antonincms:
**Describe the bug/feature**
The kapitan docker image is currently amd64 only, and this slows down our build (5x slower) when using the amd64 docker image on M1/M2 macs. It would be nice to build multi-arch image instead.

Would you accept such PR ? I would expect it to have to modify 
- The dockerfile
- The github actions docker-image.yml

PR #944, opened 2023-02-18
Build multi-arch docker images

antonincms:
Fixes issue #939

## Proposed Changes

On a M1 / M2 mac the performance are 5x worse for our use case if using the amd64 docker.
We would like to have the possibility of using a multi-plateform image, and this PR allows that : 

- Update the Dockerfile to use the TARGETARCH environment variable to pull the correct version of go.
- Update the docker-image github action pipeline to setup QEMU and buildx, 
   and then building multi-plateform images of kapitan.

Building an image on a computer of the same architecture should work as previously. 
I could add some documentation on how to do locally what the CI does (build a multi-arch image with a single tag).

This was inspired by this bit of documentation about building multi-plateform images : https://docs.docker.com/build/ci/github-actions/examples/#multi-platform-images
There is also this more general documentation about how to set this up outside of CI : https://docs.docker.com/build/building/multi-platform/

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -40,11 +40,17 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
       - name: Build PR/versioned tags
         if: github.ref != 'refs/heads/master'
         uses: docker/build-push-action@v2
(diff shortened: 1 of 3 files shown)


uberspot:
Overall looks good. I've rebased to re-run tests.
Also `I could add some documentation on how to do locally what the CI does (build a multi-arch image with a single tag).`
It would be indeed beneficial if you added a pointer for this. Maybe as a comment in the makefile or in contributing.md. Either works (?)

antonincms:
I added some comments in the CI and a paragraph in the README, is it ok ?

uberspot (approved):

ramaro (approved):
looks good, thank you @antonincms !

antonincms:
Thank you for the review !

Issue #972, opened 2023-03-20
Kapitan fails to compile with compose_node_name: true


ademariag:
**Describe the bug/feature**
When using https://github.com/kapicorp/kapitan/pull/933 and `compose_node_name: true`, kapitan fails to compile when targeting individual target or labels

```
/kapitan compile -t test.dev
Rendered inventory (0.24s)
Compiled test.dev (0.12s)

Unknown (Non-Kapitan) error occurred:
Traceback (most recent call last):
  File "/opt/venv/lib/python3.7/site-packages/kapitan/targets.py", line 179, in compile_targets
    shutil.copytree(temp_path_target, compile_path_target)
  File "/usr/local/lib/python3.7/shutil.py", line 318, in copytree
    names = os.listdir(src)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpufn67nog.kapitan/compiled/test.dev'

```

**To Reproduce**

1. Enable compose_node_name: true following https://github.com/kapicorp/kapitan/pull/933
2. run `kapitan compile -t target_name`
3. crash

PR #973, opened 2023-03-20
Fix 972: handle compose_node_name

ademariag:
Fixes issue #972

## Proposed Changes

- handle compose_node_name

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,33 @@
-*.pyc
-*.swp
-*.bak
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+*.h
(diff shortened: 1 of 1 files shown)


kapitan/targets.py
```diff
@@ -169,9 +169,10 @@ def compile_targets(
 
         # if '-t' is set on compile or only a few changed, only override selected targets
         if updated_targets:
```
gburiola, line 171: is this `if` still needed if you're no longer using that variable?

gburiola (approved):

kapitan/targets.py
ademariag: yes it's still needed to distinguish between "all targets" and "some targets" so that we copy individual paths instead of everything

PR #981, opened 2023-03-27
chore: new Poetry-only Dockerfile

gruzewski:
## New Poetry-only Dockerfile

Adds a new Dockerfile that installs all dependencies using `poetry`. The idea is to build it alongside the existing image and test it extensively.

**Testing:**
I built the new Dockerfile and re-compiled `examples/kubernetes` with it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,42 @@
+# Python compiled files
+__pycache__
+*.pyc
+*.pyo
+*.pyd
+
+# Versioning
+.venv
+.Python
+.tox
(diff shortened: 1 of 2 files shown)


uberspot:
Nice, should we also modify the github action to build this and publish other tags with it? Or is that already covered?

gruzewski:
> Nice, should we also modify the github action to build this and publish other tags with it? Or is that already covered?

I was planning to do that in a separate PR. But I am happy to do it here.

PR #982, opened 2023-03-30
Fix typo in kadet prune input handling

ademariag:
Fixes issue with kadet pruning

```
  kapitan:
    compile:
      - output_path: manifests
        input_type: kadet
        output_type: yml
        prune: false   <------- honour this prune to override the default
  ```

## Proposed Changes

- a typo was preventing "prune" to be read from the input definition ([see](https://github.com/kapicorp/kapitan/blob/f5fd147682e7e47943e2330942d7610960fec642/kapitan/inputs/base.py#LL81C23-L81C23))

### DIFF
--- a/kapitan/inputs/base.py
+++ b/kapitan/inputs/base.py
@@ -64,7 +64,7 @@ class InputType(object):
         target_name = ext_vars["target"]
         output_path = comp_obj["output_path"]
         output_type = comp_obj.get("output_type", self.default_output_type())
-        prune_input = comp_obj.get("prune", kwargs.get("prune", False))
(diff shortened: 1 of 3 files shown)


gburiola:
@ademariag I don't get this.

Where is the typo?

And shouldn't the variable be renamed to `prune_output` or `prune_compiled` as it's the compiled content that is being pruned right?

ademariag:
@gburiola this is a typo because I know that the original author (me: https://github.com/kapicorp/kapitan/issues/444) intended to use the other variable instead of prune :) 

I agree the name is not great, so I can change it to `prune_output`

gburiola (approved):

uberspot (approved):

PR #986, opened 2023-04-06
refactor(fetch): rework force-fetch process

MatteoVoges:
Reference #862 #858 

## Proposed Changes
- Reworked force-fetch functionality with property in `force_fetch: true` in kapitan dependencies
- Fixed logical error where targets were skipped, if one dependency has no entry `force_fetch`
- Now only those dependencies with `force_fetch: true` get fetched and not the entire targets dependencies

There's nothing to do regarding tests and documentation, because it exists already.

--- 


kapitan/targets.py
```diff
@@ -126,31 +128,43 @@ def compile_targets(
...
-            # fetch dependencies from targets with force_fetch set to true
+
+                for dep_entry in dependencies:
+                    if not dep_entry.get("force_fetch", False):
+                        # if no fetch is needed remove the dependency and don't fetch it
+                        target["dependencies"].remove(dep_entry)
+
+                # fetch dependencies with force_fetch set to true
+                if target.get("dependencies", []):
+                    fetch_objs.append(target)
```
MatteoVoges, line 154: idk why I reversed the logic here: so if it hasn't the flag, then we remove it...
Should I fix that?

ademariag: I think it makes sense to have a way to force on the command line "not to force" in case you are running from a place with no external connectivity

So i'd probably leave it in as it is, we can change it later if it doesn't make sense.

MatteoVoges: Absolutely, but I think it would work the same if we say, that: if `force_fetch: true` --> force_fetch, instead of removing it from the list, if it don't have `force_fetch: true`

MatteoVoges: > a way to force on the command line "not to force"
So you want a `--dont-force-fetch` flag?

ademariag: perhaps `--offline-mode` so it can be a generic way to turn off all attempts to download things?

@ramaro what to you think?

I am thinking of the case where you run it in places with no connectivity or you just don't want to change anything. It would suck if in those cases it kept failing.

Alternatively, we could handle it with an additional:

`fail_on_error: false`

PR #994, opened 2023-04-26
fix(tests): remove merge annotations from #993

MatteoVoges:
## Proposed Changes
- Fix the tests, which were failing because of merge information introduced by #993
- I noticed that [https://kapitan.dev/inventory/introduction.md](https://kapitan.dev/inventory/introduction.md) gives an 404 error, so I updated the path in `mkdocs.yaml`

--- 
### Contributed by  [](https://www.nexenio.com)

PR #993, opened 2023-04-23
Address #992

ademariag:
Fixes issue #992

## Proposed Changes

- Improved documentation layout
-
-

PR #995, opened 2023-04-27
feat(inventory): introduce OmegaConf as optional inventory backend

### LINKED ISSUE #990: Provide an alternative backend to the inventory
I think reclass is a bit deprecated and is really hard to extend. It's poorly documented, it's quite annoying to have it as a submodule and we don't even use all the features.
My solution to this problem would be to move to modern tools (python packages) that work with ruamel.yaml because it is maintained. I don't have any specific tools in mind, but some suggestions are:
* [yamlpath](https://github.com/wwkimball/yamlpath)
* [figgy](https://github.com/figtools/figgy)
* [omegaconf](https://omegaconf.readthedocs.io/en/2.3_branch/usage.html)
* [hiyapyco](https://github.com/zerwes/hiyapyco)

MatteoVoges:
Fixes #990 

## Proposed Changes
- introduces OmegaConf as new backend
- working branches are `kapicorp/nexenio-dev`, `neXenio/dev`
- see docs for all features and changes: `docs/pages/inventory/omegaconf.md`

## To do
* [X] Documentation
* [X] Tests
* [ ] useful logging for debugging
* [ ] useful error handling with hints how to fix
* [ ] Feature TODO:
  * [ ] OmegaConf can detect duplicate keys that are redundantly defined
  * [ ] make migrate as command, not flag
  * [ ] migrate should detect resolvers and don't migrate them
  * [ ] migrate should convert keys with `.` to using the `access` resolver 
  * [ ] `user_resolver_path` should be controlled by a flag
  * [ ] support `global_inventory` in multiprocessing
  * [ ] performance: hash and cache classes / targets and store them
 
--- 


MatteoVoges:
Note: the tests are failing because of the changes made in #993 
Fix for that exists already in #994

gburiola:
This is very exciting indeed. I haven't heard of OmegaConf before. Great find!!

reclass is great but after you use it for a while you find lots of small things that could be improved and it's a bit sad that it's not longer being maintained. So having an alternative is great.

However, reclass is everywhere in the kapitan code. So my suggestion is to open a separate PR where you just make kapitan code more generic and ready to accept a different type of inventory, without adding OmegaConf just yet. For example the changes you have in this PR to rename `inventory_reclass` to `get_inventory`. I think it's simpler to review and ensure it's backward compatible.

I believe there are lots of other places (in addition to `inventory_reclass` function) that needs to be adjusted.

For example:

- kapitan searchvar
- kapitan lint

MatteoVoges:
> However, reclass is everywhere in the kapitan code. So my suggestion is to open a separate PR where you just make kapitan code more generic and ready to accept a different type of inventory, without adding OmegaConf just yet. For example the changes you have in this PR to rename `inventory_reclass` to `get_inventory`. I think it's simpler to review and ensure it's backward compatible.

Yeah, I will do that. But I leave the changes in here as well, to see the full view of the change.
I don't know how we want to support several inventories. With this pr, we can toggle between OmegaConf and reclass by setting the flag `--omegaconf` or not. If we have several options, should each backend get their own _toggle_ flag? Or maybe a specification flag would make sense, like `--inv=MyInv` or `--inv=OmegaConf`.


> I believe there are lots of other places (in addition to `inventory_reclass` function) that needs to be adjusted.
> 
> For example:
> 
>     * kapitan searchvar
> 
>     * kapitan lint

I don't think so, because the rendered inventory is the same at the moment and the commands you mentioned just fetches the inventory and then do their stuff. I think, it is really important, that all backends produce the same output structure, so that we don't have any different cases in the kapitan site.

Does this make sense?

MatteoVoges:
There is some error in the `OmegaConf.merge()` function, because it doesn't support merging of list objects. This is fixed in https://github.com/omry/omegaconf/pull/1082 , but we have to wait until the next release, to use the feature

MatteoVoges:
Since the list merge in Omegaconf now works, I included the module locally and was able to test it successfully. Migrating the inventory only works with ruamel.yaml at the moment, as this is the only way to preserve comments. Also the local module has to be replaced by the PyPi module again, if it is released. 

If you compile the inventory this is the only change: (`examples/kubernetes/compiled/minikube-nginx-helm/nginx-deploy.sh`)
```diff
- KUBECTL="kubectl -n None" #(2)!
+ KUBECTL="kubectl -n minikube-nginx-helm" #(2)!
``` 

So the namespace is now correct. I think that this change was even a bug in reclass and so this is fixed now. But maybe it could be intentional...

MatteoVoges:
@ademariag @ramaro 
Most of the development is finished and ready to merge now. However there are tests and documentation missing. I will add them, if we agree on a final first-version of omegaconf, because then I don't have to rewrite everything :smile: 
I would present you the current state of the inventory in the next work group meeting with some examples.
What do you think about this?

pyproject.toml
```diff
@@ -51,6 +51,7 @@ jsonschema = "^4.17.3"
 kadet = "^0.2.2"
 python-gnupg = "^0.4.7"
 pyyaml = "^6.0"
+omegaconf = { git = "https://github.com/neXenio/omegaconf.git", branch = "dev" }
```
ramaro, line 54: why do we have to use dev branch?

kapitan/resources.py
```diff
@@ -304,6 +306,52 @@ def generate_inventory(args):
...
+
+    # if inventory is already cached theres nothing to do
+    if cached.inv:
+        return cached.inv
+
+    # get parsed args from cached.py
+    args = list(cached.args.values())[0]
+    use_omegaconf = args.omegaconf
+    try:
+        migrate_omegaconf = args.migrate
```
ramaro, line 323: not really use why we need to do this inside a try?

kapitan/resources.py
```diff
@@ -304,6 +306,52 @@ def generate_inventory(args):
...
+            # logger.info("REAL_TIME (%.2fs)", time.time() - inv_start)
+        except Exception as e:
+            if not migrate_omegaconf:
+                logger.warning("Make sure to migrate your inventory using --migrate")
+            raise InventoryError(e)
+    else:
+        logger.debug("Using reclass as inventory backend")
+        # inv_start = time.time()
+        inv = inventory_reclass(inventory_path, ignore_class_notfound)
+        # logger.info("REAL_TIME (%.2fs)", time.time() - inv_start)
```
ramaro, line 349: can we remove the commented code?

kapitan/resolvers.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python3
+
+# Copyright 2023 nexenio
```
ramaro, line 3: unfortunately we need to remove this. We need this instead:

```

# Copyright 2023 The Kapitan Authors
# SPDX-FileCopyrightText: 2020 The Kapitan Authors <kapitan-admins@googlegroups.com>
#
# SPDX-License-Identifier: Apache-2.0

```

kapitan/omegaconf_inv.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python3
+
+# Copyright 2023 nexenio
```
ramaro, line 3: same copyright msg here too

kapitan/resources.py
ademariag: use perhaps:

```
migrate_omegaconf = args.get("migrate", False)
```

MatteoVoges (commented):
Hey @ramaro and @ademariag ,
This is an outdated version of the implementation. I'm now working on a nexenio/dev branch to work more independently, because some of my PR's depend on each other and we at nexenio wanted to have a kapitan branch with all of my (upcoming) features. 
Of course I will continue sending parts to this upstream, but I think this will happen less frequently. 

I can show you my progress on friday and then you can review it properly.

MatteoVoges:
Closing, because its heavily out of date

Issue #1005, opened 2023-05-06
deps: jsonnet version are unaligned and outdated


antonincms:
**Describe the bug/feature**

The jsonnet and gojsonnet version are outdated and we could upgrade to get new stdlib functions and better perf.

Also, the gojsonnet version is older than jsonnet, which mean when using gojsonnet which has the best performance we have some function missing.

PR #1006, opened 2023-05-06
deps: upgrade jsonnet and gojsonnet to 0.20.0

antonincms:
Fixes issue #1005

## Proposed Changes

The related jsonnet changelog are :
- https://github.com/google/jsonnet/releases/tag/v0.18.0
- https://github.com/google/jsonnet/releases/tag/v0.19.0
- https://github.com/google/jsonnet/releases/tag/v0.19.1
- https://github.com/google/jsonnet/releases/tag/v0.20.0

Update search_imports function to return bytes according to the v0.19.0 changes mentioned there: https://github.com/google/jsonnet/releases/tag/v0.19.1

I hesitated whether to put bytes in cache directly or keeping current logic and encoding only at return.


MatteoVoges (approved):
lgtm :+1:

antonincms:
I rebased the branch onto master and cleaned up unrelated import reordering that my IDE does.

ramaro:
>I hesitated whether to put bytes in cache directly or keeping current logic and encoding only at return.

Thanks @antonincms - can you explain this further?

antonincms:
@ramaro I was hesitant whether to put bytes in JSONNET_CACHE (and encode them only once), or encode when pulling from the cache (as currently done in the PR). I think the former is the simpler way, and I don't think encode should matter that much from a performance standpoint.

ramaro:
thanks @antonincms !

PR #1031, opened 2023-09-01
chore: bump Dockerfile golang version to 1.20

antonincms:
## Proposed Changes

As specified by the golang [release policy](https://go.dev/doc/devel/release#policy), "each major go release is supported until there are two newer major releases", which mean that the oldest supported version is currently go 1.20. 

I propose we upgrade the golang version used in the Dockerfile to build go-jsonnet to go 1.20.7 (but we could switch to go 1.21.0 as well if preferred).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN python -m venv $VIRTUAL_ENV \
     && pip install --upgrade pip yq wheel poetry==$POETRY_VERSION
 
 # Install Go (for go-jsonnet)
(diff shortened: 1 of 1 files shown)


ademariag (approved):

antonincms:
It seems that the building CI is a bit flaky, but it works.

antonincms:
@ademariag it seems that nothing happened even with auto-merge (maybe because of skipped tests I don't know).

antonincms:
Thanks for the review !

PR #1036, opened 2023-09-10
Improve docker file to allow for caching: Part 1 of many

ademariag:
## Proposed Changes
Part 1 of a series of changes from https://github.com/ademariag/kapitan/tree/master/.github/workflows to improve the build process.

no functional changes expected

Done incrementally for easier review

* Shuffle things around to take advantage of docker caching
* Move to use poetry for pre-loading dependencies
* Adds concurrency to prevent multiple jobs from running for the same code
* reduce the number of triggered runs

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -49,6 +49,9 @@ jobs:
           echo ${{ env.REF_NAME }}
       - name: Login to DockerHub
         uses: docker/login-action@v2
+        env:
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME}}
+        if: env.DOCKERHUB_USERNAME != null
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,6 +1,9 @@
 ---
 name: Docker Build and Push
-
+concurrency:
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-image.yml
```diff
@@ -1,14 +1,27 @@
...
+      - 'mkdocs.yml'
+      - 'CNAME'
+
+  release:
+    types: [created]
+
+  pull_request:
     paths-ignore:
       - 'docs/**'
-  workflow_run:
```
ademariag, line 7: This was causing this workflow to checkout master instead of the PR we wanted to look at.

So we were pushing to master on every PR. 

We should have checked out `github.event.workflow_run.head_branch` instead

.github/workflows/docker-image.yml
```diff
@@ -1,14 +1,30 @@
 ---
 name: Docker Build and Push
-on:
+concurrency:
```
ademariag, line 3: This prevents the workflow from running multiple times for the same code

.github/workflows/docker-image.yml
```diff
@@ -63,10 +82,14 @@ jobs:
           tags: kapicorp/kapitan:${{ format('{0}', env.REF_NAME ) }},kapicorp/kapitan:latest
           platforms: linux/amd64,linux/arm64
           file: Dockerfile
-      - name: Test Dockerfile in current ref
-        run: |
-          [ ${{ env.REF_NAME }} == "master" ] && tagname="latest" || tagname=${{ env.REF_NAME }}
-          docker run -t --rm kapicorp/kapitan:${tagname} --version
+      # - name: Test Dockerfile in current ref
```
ademariag, line 85: See comment

PR #1053, opened 2023-09-15
Nexenio/overhaul vault pr

ademariag:
Fixes for PR https://github.com/kapicorp/kapitan/pull/958

## Proposed Changes

- improves tests
-
-


tests/test_copy.py
```diff
@@ -16,10 +16,14 @@
...
 from kapitan.inputs.copy import Copy
 from kapitan.utils import directory_hash
+import logging
+
+logger = logging.getLogger(__name__)
 
 search_path = ""
 ref_controller = ""
 test_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "test_copy")
+logger.info("test_path: %s", test_path)
```
MatteoVoges, line 26: remains of debugging?

MatteoVoges (approved):
Thanks. It looks like everything made it into the PR. There are some newer modifications in nexenio-dev, but for now this is fine and is ready to get integrated!

MatteoVoges:
Hey @ademariag , I tried fixing the tests already but there is some issue with the container. We tried a basic script in github actions, with just starting the container und trying to curl it, and even that fails. Maybe the ci is blocking some ports?

ademariag:
Have these tests ever worked? As in connecting to a container?

MatteoVoges:
Yeah, they did. And for some reason for some commits they still work. I don't have any clue what causes them to fail, because locally everything works fine

ademariag:
> Yeah, they did. And for some reason for some commits they still work. I don't have any clue what causes them to fail, because locally everything works fine

I think I fixed it. the issue is that "container == running" is not the same as vault being able to receive a connection.

So I have added a try-catch

PR #1058, opened 2023-09-16
Pre-commit changes

ademariag:
Fixes issue #

## Proposed Changes

- fix lint and other pre-commit changes
-
-


MatteoVoges (approved):
What about adding isort-checking in the lint job?

MatteoVoges:
> What about adding isort-checking in the lint job?

This would just require adding this action to it:
`- uses: isort/isort-action@master`

PR #1099, opened 2023-12-03
Set logger error

dav9:
Fixes kapitan compile shows output from like: 

creating ... 
copying ... -> ...

It comes from: https://github.com/pypa/distutils/blob/main/distutils/file_util.py#L135

## Proposed Changes

* setup all loggers on error
* keep kapitan and reclass on info as output such as Rendered inventory and Compiled target are emitted through it.
* kapitan inventory and compile has a paramter --logging-config that allows a yaml file.

## Docs and Tests

* [] Tests added
* [ ] Updated documentation

### DIFF
--- a/kapitan/__init__.py
+++ b/kapitan/__init__.py
@@ -7,32 +7,35 @@
 
 import os
 import sys
-import logging
 
-
-def setup_logging(name=None, level=logging.INFO, force=False):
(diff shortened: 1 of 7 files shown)


ademariag:
we removed distutils so no longer relevant. But thank you and sorry it took so long to review

PR #1114, opened 2024-01-11
feat: enable multiprocessing on compile object level

MatteoVoges:
## Proposed Changes
* new flag `--multiprocess-objects`
* splits compile_obj from targets and run them in parallel, to balance out heavy targets

## Performance Measurements 
* for inventories with targets with equally distributed `target to compile_obj` ratio --> 1:1
* for inventories with unbalanced ratios: performance gain up to 400 %

## Problems with tests
The tests are failing because of the `removal` target. The order of execution matters in this target. I think, we should change this example because compile objects should not depend on each other neither different targets should.

### DIFF
--- a/examples/kubernetes/compiled/removal/copy_target
+++ b/examples/kubernetes/compiled/removal/copy_target
@@ -1,0 +1,1 @@
+for_testing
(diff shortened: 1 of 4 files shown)


MatteoVoges:
Came up in workgroup meeting: Compile objects can depend on each other. So we need a functionality to specify some dependencies among compile objects. First iteration --> boolean `contains_dependencies` in `parameters.kapitan.vars`, so that target will not split its compile objects.

PR #1119, opened 2024-01-19
Adds Python 3.12 to the test set

ademariag:
Fixes #

## Proposed Changes

*

## Docs and Tests

* [ ] Tests added
* [ ] Updated documentation


ademariag:
More work needed before we can use 3.12. dropping

MatteoVoges:
I think we just need to update some dependencies, so that no dependency uses `distutils` anymore. After that we wait for #964 so that we don't use `distutils` ourselves.

PR #1125, opened 2024-01-26
chore: fix linter

MatteoVoges:
Fixes lint job


MatteoVoges:
What about not using the latest version of `black`? This would prevent unexpected random ci failures.

ademariag (approved):

PR #1126, opened 2024-01-27
Simplify populating each target's classes list in `ReclassInventory`

simu:
There's no need to reconstruct the list of classes for each target. Reclass provides this information in `rendered_target["classes"]` already.

This commit removes the loop which reconstructs the list and replaces it by assigning `rendered_target["classes"]` to the target's `classes` field.

Side-note: the previous implementation also mangled the list, because Python implements `+=` for lists as list concatenation, and (not so helpfully) treats strings as lists of characters when they appear in a context where a list is expected. To correctly construct the list of classes with the previous approach, the `+=` should be replaced with append:

```python
self.targets[target_name].classes.append(class_name)
```

## Docs and Tests

* [ ] Tests added
* [ ] Updated documentation


MatteoVoges (approved):
I was not aware of reclass storing this already in the target. Thanks 😃

PR #987, opened 2023-04-06
feat(dependencies): git dependencies allow submodules

MatteoVoges:
Fixes #914 

## Proposed Changes
- Kapitan dependencies with `type: git` allow checkout with submodules now.
- e.g. 
- ```yaml
  ---
  parameters: 
    kapitan:
      dependencies:
         - type: git
           source: https://github.com/kapicorp/kapitan.git
           submodules: true 
           output_path: output/directory
  ```
  fetches `kapicorp/kapitan` with its submodule `kapicorp/reclass`

## To Do
* [X] Add documentation
* [X] Add a test example

--- 
### Contributed by  [<img src="https://www.nexenio.com/wp-content/uploads/2021/10/001-nexenio-logo-white-rgb@2x-e1636042495927.png" height="15em" />](https://www.nexenio.com)

Issue #914, opened 2022-12-23
Git dependency and git modules 


metacoma:
**Describe the bug/feature**
As i can understand, at this moment there is no way to clone git repository with git submodules?

**To Reproduce**
1. define dependency, for example, `kapicorp/kapitan` git repository
```yaml
---
parameters: 
  kapitan:
    dependencies:
       - type: git 
         output_path: files/git/github/kapicorp/kapitan
         source: https://github.com/kapicorp/kapitan
         ref: master
```
2.
 ```bash
$ kapitan compile --fetch
```

3. Try to open `.gitmodules` file https://github.com/kapicorp/kapitan/blob/master/.gitmodules
```bash
$ ls -ltr files/git/github/kapicorp/kapitan/.gitmodule
ls: cannot access 'files/git/github/kapicorp/kapitan/.gitmodule': No such file or directory
```

**Expected behavior**
`kapicorp/kapitan` github repository must be cloned with git modules (in this case with https://github.com/kapicorp/reclass.git)

**Additional context**
As far as i understand, `kapitan` uses GitPython library to clone git repos. Probably, the properly submodule handling in GitPython will helps https://gitpython.readthedocs.io/en/0.3.3/tutorial.html#submodule-handling

Also, submodule option might be added into dependency definition, like
```yaml
---
parameters: 
  kapitan:
    dependencies:
       - type: git
         submodule: True 
         output_path: files/git/github/kapicorp/kapitan
         source: https://github.com/kapicorp/kapitan
         ref: master
```

PR #1133, opened 2024-02-04
Install reclass as module

ademariag:
Fixes #

## Proposed Changes

*

## Docs and Tests

* [ ] Tests added
* [ ] Updated documentation


MatteoVoges:
@ademariag Could you add the tests for submodules back again? Maybe we have to find a lightweight test-repo with submodules, that we can download. But just because the tests fail with kapitan, because we removed `reclass`, we should continue testing it with other repos. Thanks in advance

ademariag:
Why do we need to test the submodule capability? It made sense when reclass was sourced that way. Is there a real use case for it?

MatteoVoges:
> Why do we need to test the submodule capability? It made sense when reclass was sourced that way. Is there a real use case for it?

It was a feature and is documented. See #914 #987 and https://kapitan.dev/pages/external_dependencies/?h=submodule#defining-dependencies

PR #1142, opened 2024-02-16
Expose `applications` and `exports` fields in `InventoryTarget` and Jsonnet `inventory()` callback

simu:
## Proposed Changes

* Restore the Kapitan <= 0.32 behavior of exposing the Reclass `applications`  (which Commodore depends on) and `exports` fields in the inventory

## Docs and Tests

* [x] Tests added
* [ ] Updated documentation

### DIFF
--- a/kapitan/inventory/inv_reclass.py
+++ b/kapitan/inventory/inv_reclass.py
@@ -14,7 +14,6 @@ logger = logging.getLogger(__name__)
 
 
 class ReclassInventory(Inventory):
-    
     def render_targets(self, targets: list = None, ignore_class_notfound: bool = False):
         """
         Runs a reclass inventory in inventory_path
@@ -28,7 +27,7 @@ class ReclassInventory(Inventory):
         reclass_config = get_reclass_config(self.inventory_path)
         reclass_config.setdefault("ignore_class_notfound", ignore_class_notfound)
         reclass_config["compose_node_name"] = self.compose_target_name
-        
+
         try:
             storage = reclass.get_storage(
                 reclass_config["storage_type"],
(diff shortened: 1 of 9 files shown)


kapitan/resources.py
```diff
@@ -278,7 +278,11 @@ def inventory(search_paths: list, target_name: str = None, inventory_path: str =
 
     if target_name:
         target = inv.get_target(target_name)
-        return {"parameters": target.parameters, "classes": target.classes}
+        return {
+            "parameters": target.parameters,
+            "classes": target.classes,
+            "applications": target.applications,
+        }
```
MatteoVoges, line 285: I think at this point we can simply return `dataclasses.asdict(target)`

MatteoVoges (approved):
Thank you for the fix. Do you know of any more special keys in reclass, that perhaps need to be exposed?

kapitan/inventory/inv_reclass.py
```diff
@@ -44,6 +43,8 @@ def render_targets(self, targets: list = None, ignore_class_notfound: bool = Fal
             for target_name, rendered_target in rendered_inventory["nodes"].items():
                 self.targets[target_name].parameters = rendered_target["parameters"]
                 self.targets[target_name].classes = rendered_target["classes"]
+                self.targets[target_name].applications = rendered_target["applications"]
```
ademariag, line 46: Could we not also copy all key/values in `rendered_target`? why are we hardcoding specific keys?

ademariag: I think I see why, it is required for the keys to be known. ok

ademariag (approved):

kapitan/inventory/inv_reclass.py
simu: I have no hard opinion here. I didn't find a helper for dataclasses which generates a dataclass from a dict (although I didn't search for very long), so I ended up just copying the fields.

PR #1189, opened 2024-06-26
feat:  Add a --no-reveal flag in compile command to avoid revealing refs

pbrissaud:
## Proposed Changes

* Add a `--no-reveal` flag to compile command to avoid revealing secret (useful if a `reveal: true` is set in the `.kapitan` config file) 

## Docs and Tests

* [x] Tests added
* [] Updated documentation


pbrissaud:
I am not sure if the test I made is good, but I tried 😆

ademariag (approved):

PR #1232, opened 2024-09-03
Fix `fetch_git_dependency()` to clobber files in the destination if `force_fetch=True`

simu:
## Proposed Changes

When force fetching Git dependencies, we need to be able to overwrite read-only files in the destination, since Git repos may contain read-only pack and index files in `.git/`.

To do so, this PR introduces flag `clobber_files` for `utils.copy_tree()` which will completely clobber existing files in the
destination by setting a custom `copy_function` for `shutil.copytree()` which deletes existing destination files before copying them with `shutil.copy2()`.

With this change, `utils.copy_tree()` with `clobber_files=True` behaves pretty much identically to distutils' `copy_tree()` which was used in previous versions of Kapitan. 

## TODO

- [x] Rebase once #1231 is merged

## Docs and Tests

* [x] Tests added
* [x] Updated documentation

### DIFF
--- a/kapitan/dependency_manager/base.py
+++ b/kapitan/dependency_manager/base.py
@@ -129,7 +129,9 @@ def fetch_git_dependency(dep_mapping, save_dir, force, item_type="Dependency"):
                     "{} {}: subdir {} not found in repo".format(item_type, source, sub_dir)
                 )
         if force:
-            copied = copy_tree(copy_src_path, output_path)
+            # We need `clobber_files=True` here, since we otherwise can't overwrite read-only Git
+            # index and pack files if the destination already contains a copy of the Git repo.
+            copied = copy_tree(copy_src_path, output_path, clobber_files=True)
         else:
             copied = safe_copy_tree(copy_src_path, output_path)
         if copied:
--- a/kapitan/utils.py
+++ b/kapitan/utils.py
@@ -628,10 +628,27 @@ def safe_copy_tree(src, dst):
     return outputs
 
 
-def copy_tree(src: str, dst: str) -> list:
+def force_copy_file(src: str, dst: str, *args, **kwargs):
+    """Copy file from `src` to `dst`, forcibly replacing `dst` if it's a file, but preserving the
+    source file's metadata.
+
+    This is suitable to use as `copy_function` in `shutil.copytree()` if the behavior of distutils'
+    `copy_tree` should be mimicked as closely as possible.
+    """
+    if os.path.isfile(dst):
+        os.unlink(dst)
+    shutil.copy2(src, dst, *args, **kwargs)
+
+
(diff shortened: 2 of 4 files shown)


kapitan/dependency_manager/base.py
```diff
@@ -129,7 +129,9 @@ def fetch_git_dependency(dep_mapping, save_dir, force, item_type="Dependency"):
                     "{} {}: subdir {} not found in repo".format(item_type, source, sub_dir)
                 )
         if force:
-            copied = copy_tree(copy_src_path, output_path)
+            # We need `clobber_files=True` here, since we otherwise can't overwrite read-only Git
+            # index and pack files if the destination already contains a copy of the Git repo.
+            copied = copy_tree(copy_src_path, output_path, clobber_files=True)
```
ademariag, line 134: I wonder if we should check that a user is not providing a system directory outside of the kapitan path

simu: I think it would make sense to restrict dependency output directories to Kapitan's working directory, but I'd check this for all dependencies, and not just Git repos (probably best to do it when collecting all dependencies that need to be fetched)

simu: We could do something like the following to raise an Exception if a dependency has an absolute output path which points to outside Kapitan's `output_path` (this would probably need a better Exception type).

```diff
diff --git a/kapitan/dependency_manager/base.py b/kapitan/dependency_manager/base.py
index d19678c9..53702c22 100644
--- a/kapitan/dependency_manager/base.py
+++ b/kapitan/dependency_manager/base.py
@@ -42,6 +42,9 @@ def fetch_dependencies(output_path, target_objs, save_dir, force, pool):
     http_deps = defaultdict(list)
     helm_deps = defaultdict(list)
 
+    # make sure base output_path is normalized and an absolute path
+    output_path = os.path.abspath(os.path.normpath(output_path))
+
     # this dict is to make sure no duplicated output_paths exist per source
     deps_output_paths = defaultdict(set)
     for target_obj in target_objs:
@@ -54,6 +57,10 @@ def fetch_dependencies(output_path, target_objs, save_dir, force, pool):
                 # The target key "output_path" is relative to the compile output path set by the user
                 # point to the full output path
                 full_output_path = normalise_join_path(output_path, item["output_path"])
+                if not full_output_path.startswith(output_path):
+                    raise ValueError(
+                        f"Dependency output path pointing outside Kapitan output_path: {full_output_path}"
+                    )
                 logger.debug("Updated output_path from %s to %s", item["output_path"], output_path)
                 item["output_path"] = full_output_path
```

ademariag: Will handle in a different PR

ademariag (approved):

PR #1249, opened 2024-10-17
Allow legacy VaultKV configuration keys in `kapitan.secrets.vaultkv`

simu:
## Proposed Changes

* Add validation aliases which allow the old envvar style configuration for the VaultKV references backend config through the inventory
* Update the vaultkv documentation to note that the envvar style configs in the inventory are deprecated.

## TODO

* [x] Document limitation regarding duplicated configs through different field aliases. Due to some implementation details in pydantic-settings, having e.g. the Vault address set both through environment variable `VAULT_ADDR` and initializer field `addr` will cause a validation error unless `extra` is set to `ignore` or `allow` neither of which have the desired effect of protecting users against config errors in the inventory.

## Docs and Tests

* [x] Tests added
* [x] Updated documentation


simu:
Behavior which causes the validation error when providing `VAULT_ADDR` in the environment while `addr` is set in the refs file is documented in https://github.com/pydantic/pydantic-settings/issues/245

ademariag (approved):

PR #1258, opened 2024-10-30
#1257 Kapitan does not prune Kadet input

### LINKED ISSUE #1257: [bug]: KapitanInputTypeKadetConfig inherits prune from KapitanInputTypeBaseConfig
### Description of the bug

After adding Pydantic types for Inventory, the assumption was introduced that Kadet Input in Kapitan should always prune objects before serialization to a file. 
However, there are many cases for which this is not the case, such as [Flux's ImageRepository](https://fluxcd.io/flux/components/image/imagerepositories/#:~:text=To%20grant%20access%20to%20all%20namespaces%2C%20an%20empty%20matchLabels%20can%20be%20set%3A), where it is expected that the object is empty.

### Steps to reproduce

```bash
1. Use Kadet as input for an object with an empty dictionary as a value to a key.
2. Run `kapitan compile`.
3. See that neither the key nor value appears in the compiled directory.
```

### What happened?

The key (with an empty dict as a value) should be in the compiled directory.

### Additional context

This was introduced in [PR #1234](https://github.com/kapicorp/kapitan/pull/1234). Specifically in [line 36 of kapitan/inventory/model/input_types.py](https://github.com/kapicorp/kapitan/blob/master/kapitan/inventory/model/input_types.py#L36).

### Version (kapitan)

master

### Version (kapitan)

I use kapitan directly (pip, docker)

- [ ] pyenv
- [ ] venv
- [ ] no environment

### OS information

_No response_

### Relevant log output

_No response_

JordanLloydHall:
Fixes #1257

## Proposed Changes

* `KapitanInputTypeKadetConfig.prune` defaults to `False` instead of `True` (inherited).
* Type hint for `KapitanInputTypeJsonnetConfig.prune` is consistent with `KapitanInputTypeBaseConfig.prune`.

### DIFF
--- a/kapitan/inventory/model/input_types.py
+++ b/kapitan/inventory/model/input_types.py
@@ -39,7 +39,7 @@ class KapitanInputTypeBaseConfig(BaseModel):
 class KapitanInputTypeJsonnetConfig(KapitanInputTypeBaseConfig):
     input_type: Literal[InputTypes.JSONNET] = InputTypes.JSONNET
     input_paths: List[str]
-    prune: bool = False
+    prune: Optional[bool] = False
 
 
 class KapitanInputTypeExternalConfig(KapitanInputTypeBaseConfig):
@@ -79,6 +79,7 @@ class KapitanInputTypeKadetConfig(KapitanInputTypeBaseConfig):
     output_type: OutputType = OutputType.YAML
     input_paths: List[str]
     input_value: Optional[dict] = None
+    prune: Optional[bool] = False
 
 
 class KapitanInputTypeRemoveConfig(KapitanInputTypeBaseConfig):


ademariag (approved):

kapitan/inventory/model/input_types.py
```diff
@@ -39,7 +39,7 @@ class KapitanInputTypeBaseConfig(BaseModel):
 class KapitanInputTypeJsonnetConfig(KapitanInputTypeBaseConfig):
     input_type: Literal[InputTypes.JSONNET] = InputTypes.JSONNET
     input_paths: List[str]
-    prune: bool = False
+    prune: Optional[bool] = False
```
gburiola, line 42: Hey @alledm.

What do you think of making this default to False in the baseconfig on line 36?

I also noticed that configuring this in the inventory (`prune: false`) config does not work

```
parameters:
  kapitan:
    compile:

    - output_path: manifests
      input_type: kadet
      output_type: yaml
      prune: false
```

ademariag: Yes makes sense. this is something that didn't notice. will fix it soon

JordanLloydHall: Ideally this should be default `None` as the setting precedence seems to be:
1. This class' `prune` attribute. 
2. The `prune` cmd/dot setting.
3. Default `False`.

Although while this key exists in the class at all, neither the cmd nor the dot files will do anything. This is because whenever we check if the `prune` key exists rather than if an attempt to get it's value returns `None` at the relevant call sites.

PR #1319, opened 2025-06-03
Implement kadet input type cache

### LINKED ISSUE #1318: [feat]: Cache unchanged compiled items to disk
### Description of the feature

Kapitan should have a mechanism to cache repeatedly compiled items that haven't changed to reduce compilation times

### Additional context

_No response_

ramaro:
Fixes #1318 

## Proposed Changes

* Introduces a generic caching mechanism for Kapitan inputs. Its main purpose is to store and retrieve compiled objects to avoid re-compiling them on every run, thus improving performance.
* As part of the caching mechanism via the InputCache class, a file-based cache is available as a content addressable storage mechanism with the intention of being generic and good enough for other input types.
* InputCache assumes all compiled objects are "pickleable" as the implementation uses Python's pickle library.
* Removes the old cache implementation as it was not tested, valid nor maintained.
* The cache is only enabled when the `--cache` / `-c` flags are set in the `compile` command.

### Notes
* Compiled objects will still be inspected at compile time to reveal references
* This implementation assumes that either `XDG_CACHE_HOME` or `HOME` env vars are set as the cache home so that the cache artifacts can be written to `"{cache_home}/.cache/kapitan/{input_type_name}"` 

## Docs and Tests

* [x] Tests added
* [ ] Updated documentation

### DIFF
--- a/docs/pages/commands/kapitan_compile.md
+++ b/docs/pages/commands/kapitan_compile.md
@@ -194,9 +194,8 @@ The `--embed-refs` flags tells **Kapitan** to embed these references on compile,
           --embed-refs          embed ref contents
           --inventory-path INVENTORY_PATH
                                 set inventory path, default is "./inventory"
-          --cache, -c           enable compilation caching to .kapitan_cache and
-                                dependency caching to .dependency_cache, default is
-                                False
+          --cache, -c           enable compilation caching to $XDG_CACHE_HOME/kapitan,
+                                default is False [this is EXPERIMENTAL]
           --cache-paths PATH [PATH ...]
                                 cache additional paths to .kapitan_cache, default is
                                 []


kapitan/inputs/kadet.py
```diff
@@ -102,34 +116,47 @@ def compile_file(self, config: KapitanInputTypeKadetConfig, input_path, compile_
         target_name = self.target_name
```
rjmco, line 116: I would suggest changing slightly the initial line of this function's docstring from:
```
Compile a kadet input file.
```
to
```
Compile a kadet input file, if resulting compilation is not already cached
```

kapitan/inputs/cache.py
```diff
@@ -0,0 +1,80 @@
...
+            # dont write if already exists
+            if cached_path.exists():
+                return inputs_hash
+
+            if not cached_path_lock.exists():
+                sub_path.mkdir(parents=True, exist_ok=True)
+                with open(cached_path_lock, "wb") as fp:
+                    logger.debug("Writing cache file: %s", cached_path_lock)
+                    self.dump_output(output_obj, fp)
+                cached_path_lock.rename(Path(str(cached_path_lock).removesuffix(".lock")))
```
rjmco, line 56: What if for some reason, someone aborts execution and this step is never executed? Should there be a recommendation to remove the `.lock`? Would it be to:
1. Delete the whole cache?
2. Delete just the `.lock` file?
3. Remove the `.locl` suffix?

kapitan/cli.py
```diff
@@ -264,19 +268,10 @@ def build_parser():
...
         "--cache",
         "-c",
-        help="enable compilation caching to .kapitan_cache\
-        and dependency caching to .dependency_cache, default is False",
+        help="enable compilation caching to $XDG_CACHE_HOME/kapitan, default is False",
         action="store_true",
         default=from_dot_kapitan("compile", "cache", False),
     )
-    compile_parser.add_argument(
-        "--cache-paths",
```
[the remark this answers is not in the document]
ramaro: no because it was dead code 😅

PR #1329, opened 2025-08-02
add inventory-pool-cache option

bartba-isolabs:
## Proposed Changes

* Adds an option to skip passing inventory cache dictionary to process pool
   * For large inventories, passing this dictionary introduces a significant overhead that increases the compile time

## Docs and Tests

* [ ] Tests added
* [ ] Updated documentation


bartba-isolabs:
On further tests, I can't get `compile_target` method to use the `globals_cached` param at all.
[This line](https://github.com/kapicorp/kapitan/blob/master/kapitan/targets.py#L255) appears to always evaluate to `False` because of `cached.inv` already being populated (even for a new process in the pool).

I think the best option would be to remove this param altogether (and remove the `inventory-pool-cache` toggle I added) @ademariag.

bartba-isolabs:
> For my own understanding, what does the inventory pool cache do normally and why does it slow down compile times so much for large inventories?

My understanding is that it's supposed to speed up compilation time for the new processes in the pool. On further inspection, it does not seem to ever be used actually (see my previous comment).

As to why does it slow the compile so much: the parameters you pass on to a process need to be serialised/deserialised @andrewnicolalde .

PR #1338, opened 2025-09-01
Support for Python 3.13

simu:
Now that we've updated to (go)jsonnet 0.21, there's no remaining blockers for supporting Python 3.13. I've updated the references to supported Python versions that I found when grepping for `3.12`.

## Proposed Changes

* Bump maximum supported Python version to 3.13
* Update GitHub actions to cover Python 3.13
* Update bug report issue template supported Python versions

## Docs and Tests

* [ ] Tests added
* [x] Updated documentation

### DIFF
--- a/.github/workflows/pex-build-upload.yml
+++ b/.github/workflows/pex-build-upload.yml
@@ -26,6 +26,7 @@ jobs:
             3.10
             3.11
             3.12
+            3.13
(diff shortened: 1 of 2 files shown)


simu:
cc @ademariag @ramaro (I can't request reviews from you guys, so I'm doing it the old fashioned way :wink:)

PR #1345, opened 2025-09-17
Remove unused run dependencies

timp87:
Fixes https://github.com/kapicorp/kapitan/issues/1344

## Proposed Changes

* remove unused run dependencies

## Docs and Tests

* [ ] Tests added
* [ ] Updated documentation


ademariag:
> @ademariag does this look OK for you?

Looks ok, but now that I think about it, one of the reasons why we would have extra dependencies is in case you run code (i.e. generators) that might depend on it. We don't have a clean way to install dependencies based on the code we run. but perhaps we need to look into it

Will test and let you know

PR #1411, opened 2026-03-03
feat: use mise for external dependencies

jjmaestro:
## Goal

Standardize the repository on `mise` to manage and install external tools while keeping `uv` as the Python dependency workflow.

## Changes

This branch:

- introduces `mise.toml` and `mise.lock` as the source of truth for external tool versions
- rewrites `Makefile` setup and test commands to run through `mise exec`
- updates CI workflows to use `mise` instead of hand-installed tools
- updates contributor docs to describe the `mise + uv` workflow
- removes the stale `default.nix` shell

### Why remove `default.nix`

`default.nix` referred to stale tooling like `poetry`, `black`, and a `pip`-prefix setup that no longer matched the repository. It was no longer part of the documented or automated workflow. Also, it seemed like it was added back by mistake: it first appeared in ec94fc37, was adjusted in ec93dd64 and c2a32354, and then reappeared in 65e98654 in #1319 even though nothing in the current Makefile, CI, README, or contributor docs referenced it.

## Final state

After this branch:

- external tools are pinned in one place
- local development and CI use the same tool bootstrap path
- `Makefile` commands no longer carry bespoke installer logic
- contributor setup is simpler to explain and reproduce


ademariag:
Sorry for the delay @jjmaestro ! WIll try to review the other changes as well.

ademariag (approved):

PR #1412, opened 2026-03-03
fix: preview docs

jjmaestro:
> [!NOTE]
> Stacked on top of #1411

## Goal

Fix the generation of local docs and make previewing docs easier to use when the default local port is already occupied.

## Changes

This branch:

- fixes `make docs_serve`
- adds `DOCS_DEV_ADDR ` to `Makefile`, allowing easy address override
- updates the command output to show the effective preview address
- refreshes contributor docs to describe `make docs_serve`
- documents how to override the preview address from the command line


ademariag (changes requested):
Please reinstate default.nix

jjmaestro:
@ademariag but the deletion was done in #1411, I thought it was clear that all of the `default.nix` stuff was stale. I do have to update this PR since all of the `mise` foundation came in that PR, though. I'll do that today!

PR #1436, opened 2026-03-24
fix: replace bare except with except Exception

harshadkhetpal:
## Summary

Replace bare `except:` clauses with `except Exception:` in `base64.py` and `cli.py`.

Bare `except:` catches all exceptions including `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`, which is almost never intended. Using `except Exception:` follows PEP 8 best practices.

**Changes:**
```python
# Before (kapitan/refs/base64.py:48)
except:
    return self.data

# After
except Exception:
    return self.data

# Before (kapitan/cli.py:84)
except:
    sys.exit(1)

# After
except Exception:
    sys.exit(1)
```

## Testing
No behavior change for normal exceptions — style/correctness fix only.

### DIFF
--- a/kapitan/cli.py
+++ b/kapitan/cli.py
@@ -81,7 +81,7 @@ def trigger_compile(args):
             ref_controller=ref_controller,
             args=args,
         )
-    except:
+    except Exception:
         sys.exit(1)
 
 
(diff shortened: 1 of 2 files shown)


Moep90:
Superseded by #1461 which already fixed bare except violations across 19 files (including the two files changed here).

PR #1445, opened 2026-04-24
feat: add literal resolver to emit backslash dollar syntax in OmegaConf

Moep90:
## [feat]: Repurpose `escape` resolver to emit literal `${content}` in OmegaConf

---

### Summary

Repurposes the existing `${escape:content}` resolver in the OmegaConf inventory backend to emit
`${content}` as a verbatim string in the final compiled output — bypassing both OmegaConf
resolution passes entirely.

This is useful when a target needs to produce a literal `${...}` expression in its output (e.g.
a Terraform resource reference or a shell variable) without Kapitan attempting to resolve it as
an inventory interpolation.

```yaml
parameters:
  google_sa_email: ${escape:google_service_account.cluster.email}
  # compiled value: ${google_service_account.cluster.email}

  shell_home: ${escape:HOME}
  # compiled value: ${HOME}

  # Works embedded in a larger string too
  greeting: "Hello ${escape:USER}, welcome to ${cluster}"
  # compiled value: Hello ${USER}, welcome to my-cluster
```

---

#### Implementation

- **`kapitan/inventory/backends/omegaconf/resolvers.py`**
  - `escape_interpolation(content)` now returns content wrapped in opaque `LITERAL_MARKER_*`
    markers instead of a raw `${content}` string — this ensures the value is never picked up and
    re-resolved during the second OmegaConf pass, even when `content` matches a real inventory key
  - `process_literals()` already handled these markers; no change needed there
  - Removed the now-superseded `literal_interpolation` function and its `LITERAL_BACKSLASH_MARKER_*`
    constants and `_COMBINED_LITERAL_PATTERN` / `_replace_literal` helpers
  - Removed the `literal` resolver registration from `register_resolvers()`

#### How it works

The OmegaConf backend uses a two-pass resolution strategy. To produce a final value that contains
`${...}`, a resolver must return a string that:

1. Is not itself an OmegaConf interpolation (so it survives both passes unchanged)
2. Can be post-processed after resolution is complete

`escape_interpolation` wraps `content` in opaque marker strings. After both OmegaConf passes run,
`process_literals()` substitutes the markers for the final `${content}` value.

---

### Docs and Tests

- [x] **Test updated** — `test_escape_resolver_produces_dollar_syntax` in `tests/test_omegaconf.py` covers:
  - Standalone values (`${escape:HOME}` → `${HOME}`)
  - Embedded mid-string use (`Hello ${escape:USER}, welcome to ${cluster}` resolves surrounding interpolations but not the escaped one)
  - Key-collision protection — `${escape:environment}` produces `${environment}` even though `environment` is a real inventory key that would otherwise resolve to `dev`
- [x] **Test inventory updated** — `tests/test_resources/omegaconf/inventory/classes/components/test-resolvers.yml`
  `escape_test` section extended with the mid-string and key-collision cases
- [x] **Docs updated** — `docs/pages/inventory/omegaconf.md` Utility Resolvers table entry for
  `escape` updated to reflect the new behaviour

### DIFF
--- a/docs/pages/inventory/omegaconf.md
+++ b/docs/pages/inventory/omegaconf.md
@@ -178,6 +178,7 @@ deployments:
 | Resolver | Description | Example |
 |----------|-------------|---------|
 | `escape` | Create an interpolation for next resolution pass | `${escape:some.path}` → `${some.path}` |
+| `literal` | Emit a literal `\${content}` string (never resolved) | `${literal:HOME}` → `\${HOME}` |
 | `default` | Chain of fallback values using oc.select | `${default:key1, key2, fallback}` |
 | `write` | Write resolved content to another inventory location | `${write:destination.path, source.path}` |
 | `from_file` | Read content from a file | `${from_file:path/to/file.txt}` |
--- a/docs/pages/inventory/omegaconf.md
+++ b/docs/pages/inventory/omegaconf.md
@@ -177,8 +177,7 @@ deployments:
 
 | Resolver | Description | Example |
 |----------|-------------|---------|
-| `escape` | Create an interpolation for next resolution pass | `${escape:some.path}` → `${some.path}` |
-| `literal` | Emit a literal `\${content}` string (never resolved) | `${literal:HOME}` → `\${HOME}` |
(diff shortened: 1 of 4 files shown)


ramaro (approved):
thank you for the fix!

PR #1451, opened 2026-04-24
feat: allow git dependencies to be pinned to a tag or commit SHA

Moep90:
## Summary

The `ref` field on git dependencies previously only documented branch names. This PR clarifies that it accepts any git-resolvable value — branch name, tag, or full/short commit SHA — and adds a guard for `ref=None` that was previously an unconditional call which would fail.

### `kapitan/dependency_manager/base.py`

- **Guard `ref=None`**: replaced unconditional `repo.git.checkout(dep.ref)` with `if dep.ref: repo.git.checkout(dep.ref)`. When `ref` is omitted the cloned HEAD is used as-is.
- **Reuse `Repo` object**: moved `Repo(cached_repo_path)` outside the per-dep loop — one object opened once per source rather than once per dep.
- **Simplify `exists_in_cache`**: replaced O(n) `os.listdir` scan with a single `os.path.exists` call.
- **Update docstring**: removed stale "fetches master branch by default" phrasing.

### `tests/conftest.py`

- Added `tagged_git_repo` fixture: an in-memory git repo with two commits where `v1.0.0` is tagged on the first commit. Exposes `tagged_git_repo`, `tagged_commit_sha`, and `git_tag_name` on the test instance via `_attach_fixture`.

### `tests/test_dependency_manager.py`

- Added `GitRefPinningFetchTest` with three integration tests:
  - `test_fetch_git_ref_as_tag` — `ref=<tag>` checks out the tagged commit
  - `test_fetch_git_ref_as_commit_sha` — `ref=<sha>` checks out that exact commit
  - `test_fetch_git_ref_none_stays_on_head` — `ref=None` stays on HEAD without error

### `docs/pages/external_dependencies.md`

- Updated the `ref` line in the syntax reference: `"branch, tag, or commit SHA (optional)"`.
- Expanded annotation #3 to explain all three ref types and the `master` default.

## Backward Compatibility

`ref` defaults to `"master"` (unchanged). Any existing inventory using `ref: master` or omitting `ref` entirely continues to work without modification.

## Testing

```
poetry run pytest tests/test_dependency_manager.py --no-cov -q
# 11 passed
```

### DIFF
--- a/docs/pages/external_dependencies.md
+++ b/docs/pages/external_dependencies.md
@@ -83,13 +83,13 @@ Kapitan also supports caching Use the `--cache` flag to cache the fetched items 
           output_path: path/to/dir
           source: git_url # mkdocs (1)!
           subdir: relative/path/from/repo/root (optional) # mkdocs (2)!
-          ref: tag, commit, branch etc. (optional) # mkdocs (3)!
+          ref: branch, tag, or commit SHA (optional) # mkdocs (3)!
           submodules: true/false (optional) # mkdocs (4)!
     ```
 
     1. Git types can fetch external `git` repositories through either HTTP/HTTPS or SSH URLs.
     2. Optional supports for cloning just a sub-directory
-    3. Optional support for accessing them in specific commits and branches (refs).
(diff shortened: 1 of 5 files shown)


ramaro (approved):

ramaro:
@Moep90 thank you!

PR #1453, opened 2026-05-08
chore(release): add supply-chain metadata (CycloneDX SBOMs and SHA256 checksums)

Moep90:
# Summary

Adds software supply-chain metadata to every Kapitan release:

- **CycloneDX SBOM** (Python package) attached to PyPI releases
- **CycloneDX SBOM** (Docker image, one per Python × platform variant) attached to Docker releases
- **SHA256SUMS** manifest covering the wheel, sdist, and Python SBOM
- **SHA256SUMS.pex** for the standalone PEX binary

All artifacts are uploaded to the GitHub release page and can be independently verified without fetching the packages themselves.

---

# Changed files

| File | Change |
|------|--------|
| `scripts/generate_sbom.sh` | New script — exports production+extras deps via `uv`, generates a CycloneDX JSON SBOM with `cyclonedx-py` |
| `scripts/generate_checksums.sh` | New script — hashes wheel + sdist + SBOM into a `sha256sum`-compatible manifest |
| `scripts/validate_sbom.sh` | New script — validates `bomFormat == CycloneDX` and reports component count; shared by both SBOM generation paths |
| `.github/workflows/python-pypi-publish.yml` | Adds SBOM + checksum generation and uploads them alongside the wheel/sdist on release |
| `.github/workflows/pex-build-upload.yml` | Adds `SHA256SUMS.pex` for the PEX binary; adds `contents: write` permission; fixes `workflow_run` trigger name |
| `.pre-commit-config.yaml` | Adds `shellcheck-py` hook (`--severity=warning`) for all shell scripts |

---

# Design decisions

## SBOM scope: production + all optional extras

`uv export` is called with `--all-extras --no-group test --no-group docs` so the SBOM reflects the maximal install surface — including `gojsonnet`, `omegaconf`, and `reclass-rs` — not just the minimal base set.  
The `dev` group is excluded implicitly by `--no-dev`.

## `--pyproject` intentionally omitted

`cyclonedx-bom` 4.x `--pyproject` flag crashes with `'str' object has no attribute 'get'` (observed on 4.6.1). The `requirements` sub-command is used instead. Re-evaluate when bumping to `cyclonedx-bom>=5`.

## Comment and self-reference stripping

`uv export` emits a human-readable header block (`# This file was autogenerated…`) and a bare `.` self-reference from the editable install even when `--no-editable` is set. Both cause `cyclonedx-py` to log spurious warnings. They are stripped with `grep -Ev '^\s*(#|\.\s*$|$)'` before the requirements are passed to `cyclonedx-py`.

## Version extraction without an installed package

After `uv build` the built wheel is present in `dist/` but kapitan is **not** installed in the runner Python environment, so `importlib.metadata.version('kapitan')` raises `PackageNotFoundError`. Version is extracted via `${GITHUB_REF_NAME#v}` on release events and by parsing the wheel filename on `workflow_dispatch`.

## Reproducible output

`--output-reproducible` is passed to `cyclonedx-py` to strip timestamps and random UUIDs, enabling byte-for-byte identical SBOMs for the same dependency set.

## Filename convention

| Artifact | Pattern |
|----------|---------|
| Python SBOM | `kapitan-{VERSION}.python.sbom.cdx.json` |
| Docker SBOM | `kapitan-{VERSION}-py{PY}-{PLATFORM}-docker.sbom.cdx.json` |
| Python checksum manifest | `SHA256SUMS` |
| PEX checksum manifest | `SHA256SUMS.pex` |

The `.python.` infix distinguishes the package-level SBOM from the per-platform Docker image SBOMs.  
Platform slugs (`linux-amd64`, `linux-arm64`) are derived from `matrix.platform.platform` at runtime, not hardcoded, so adding a third platform requires no workflow changes.

---

# Security considerations

- Third-party GitHub Actions are pinned to full commit SHAs throughout.
- `contents: write` permission is scoped only to the jobs that upload release assets; the `build` matrix job holds `contents: read`.
- Docker SBOM steps are gated on `github.event_name != 'pull_request'` to prevent untrusted code from triggering PyPI-adjacent dependency resolution.
- The `upload-docker-sboms` job verifies that at least one SBOM was downloaded before attempting a release upload, failing fast if artifacts have expired (1-day retention).

---

# Testing

All three shell scripts pass `bash -n` syntax check.  
All three workflow YAML files parse cleanly with `python3 -c "import yaml; yaml.safe_load(...)"`.  
Local end-to-end run of `generate_sbom.sh 0.35.1 /tmp/sbom-test`:

```
Exported 87 dependency lines
SBOM written to: /tmp/sbom-test/kapitan-0.35.1.python.sbom.cdx.json
SBOM validation: OK (file=kapitan-0.35.1.python.sbom.cdx.json, bomFormat=CycloneDX, components=87)
```

`sha256sum -c SHA256SUMS` → all entries `OK`.

---

# Checklist

- [x] Shell scripts are executable (`100755` in git index)
- [x] SBOM covers all optional extras (`gojsonnet`, `omegaconf`, `reclass-rs`)
- [x] No `cyclonedx-py` warnings in SBOM generation output
- [x] `sha256sum -c SHA256SUMS` passes locally
- [x] Version extraction works for both `release` and `workflow_dispatch` triggers
- [x] Platform slug derived at runtime — no ternary hardcoding
- [x] `upload-docker-sboms` job uses pinned runner from `setup` output
- [x] All third-party actions pinned to commit SHA
- [x] No `contents: write` on PR-accessible jobs


.github/workflows/pex-build-upload.yml
```diff
@@ -5,13 +5,15 @@ on:
...
+      - Publish Python Package to PyPI
     types:
       - completed
 
 jobs:
   pex_build_publish:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
```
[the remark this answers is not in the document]
Moep90: Splitting pypi-publish into separate upload jobs would add artifact-passing complexity for negligible security gain. The workflow only triggers on release/workflow_dispatch.

.github/workflows/python-pypi-publish.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   pypi-publish:
     runs-on: ubuntu-latest
     permissions:
-      contents: read
+      contents: write
```
ramaro, line 20: why do we need write here?

.github/workflows/pex-build-upload.yml
```diff
@@ -50,8 +69,20 @@ jobs:
             -o dist/kapitan.linux-x86_64.pex
           dist/kapitan.linux-x86_64.pex --help
 
+      - name: Generate PEX checksum
+        run: |
+          (cd dist && sha256sum kapitan.linux-x86_64.pex) > dist/SHA256SUMS.pex
+          echo "PEX checksum:"
+          cat dist/SHA256SUMS.pex
+          (cd dist && sha256sum -c SHA256SUMS.pex)
```
ramaro: should we rename`SHA256SUMS.pex` into something else without the `.pex` extension? Typically a `.pex` file is executable but here we just want the checksums, right?

.github/workflows/python-pypi-publish.yml
Moep90: Required for softprops/action-gh-release to upload PEX + checksum
assets to the GitHub release page. Only actually used when the tag
upload condition below evaluates to true (release tags only).

ademariag (approved):

PR #1457, opened 2026-05-10
feat: wildcard class patterns in inventory `classes:` lists 

### LINKED ISSUE #1084: [feat]: load classes generically according to a given pattern
### Description of the feature

Reference: [Slack Thread](https://kubernetes.slack.com/archives/C981W2HD3/p1699355598600929)

Usage:
```yaml
classes:
  - * # (load all classes)
  - clusters.* # (all classes in clusters/ directory
  - dev-* # (all classes with prefix dev- (in all (sub)-directories)
```

We could extend this request by not just supporting `*`, but other symbols as well.

### Additional context

_No response_

Moep90:
# Summary

Allows `classes:` entries in target and class files to be glob patterns that
auto-expand to all matching class names discovered under `inventory/classes/`.

```yaml
classes:
  - common          # unchanged – exact names still work as before
  - clusters.*      # expands to clusters.dev, clusters.prod, … (sorted)
  - dev-*           # matches basename in any subdir: dev-common, apps.dev-api, …
  - "*"             # all classes (quote required to avoid YAML ambiguity)
```

Closes #1084.

---

# Motivation

Managing large inventories means maintaining long, fragile lists of explicit
class names. A pattern like `clusters.*` or `dev-*` auto-discovers all matching
classes and removes the need to update every target file when a new class is
added.

---

# Implementation

**Backend-agnostic pre-expansion.** When any YAML file under
`inventory/targets/` or `inventory/classes/` contains a wildcard `classes:`
entry, Kapitan materializes a temporary copy of the inventory tree with those
entries are pre-expanded before handing them to the chosen backend. This means the
feature works identically with **reclass**, **reclass-rs**, and **omegaconf** —
No per-backend code is needed.

The hot path is zero-cost: a cheap text-level scan skips full YAML parsing for
any file that contains no glob metacharacter (`*`, `?`, `[`), which is the
the overwhelming majority of files in a typical inventory.

**Pattern semantics** (implemented in `kapitan/inventory/wildcards.py`):

| Pattern | Matched against |
|---------|----------------|
| Pattern contains `.` (e.g., `clusters.*`) | Full dotted class name |
| Pattern with no `.` (e.g. `dev-*`) | Basename segment after last `.`, so it matches in all subdirs |

Expansion uses Python's `fnmatch`, which supports `*`, `?`, and `[…]`.

**Ordering and deduplication:** Each pattern expands to a lexicographically
sorted list inserted at the pattern's position. Later duplicates (across exact
entries and expansions) are silently dropped, preserving the first occurrence.

**Unmatched patterns:** raise `InventoryError` by default; silently dropped
when `--ignore-class-not-found` is set, mirroring exact-name behavior.

**`init.yml` convention:** `clusters/init.yml` maps to the class name
`clusters`, not `clusters.init` — consistent with reclass semantics.

**`migrate()` safety:** `Inventory.original_inventory_path` preserves the
user's on-disk path; `migrate()` always operates there, not on the temp copy.

---

# Files changed

| File | Change |
|------|--------|
| `kapitan/inventory/wildcards.py` | New module: `is_pattern`, `discover_classes`, `expand_class_patterns`, `materialize_expanded_inventory` |
| `kapitan/inventory/inventory.py` | Call `materialize_expanded_inventory` in `Inventory.__init__`; expose `original_inventory_path` |
| `kapitan/inventory/backends/omegaconf/__init__.py` | `migrate()` uses `original_inventory_path` |
| `examples/kubernetes/inventory/targets/all-glob.yml` | Demonstrate `cluster.*` and `component.*` wildcard usage |
| `docs/pages/inventory/classes.md` | New "Wildcard class patterns" section with examples, ordering rules, dedup, unmatched-pattern behavior |
| `tests/test_class_wildcards.py` | 40 tests: unit (pure helpers) + integration (reclass, reclass-rs, omegaconf) |

---

# Test plan

```bash
# Focused:
pytest tests/test_class_wildcards.py

# Broader regression:
pytest tests/test_inventory.py tests/test_omegaconf.py tests/test_compile.py

# Linting:
ruff check kapitan/inventory/wildcards.py kapitan/inventory/inventory.py
ruff format --check kapitan/inventory/wildcards.py tests/test_class_wildcards.py
```

**Results (this env):**
- `test_class_wildcards.py`: 40 passed, 1 skipped (omegaconf `.yaml`-loader pre-existing limitation, documented in test).
- `test_inventory.py`: 8/8.
- `test_omegaconf.py`: 19/19.
- `test_compile.py`: 17/17 (excluding 3 pre-existing failures unrelated to this PR).
- `ruff check`: clean.

---

# Limitations / known gaps

- The omegaconf backend's class-file resolver only looks for `.yml` files
  (pre-existing, unrelated to this PR). Wildcard expansion correctly
  *discovers* `.yaml` classes; loading them fails inside omegaconf.
  Tracked separately.

---

# Checklist

- [x] Works with reclass, reclass-rs, omegaconf
- [x] Exact class names unchanged
- [x] Deterministic ordering (lexicographic)
- [x] Duplicate deduplication
- [x] Unmatched pattern → `InventoryError` (or silent with `--ignore-class-not-found`)
- [x] Hidden files, non-YAML files excluded from expansion
- [x] `.yml` and `.yaml` both discovered
- [x] Wildcards in nested class files expanded (not just targets)
- [x] `migrate()` writes to the user's original inventory, not the temp copy
- [x] Zero overhead when no wildcards present
- [x] `ruff check` / `ruff format` clean
- [x] Docs updated
- [x] Tests cover all cases listed in the issue


BacLuc:
Thank you very much. This is what i am looking for.

simu:
> The hot path is zero-cost: a cheap text-level scan skips full YAML parsing for any file that contains no glob metacharacter (*, ?, [), which is the the overwhelming majority of files in a typical inventory.

Just FYI: for one of our representative inventories, roughly 30% of the class files contain one or more of the meta characters (since both `*` and `[` are valid and relatively commonly used YAML control characters and `?` is used for Kapitan secret references).

> Each pattern expands to a lexicographically sorted list inserted at the pattern's position.

While this is probably the best we can do to ensure deterministic output, IMO the documentation should be more explicit about the fact that adding a new class which matches an existing pattern may or may not have the desired effect depending on its name, i.e. config overrides may be ordered incorrectly unless all classes that match a pattern are named carefully to ensure correct include order (dumb example: a pattern matching `defaults.yml` and `config.yml` will include `config.yml` before `defaults.yml` which may or may not be the order that the user would like to have).

> Later duplicates (across exact entries and expansions) are silently dropped, preserving the first occurrence. 

Not sure about OmegaConf, but reclass already does this anyway when processing class includes.

simu:
> Pattern with no `.` (e.g. `dev-*`) are treated as basename segment after last `.`, so it matches in all subdirs

This seems somewhat unintuitive to me, I'd expect this to be a regular prefix match (i.e. matches all classes that start with `dev-`, including classes in directories which start with `dev-`).

From what I can see, all directory basename match should be trivially possible with `*dev-*` without any special handling and much clearer communication of intent in the user-visible class include (at the cost of having to quote this pattern to not make YAML parsing very unhappy).

Other things that I've found after playing with this for a bit longer:

* The feature doesn't work out of the box for inventories that use symlinks to pull in classes (unless the symlinks are explicitly generated with absolute paths as targets -- Commodore currently uses relative paths)
* This is a hard breaking change for users who use any of the metacharacters in their class names. Technically, nothing is stopping you from having a class `config[html].yml`, but this PR wouldn't include that class anymore because it'll treat `[html]` as a single character match for any of `h`, `t`, `m` or `l`.
* This feature breaks on inventories that want to use patterns and Reclass references in the same class include because the naive check for presence of at least one class for the expanded pattern doesn't resolve the Reclass reference

Moep90:
@simu thanks, valid points. I think the general issue here is that we're generally unaware of how people use Kapitan.

I agree that the original always-on behavior is too risky, so I’d like to adjust the PR before trying to merge it.

Proposed changes
1. Add a feature flag: `--enable-class-wildcards`, default off.
2. Thread `enable_class_wildcards=False` through `Inventory.__init__()` and `materialize_expanded_inventory()`.
3. Return the original inventory path unchanged unless the flag is enabled.
4. Match patterns only against full dotted class names.
5. Preserve default behavior for existing inventories.
6. Add regression tests for literal metacharacters, Reclass refs, ordinary parameter values, ordering, unmatched patterns, and symlinks.
7. Update docs to describe opt-in behavior, ordering risks, unmatched-pattern handling, literal-metacharacter limitations, and symlink limitations.

WDYT?

simu:
> @simu thanks, valid points. I think the general issue here is that we're generally unaware of how people use Kapitan.
> 
> I agree that the original always-on behavior is too risky, so I’d like to adjust the PR before trying to merge it.


In light of generally being unaware how people use Kapitan, I think opt-in should be the preferred mode for all new features that could break some people's inventories.

> 
> Proposed changes
> 
>     1. Add a feature flag: `--enable-class-wildcards`, default off.
> 
>     2. Thread `enable_class_wildcards=False` through `Inventory.__init__()` and `materialize_expanded_inventory()`.
> 
>     3. Return the original inventory path unchanged unless the flag is enabled.

I think having the feature opt-in would solve most of the immediate hypothetical compatibility issues (note that we do not currently have any class names that include pattern meta-characters).

> 
>     4. Match patterns only against full dotted class names.

I did a quick local test with only matching against full class names, and was able to easily make all tests in the PR pass without the special case basename matching. However, please note that there's another trade-off: the pattern which emulates the current basename matching most closely (`*dev-*`) will sometimes match additional classes. Example:

classes: `apps.dev-api`, `apps.dev-backend`, `dev-tools`, `containerdev-api`, `containerdev-backend`

with the current basename match, `dev-*` will only match `apps.dev-api`, `apps.dev-backend` and `dev-tools`.

however, full match with `*dev-*` will match all five classes and full match with `*.dev-*` won't match `dev-tools` anymore.

> 
>     5. Preserve default behavior for existing inventories.

I think this should be the default for any new features that modify inventory rendering behavior. For this PR, this will be achieved with 1-3.

> 
>     6. Add regression tests for literal metacharacters, Reclass refs, ordinary parameter values, ordering, unmatched patterns, and symlinks.

I don't see an easy solution for making the wildcard feature work for class includes that contain Reclass refs and wildcards (there's a circular dependency: you need to resolve the ref in order to expand the wildcard, but you need to expand the wildcards in order for reclass to be able to resolve the ref). Class includes that just contain a Reclass ref and no wildcards continue to work fine.

The symlink issue isn't a huge deal, we could completely fix this by simply changing the `shutil.copytree()` in `materialize_expanded_inventory()` to use `symlinks=False`. Alternatively, we can easily adjust Commodore to not generate relative symlinks.

IMO, ordering isn't so much a regression as just a pitfall that needs to be documented and the same for unmatched patterns.

> 
>     7. Update docs to describe opt-in behavior, ordering risks, unmatched-pattern handling, literal-metacharacter limitations, and symlink limitations.

It would definitely be helpful to have comprehensive docs for a feature like this.

> 
> 
> WDYT?

Overall, I like the feature, and the proposed changes sound like a solid plan to minimize the impact for existing users.

Moep90:
@simu thanks for the detailed review, all addressed:                                                                                                                          
                                                                                                                                                                               
  - Pattern semantics: removed the special basename rule. Every pattern matches against the full dotted class name, no surprises.
  - Exact-match precedence: literal class names that contain metacharacters (e.g., config[html]) are preserved and never treated as patterns.                                     
  - Reclass references: any ${…} entry is passed through verbatim, so ${?optional} won't be expanded.
  - Symlinks: followlinks=True for Commodore-style symlinks, plus broken external relative symlinks are rewritten to absolute paths in the temp copy.                            
  - ignore_class_notfound_regexp: documented the limitation — regex is evaluated against expanded names, not pattern strings.                                                    
  - Regression coverage: added a compile test in test_compile.py that exercises --enable-class-wildcards end-to-end.                                                             
                                                                                                                                                                                 
  Let me know if anything else stands out!

kapitan/inventory/wildcards.py
```diff
@@ -0,0 +1,359 @@
...
+                # Fast path: if the raw file contains no glob metacharacter
+                # there cannot be a wildcard class entry to expand.
+                try:
+                    with open(path, encoding="utf-8") as fh:
+                        text = fh.read()
+                except OSError:
+                    continue
+                if not any(c in text for c in GLOB_METACHARACTERS):
+                    continue
+                data = _safe_load_yaml(path)
```
simu: You could avoid re-reading the file contents from disk here by passing `text` to `_safe_yaml_load()` instead of `path`:


```suggestion
                data = _safe_load_yaml(text)
```

kapitan/inventory/wildcards.py
```diff
@@ -0,0 +1,359 @@
...
+        for name in list(dirs) + filenames:
+            dest_link = os.path.join(dirpath, name)
+            if not os.path.islink(dest_link):
+                continue
+            target = os.readlink(dest_link)
+            if os.path.isabs(target):
+                continue
+            # Does the relative target resolve from the new location?
+            dest_resolved = os.path.normpath(
+                os.path.join(os.path.dirname(dest_link), target)
```
simu: nitpick: you don't need `os.path.dirname(dest_link)` here, you can just use `dirpath`:


```suggestion
                os.path.join(dirpath, target)
```

kapitan/inventory/wildcards.py
```diff
@@ -0,0 +1,359 @@
...
+            if os.path.exists(dest_resolved):
+                continue
+            # Try to resolve from the corresponding original location.
+            rel_dir = os.path.relpath(dirpath, dest_root)
+            src_link = os.path.join(src_root, rel_dir, name)
+            if not os.path.islink(src_link):
+                continue
+            src_target = os.readlink(src_link)
+            if os.path.isabs(src_target):
+                continue
```
simu, line 309: Realistically we should never hit these cases, because `src_link` must be a relative symlink if we get here: if `dest_link` isn't a symlink we already skipped the direntry in line 274 and if `dest_link` is absolute we already skipped the direntry in line 277.

kapitan/inventory/wildcards.py
```diff
@@ -0,0 +1,359 @@
...
+    dest = os.path.join(tmp_root, base_name)
+    shutil.copytree(inventory_path, dest, symlinks=True)
+
+    # Fix relative symlinks that now point outside the temp tree.
+    _fix_relative_symlinks(inventory_path, dest)
+
+    logger.debug(
+        f"Materialized inventory with wildcard class expansion at {dest} "
+        f"(rewrote {len(files_to_expand)} file(s))"
+    )
```
simu: The log message and position don't fully match. We're about to "rewrite N files" but the log message states "(rewrote N files)". I'd probably just move the log message below the for loop that does the rewriting.

docs/pages/inventory/classes.md
```diff
@@ -25,6 +25,144 @@ classes:
...
+* Patterns are matched against the **full dotted class name** using
+  `fnmatch`. There is no special-case based on whether a pattern contains
+  `.`. For example, `dev-*` matches only top-level classes named
+  `dev-...`; to match `apps.dev-api` you must write `apps.dev-*` or
+  `*.dev-*`.
+* Class entries that look like Reclass / Kapitan parameter references —
+  any string containing `${` and `}`, including `${var}` and `${?var}` —
+  are passed through to the backend unchanged, even if they contain a
+  glob metacharacter such as `?`. The wildcard expander cannot resolve
+  such references; the backend handles them after class inheritance.
```
simu, line 74: Maybe we should make it even more explicitly clear that mixing Reclass references and patterns in a single class include doesn't work because pattern expansion happens before any Reclass references can be resolved.


```suggestion
* Class entries that look like Reclass / Kapitan parameter references —
  any string containing `${` and `}`, including `${var}` and `${?var}` —
  are passed through to the backend unchanged, even if they contain a
  glob metacharacter such as `?`. The wildcard expander cannot resolve
  such references; the backend handles them after class inheritance.
  Notably, this also means that pattern expansion isn't supported in
  class includes that use Reclass references.
```

simu (commented):
@Moep90 looks good now. I was able to run `kapitan inventory` on one of our production inventories (with a small patch to trigger the inventory materialization) without any issues with the latest PR version.

The docs are now very clear on what's possible as well, although it might make sense to make it even more explicitly clear that it's not possible to have Reclass references and wildcard matches in the same class include (cf. suggestion).

kapitan/inventory/wildcards.py
Moep90: moved

kapitan/inventory/wildcards.py
Moep90: Added a [comment](https://github.com/kapicorp/kapitan/pull/1457/changes#diff-06a84982d07847b423dc31a68bbf4ccf6939e74b5d79845539b097a6d3992e3aR283-R286)

kapitan/targets.py
```diff
@@ -283,6 +283,22 @@ def compile_target(
...
+        cached.inv
+        and hasattr(cached.inv, "targets")
+        and target_name in cached.inv.targets
+    ):
+        inv_target = cached.inv.targets[target_name]
+        if inv_target.parameters and inv_target.parameters.kapitan is not None:
+            inv_target.parameters.kapitan.target_full_path = (
+                target_config.target_full_path
+            )
+
```
ramaro: I am confused about this. Is the behaviour mentioned in the comment introduced with this change? Or did it happen before? I don't fully follow the problem.

Moep90: @ramaro This behavior is pre-existing, not introduced by the PR's feature commits.
The merge base 1c79e97e already had load_target_inventory() inside the `multiprocessing.Pool` block, so workers forked before target_full_path was set on `cached.inv`.
Because the forked worker inherits a truthy `cached.inv`, the not `cached.inv` guard skips `from_dict(globals_cached)`, leaving the stale copy in place.
The PR's commits never touched `kapitan/targets.py`.
Master later fixed the root cause structurally in 54e1244d by moving `load_target_inventory()` outside the pool block. Since this PR has since merged with master, that structural fix is already present.
The sync block in 1778101d was added to fix failing `reclass-rs` and `omegaconf` tests at the time. Now that the branch has the master merge, it's defensive/redundant.

Happy to remove it if you prefer, or keep it as a defense-in-depth.

ramaro (approved):

ramaro:
@Moep90 thank you, LGTM

PR #1458, opened 2026-05-11
feat: Add coverage check script and corresponding unit tests

Moep90:
# Summary

Adds a coverage check script with unit tests so coverage validation can be run explicitly and consistently in local development and CI.

# What changed

- Added a script for checking coverage against the configured threshold.
- Added unit tests for successful and failing coverage-check scenarios.
- Made coverage enforcement easier to invoke without relying only on implicit pytest/coverage behavior.

# Why

This makes coverage validation more transparent and maintainable. It also gives CI a clear command to run when verifying whether coverage expectations are still met.

# Validation
- Unit tests for the coverage check script pass.
- Existing test suite remains green.
- Coverage threshold behavior is preserved.

# Out of scope
- Changing the coverage threshold.
- Adding broad new test coverage unrelated to the coverage-check script.
- Modifying production code behavior.

### DIFF
--- a/scripts/check_coverage.py
+++ b/scripts/check_coverage.py
@@ -1,0 +1,165 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2020 The Kapitan Authors <kapitan-admins@googlegroups.com>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Per-module coverage gate for security-sensitive Kapitan modules.
+
+Usage::
+
+    python scripts/check_coverage.py [--data-file .coverage]
+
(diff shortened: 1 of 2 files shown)


Moep90:
Closing as out of scope / redundant.

Coverage enforcement already exists on master via pytest `fail_under = 64` in pyproject.toml (and `uv run coverage report --fail-under`); a separate 165-line `scripts/check_coverage.py` duplicates that. The branch is also ~40 commits behind master and was stacked on an unrelated fetch-retry branch.

If an explicit CI command is still wanted, it should be a small wrapper opened directly off master rather than stacked here.

PR #1464, opened 2026-05-15
Introduce rapidyaml emitter option

ramaro:
## Proposed Changes

Adds support for rapidyaml emitter option via the compile flag `--yaml-use-rapidyaml`

In my local tests, this cuts compile times in a rather large codebase by a factor of 3.

It supports the current yaml styles implement (literal, folded, double-quotes), however the initial compilation will output formatting differences when compared with a previous compilation with PyYAML _but_ it is expected to be semantically equivalent.

## Docs and Tests

* [x] Tests added
* [x] Updated documentation


Moep90:
@ramaro IMHO LGTM

Moep90 (approved):

PR #1465, opened 2026-05-15
feat: add output_file option to Kustomize input type for single-file output

### LINKED ISSUE #1455: [feat]: Support Single-File Output for `kustomize` Input
# Summary

Kapitan's `input_type: kustomize` currently renders Kustomize output and then splits each YAML document into a separate file named from the resource kind and name. It would be useful to optionally preserve the full `kustomize build` output as a single multi-document YAML file, similar to how Helm input can write a chart render to one output file.

# Motivation

`kustomize build` naturally emits a single multi-document YAML stream. Kapitan currently parses that stream and re-dumps each document separately:

```text
<metadata.name>-<kind>.yaml
```

This has two drawbacks:

- Users cannot choose a single output file such as `multus-install.yml`.
- YAML formatting is not preserved. For example, ConfigMap multiline values may be re-emitted as escaped quoted strings instead of block scalars.

For deployment workflows that expect one manifest bundle per component, the current behavior is less ergonomic than Helm input output.

# Proposed Behavior

Add an optional field to the `kustomize` input configuration:

```yaml
parameters:
  kapitan:
    compile:
      - input_type: kustomize
        input_paths:
          - components/kustomize/multus
        output_path: manifests
        output_file: multus-install.yml
        namespace: kube-system
```

When `output_file` is set, Kapitan should write the raw `kustomize build` stdout to:

```text
compiled/<target>/manifests/multus-install.yml
```

When `output_file` is not set, Kapitan should keep the current behavior of splitting resources into individual files.

# Benefits

- Matches Kustomize's native output model.
- Enables one-file manifest bundles for GitOps and apply workflows.
- Preserves Kustomize-rendered YAML formatting.
- Keeps existing behavior backward compatible.
- Brings Kustomize closer to Helm input ergonomics.

# Compatibility

This should be backward compatible if `output_file` is optional and defaults to the existing split-file behavior.

# Example

Given:

```yaml
parameters:
  kapitan:
    compile:
      - input_type: kustomize
        input_paths:
          - components/kustomize/multus
        output_path: manifests/multus
        output_file: install.yml
```

Expected output:

```text
compiled/lab/manifests/multus/install.yml
```

containing the complete multi-document YAML stream from `kustomize build`.

## Alternative Workaround

Currently users can use `input_type: external` to run:

```sh
kustomize build components/kustomize/multus > compiled/lab/manifests/multus-install.yml
```

However, this duplicates behavior that the built-in Kustomize input already performs and requires custom wrapper scripts.

Moep90:
Fixes #1455

## Summary

This PR adds optional `output_file` support to Kapitan's `input_type: kustomize`.

When `output_file` is set, Kapitan writes the raw `kustomize build` stdout to a single file under the configured `output_path`. This preserves Kustomize's native multi-document YAML stream instead of parsing and re-serializing it into separate resource files.

The existing split-file behavior remains the default when `output_file` is not configured.

## Motivation

The current Kustomize input renders `kustomize build` output and then splits each YAML document into a separate file named from the resource kind and metadata name.

That behavior remains useful, but it has two practical drawbacks for some workflows:

- Users cannot ask Kapitan to write one complete manifest bundle, such as `multus-install.yml`.
- YAML formatting from Kustomize is not preserved because the output is parsed and re-dumped. This can affect readability, for example when ConfigMap multiline values are emitted as escaped quoted strings instead of block scalars.

Kustomize naturally emits a single multi-document YAML stream. This PR makes it possible to preserve that output directly, similar to the single-file output ergonomics already available for Helm-based workflows.

### Kustomize input schema

Adds an optional `output_file` field to `KapitanInputTypeKustomizeConfig`:

```python
output_file: str | None = None
```

### Kustomize compilation behavior

When `output_file` is set:

- Kapitan writes raw `kustomize build` stdout directly to the requested file.
- The output is not parsed, split, or re-serialized.
- Nested relative paths such as `bundles/install.yml` are supported.
- Path traversal and absolute paths are rejected to prevent writes outside the compile directory.

When `output_file` is not set:

- The existing per-resource split behavior is preserved unchanged.

### Documentation

Updates the Kustomize input type documentation with:

- The new `output_file` configuration option.
- A single-file output example.

### Tests

Adds regression coverage for:

- Single-file raw output.
- Nested output file paths.
- Backward-compatible split-file behavior when `output_file` is omitted.
- Rejection of unsafe paths such as `../escape.yml` and absolute paths.
- Schema validation for the new field.

## Example

```yaml
parameters:
  kapitan:
    compile:
      - input_type: kustomize
        input_paths:
          - components/kustomize/multus
        output_path: manifests/multus
        output_file: install.yml
```

This writes the complete `kustomize build` output to:

```text
compiled/<target>/manifests/multus/install.yml
```

## Backward Compatibility

This change is backward compatible.

`output_file` is optional and defaults to `None`, so existing `input_type: kustomize` configurations continue to split resources into individual files exactly as before.

## Validation

The following checks pass:

- [x] `pytest tests/test_kustomize_input.py -v`
- [x] `pytest tests/ -v --no-cov -m "not slow"`
- [x] `make lint && make format`

## Note on `tests/conftest.py`

`isolated_test_resources` now ignores `.coverage*` files when copying the test resources tree (`ignore=_ignore_coverage_files`, `ignore_dangling_symlinks=True`). This is unrelated to the Kustomize feature: under parallel test runs, coverage data files could leak into `test_resources` and break the isolated copy. Calling it out so it is not mistaken for scope creep.


kapitan/inputs/kustomize.py
```diff
@@ -159,21 +159,52 @@ def compile_file(
...
+                        yaml.dump(doc, out, default_flow_style=False)
+
+        except KustomizeTemplateError:
+            raise
         except Exception as e:
             raise KustomizeTemplateError(
                 f"Failed to compile Kustomize overlay: {e!s}"
             ) from e
+
+    def _write_output_file(
```
ramaro: to be honest, this method should use `.to_file()` instead of re-inventing kapitan file writing allover again. 
[to_file()](https://github.com/kapicorp/kapitan/blob/master/kapitan/inputs/base.py#L136) manages output types and ref revealing for the input type.

ramaro (changes requested):

kapitan/inputs/kustomize.py
Moep90: Good call. Done in 2a0e9ac. The single-file path now delegates to `to_file()`, so it picks up ref revealing and the standard output handling instead of re-implementing the write. The raw multi-document `kustomize build` stream is preserved by writing it as plain output (`output_type` forced to `PLAIN` for that write), and `CompiledFile` handles directory creation. I kept a slim `_resolve_output_file()` just for the path-traversal guard.

I left the per-resource split path on its manual `yaml.dump` for now, to avoid changing its output formatting or filenames in this PR. Happy to migrate that to `to_file()` in a follow-up if you'd like it consistent.

PR #1494, opened 2026-05-18
fix(docs): resolve fenced code block rendering and add rendering verification

Moep90:
## Problem

`pymdown-extensions` 10.19.1 silently broke all fenced code blocks, including `yaml`, `tree`, and `mermaid` blocks, on kapitan.dev.

They rendered as inline text inside `<p>` tags instead of `<pre><code>` elements.

**Before, broken in 10.19.1:**

```html
<h2 id="targets">Targets</h2>
<p>```yaml</p>
<h1 id="inventorytargetsproductionyaml">inventory/targets/production.yaml</h1>
<p>classes:  - common  - acme.applications.super</p>
<p>parameters:    description: My Super Application```</p>
```

**After, fixed in 10.21.3:**

```html
<div class="highlight">
  <pre><span></span><code class="language-yaml">
    <span class="c1"># inventory/targets/production.yaml</span>
    <span class="nt">classes</span><span class="p">:</span>
    ...
  </code></pre>
</div>
```

The same bug affected:

```tree result="shell"
...
```

and:

```mermaid
...
```

## Root cause

`pymdownx.superfences` 10.19.1 together with `markdown` 3.10 failed to parse fenced blocks when loaded alongside the full MkDocs extension set.

Upgrading to `pymdown-extensions >=10.21` restores correct parsing.

## Changes

- **`pyproject.toml`**: bump `pymdown-extensions` from `~=10.9` to `>=10.21,<11`
- **`tests/test_docs_rendering.py`**: add a pytest test that loads the same Markdown extensions as `mkdocs.yml` and asserts fenced blocks render as `<pre>` with no literal backticks leaking into HTML
- **`.github/workflows/documentation.yml`**: run rendering verification after `mkdocs build`
- **`.pre-commit-config.yaml`**: add a local hook that runs the rendering test on `docs/**` or `mkdocs.yml` changes
- **`Makefile`**: add a `docs_check` target

## Test plan

- [ ] `uv run pytest tests/test_docs_rendering.py -v --no-cov` passes
- [ ] `make docs_check` passes
- [ ] `uv run mkdocs build --strict` produces no broken fenced blocks
- [ ] Deployed site shows properly formatted code blocks


Moep90:
# Before

Screenshot_20260518_184530

# Fixed

Screenshot_20260518_185033

ademariag (approved):

PR #1498, opened 2026-05-19
fix(helm): filter None values at the Helm input source

### LINKED ISSUE #1396: [bug]: --yaml-dump-null-as-empty produces empty documents that kubectl rejects
### Description of the bug

The `--yaml-dump-null-as-empty` flag (added in PR #861 to fix #843) does not fully resolve the null document issue. Instead of outputting `--- null`, it outputs an empty document (`---` followed by `...`) which kubectl still rejects:

```
error converting YAML to JSON: yaml: did not find expected node content
```

The flag changes the symptom but doesn't fix the underlying problem.

### Steps to reproduce

```bash
1. Create a Helm chart with a comment before the first `---`:


# This is a comment
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: example


2. Compile with Kapitan using the flag:

kapitan compile -t my-target --yaml-dump-null-as-empty


3. Inspect the output - it contains:

apiVersion: v1
kind: ServiceAccount
metadata:
  name: example
---
...


4. Apply with kubectl:

kubectl apply -f compiled/.../output.yaml


5. kubectl rejects with: `error converting YAML to JSON: yaml: did not find expected node content`
```

### What happened?

**Expected:** `--yaml-dump-null-as-empty` should produce valid YAML that kubectl accepts.

**Actual:** The flag produces `---\n...` (empty document) instead of `--- null`, but kubectl rejects both.

**Suggested fix:** Filter out null/empty documents entirely rather than changing their representation. When a Helm template produces a null document (from comments before `---`), Kapitan should skip serializing that document altogether.

### Additional context

- Affected charts: Any Helm chart with comments before the first `---` (e.g., `flink`, `kube-prometheus-stack`)
- Current workaround: Post-processing with `sed -i -e '/^---$/{ N; /^---\n\.\.\.$/d; }' file.yaml`
- Related: #843, PR #861 

### Version (kapitan)

v0.31.0 (stable)

### Version (Python)

I use kapitan directly (pip, docker)

- [ ] pyenv
- [ ] venv
- [ ] no environment

### OS information

Ubuntu 22.04

### Relevant log output

```bash
$ kubectl apply --dry-run=client -f flink-rbac.yaml
error converting YAML to JSON: yaml: did not find expected node content
```

Moep90:
## Summary

When compiling Helm charts, comments before the first `---` (or between documents) can result in `None` entries from `yaml.safe_load_all`. These empty YAML documents cause `kubectl` to reject the output with:

```
error converting YAML to JSON: yaml: did not find expected node content
```

This PR moves the filtering to the **Helm input source** rather than the shared `write_yaml` serializer, as suggested by @ramaro. This follows the same pattern already used by the kustomize input.

## Changes

- **Removed** the generic None-filtering from `kapitan/inputs/base.py` `write_yaml()`
- **Added** `if doc is not None` filtering in `helm.py` where multi-document YAML is loaded:
  - `compile_file()` — when reading rendered chart files
  - `load_chart()` — when parsing `helm template` output for `HelmChart`
- **Removed** `tests/test_yaml_null_filter.py` (tests for the base.py behavior)
- **Removed** null-filter tests from `tests/test_yaml_use_rapidyaml.py`
- **Added** `test_safe_load_all_filters_none_docs` in `tests/test_helm_input.py`

Fixes #1396


kapitan/inputs/base.py
```diff
@@ -337,6 +337,12 @@ def write_yaml(self, obj: object):
         else:
             obj = self.revealer.compile_obj(obj, target_name=target_name)
 
+        # Filter out None values from multi-document streams (e.g., empty
+        # YAML documents produced by helm template with comments before ---).
+        # https://github.com/kapicorp/kapitan/issues/1396
+        if not isinstance(obj, Mapping) and isinstance(obj, list | tuple):
+            obj = [item for item in obj if item is not None]
+
```
ramaro: shouldn't this be filtered out at the source? And by that I mean, shouldn't the Helm input do this instead? This function is too crucial to always be running this filtering, I think.

Moep90:
@ramaro You're absolutely right, thanks for the review! I moved the filtering into the Helm input where the YAML is loaded:

- helm.py:compile_file() -- filters when reading rendered chart files
- helm.py:load_chart() -- filters when parsing helm template output for HelmChart

This follows the same pattern already used by the kustomize input. I also removed the filtering from base.py and dropped the base.py-specific tests, replacing them with a helm input test.

ramaro (approved):

ramaro:
@Moep90 thanks for the fix!

PR #1499, opened 2026-05-19
ci: pin CUE version and delegate tool installs to Makefile

Moep90:
## Summary

This PR stabilizes the external tool installation step in CI, which has been frequently flaky due to unauthenticated GitHub API rate limits when fetching the latest CUE release.

### Changes

- **Pin CUE to v0.12.0** — eliminates the `api.github.com` call that was hitting the 60 req/hour unauthenticated rate limit on shared GitHub Actions runners.
- **Make `Makefile` arch-aware** — `install_cue` now detects `amd64`/`arm64` so it works on both CI runner types.
- **Delegate to Makefile in CI** — `test-build-publish.yml` now uses `make install` and `make install_external_tools` instead of duplicating shell blocks, giving us a single source of truth.
- **Add curl retries** — all `curl` invocations in the Makefile now use `--retry 3 -fsSL`.
- **Bump cache key** — `actions/cache` key bumped to `v2` to force a fresh cache with the new binaries.

### Related failures

- https://github.com/kapicorp/kapitan/actions/runs/26078275861/job/76673793599?pr=1498

### DIFF
--- a/.github/workflows/test-build-publish.yml
+++ b/.github/workflows/test-build-publish.yml
@@ -109,8 +109,7 @@ jobs:
         uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
       - name: Install libraries dependencies
         if: needs.check-changes.outputs.has-code-changes == 'true'
-        run: |
-          uv sync --locked --all-extras --dev
+        run: make install
 
       - name: Cache external tools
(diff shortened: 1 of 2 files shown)


Moep90:
Superseded by #1411 — mise now handles all external tool installation, which removes the GitHub API rate-limit calls and Makefile install targets this PR was patching.

PR #1503, opened 2026-05-21
fix(vault): reject symbolic link token files

Moep90:
## Summary

The Vault token file at `~/.vault-token` was read without checking if it is a symbolic link. An attacker with filesystem access could replace the token file with a symlink pointing to an arbitrary file (e.g. `/etc/shadow`), causing Kapitan to read and use its contents as a Vault token.

## Changes

- Add an `os.path.islink()` check in `VaultClient.read_token_from_file()` that raises `VaultError` when the token file is a symbolic link.
- Add unit tests covering symlink rejection, regular file acceptance, and empty file rejection.

## Test plan

- `tests/test_vault_resources_security.py` — new test file with 3 focused unit tests that do not require a running Vault server:
  - `test_read_token_from_file_rejects_symlink`
  - `test_read_token_from_file_accepts_regular_file`
  - `test_read_token_from_file_rejects_empty_file`


ramaro (approved):

ramaro:
@Moep90 LGTM, thank you

Issue #1512, opened 2026-05-26
[test] Add test markers (slow/integration/unit) and fix test isolation issues


Moep90:
The test suite has 391 test functions across 35 files, but not a single one is decorated with the `@pytest.mark.slow`, `@pytest.mark.integration`, or `@pytest.mark.unit` markers we declare in `pyproject.toml` (lines 174-177).

So `uv run pytest -m "not slow"` is a no-op. Tests that run full compilation pipelines, Docker containers, or external tools (Helm, Kustomize, Vault) all fire on every invocation.

### Other test isolation problems

1. Module-level `tempfile.mkdtemp()` leaks. At least 8 test files create temp dirs at import time that never get cleaned up:
   - `tests/test_azure.py:15`
   - `tests/test_gpg.py:28,46`
   - `tests/test_cli.py:25`
   - `tests/test_vault.py:25`
   - `tests/test_vault_transit.py:18`
   - `tests/test_refs.py:23`
   - `tests/test_inventory.py:73` (class-level, shared across instances, which is dangerous under `pytest-xdist`)

2. Tests write into shared source directories. `tests/test_compile.py` calls `os.chdir(TEST_RESOURCES_PATH)` and compiles straight into the source tree. `tearDown` only changes the directory back and calls `reset_cache()`, it doesn't delete the `compiled/` directory it left behind. A crashing test leaves stale output around, so later runs can fail or pass for the wrong reasons.

3. `pytest-random-order` is installed but not actually enabled. It's declared in `[dependency-groups] test` but missing from `addopts`, so tests always run in source order and order-dependent state leaks stay hidden.

### Proposed solution

Markers:
- Mark the compilation tests in `test_compile.py` (24 methods) and the `test_dependency_manager.py` tests that fetch real Git/Helm/HTTP artifacts as `@pytest.mark.slow` or `@pytest.mark.integration`.
- Mark the pure unit tests (e.g. `test_cached.py`, the simple `test_refs.py` cases) as `@pytest.mark.unit`.
- Update CI to run unit tests first, then integration tests in a separate job.

Isolation:
- Swap every module-level `tempfile.mkdtemp()` for a `tempfile.TemporaryDirectory()` context manager inside `setUp`/`tearDown` or pytest fixtures.
- Move the compile tests onto the existing `isolated_kubernetes_inventory`, `isolated_terraform_inventory`, and `isolated_docker_inventory` fixtures in `conftest.py`.
[...]

PR #1516, opened 2026-05-26
refactor(test): add API stability contract tests for downstream consumers

Moep90:
## Summary

Add API stability contract tests that encode the stable public-ish API surface that downstream tools (orchestrators, admission controllers, component generators) rely on when embedding Kapitan as a library.

This PR also adds a minimal `examples/reclass` inventory that the contract tests exercise.

### New file: `tests/test_external_contracts.py`

Added **24 contract tests** covering:

**Happy-path contracts:**
- **Cached args contract** — `compile_targets()` accepts the namespace shape external tools build (~20 attributes: `inventory_backend`, `targets`, `parallelism`, `fetch`, `reveal`, `jinja2_filters`, etc.).
- **Compile targets signature** — `compile_targets(inventory_path, search_paths, ref_controller, args)` must remain stable.
- **Load target inventory** — returns a list of target objects with `.vars.target` accessible; empty list returns all targets.
- **RefController API** — constructor and `register_backend()` accepting custom `PlainRefBackend` subclasses.
- **Reset cache isolation** — multiple `compile_targets()` invocations in the same process do not leak state.
- **Output directory structure** — `compiled/<target-name>/` exists after compilation.
- **Inventory backend contract** — `inventory()` returns a dict with `parameters` per target.
- **Refs contract** — `REF_TOKEN_TAG_PATTERN` matches standard `?{backend:path}` tags; `Revealer` with `embed_refs=True` resolves embedded refs.
- **Kadet contract** — `inventory()`, `BaseObj`, and `Dict` remain importable and functional.

**Edge-case contracts:**
- Non-existent target name raises `CompileError` (not a generic exception).
- `parallelism=1` works correctly (single-process mode).
- Backend switching in the same process (reclass → reclass-rs) does not leak cached state.
- `REF_TOKEN_TAG_PATTERN` rejects malformed ref strings (missing `?`, missing backend, etc.).
- `Revealer` leaves plain strings without ref tags completely untouched.
- `RefController[ missing_ref ]` raises `KeyError` (not a generic exception).
- `PlainRef` handles Unicode correctly.

### New example inventory: `examples/reclass`

A minimal reclass example inventory with three targets (`simple`, `advanced`, `compile-test`) used by the contract tests. This is a self-contained example that downstream users can also reference.

## Test plan

- [ ] Run `uv run pytest tests/test_external_contracts.py --no-cov` — 24 tests pass
- [ ] Run full suite: `uv run pytest tests/ --no-cov`

Related to #1512

### DIFF
--- a/examples/reclass/inventory/classes/components/app.yml
+++ b/examples/reclass/inventory/classes/components/app.yml
@@ -5,7 +5,7 @@ classes:
 parameters:
   app_name: myapp
 
-  # List override (reclass replaces lists by default)
+  # List extension (reclass appends to lists by default)
   tags:
     - app-tag
 
--- a/tests/test_external_contracts.py
+++ b/tests/test_external_contracts.py
@@ -490,11 +490,11 @@ class TestRefsContract:
         revealed = revealer.reveal_obj(obj)
         assert revealed["password"] == "secret-value"
 
-    def test_revealer_leaves_plain_strings_untouched(self):
+    def test_revealer_leaves_plain_strings_untouched(self, temp_dir):
         """Revealer must not modify strings that contain no ref tags."""
         from kapitan.refs.base import RefController, Revealer
 
-        rc = RefController("/tmp")
+        rc = RefController(temp_dir)
         revealer = Revealer(rc)
 
         obj = {"message": "hello world", "number": 42}
--- a/tests/test_external_contracts.py
+++ b/tests/test_external_contracts.py
@@ -1,0 +1,536 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+"""API stability tests for external consumers.
+
+These tests encode the stable public-ish API surface that downstream tools
+rely on when embedding Kapitan as a library. They are not tests of internal
+implementation details — they exercise the *contracts* that must hold across
+Kapitan releases so that external orchestrators do not break.
+"""
+
+from __future__ import annotations
+
+import os
+from argparse import Namespace
+
+import pytest
+
+import kapitan.cached
+from kapitan.cached import reset_cache
+from kapitan.cli import build_parser
+from kapitan.defaults import DEFAULT_JINJA2_FILTERS_PATH
+from kapitan.inventory import InventoryBackends
+from kapitan.resources import inventory as get_inventory
+from kapitan.targets import compile_targets, load_target_inventory
+from tests.test_helpers import IsolatedTestEnvironment
+
+
+RECLASS_EXAMPLE_PATH = os.path.join(os.getcwd(), "examples", "reclass")
+
+
+def _library_style_args(**overrides):
+    """Build an args namespace matching the shape external tools set.
+
+    Downstream orchestrators (e.g. Commodore, custom wrappers) populate
+    ~20 attributes on ``cached.args`` before calling ``compile_targets()``.
+    This helper produces a namespace with the same shape so that changes to
+    attribute names or required defaults are caught immediately.
+    """
+    args = Namespace(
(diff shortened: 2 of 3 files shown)


tests/test_external_contracts.py
```diff
@@ -0,0 +1,560 @@
+#!/usr/bin/env python3
```
Moep90, line 1: @simu could you take a look at this when you have a moment? I tried to cover the external contracts derived from the Commodore usage patterns.

examples/reclass/inventory/classes/components/app.yml
```diff
@@ -0,0 +1,15 @@
+---
+classes:
+  - features.lists
+
+parameters:
+  app_name: myapp
+
+  # List override (reclass replaces lists by default)
```
simu: I don't think that's true with Kapitan's reclass version (and also contradicted by the expected output!)


```suggestion
  # List extension (reclass appends to lists by default)
```

tests/test_backend_examples.py
```diff
@@ -0,0 +1,607 @@
...
+Scalar override                   | yes     | yes        | yes
+Exports                           | yes     | no*        | yes
+Escaped interpolation             | yes     | yes        | yes
+Deferred interpolation            | n/a     | n/a        | yes
+Custom resolvers                  | n/a     | n/a        | yes
+Conditional resolvers             | n/a     | n/a        | yes
+Arithmetic resolvers              | n/a     | n/a        | yes
+Compile output structure          | yes     | yes        | yes
+
+* Known difference: reclass-rs returns empty exports for some inventories.
```
simu, line 44:
```suggestion
* Known difference: reclass-rs returns empty exports for all inventories. A draft PR for exports support in reclass-rs exists at https://github.com/projectsyn/reclass-rs/pull/168. Currently needs user input in the form of example inventories that use exports.
```

tests/test_external_contracts.py
```diff
@@ -0,0 +1,560 @@
...
+            ignore_class_notfound=False,
+        )
+        inv = r.inventory()
+        data = inv.as_dict()
+
+        assert isinstance(data, dict)
+        assert "nodes" in data
+        for target_name in ("simple", "advanced", "compile-test"):
+            assert target_name in data["nodes"], f"missing target {target_name}"
+            assert "parameters" in data["nodes"][target_name]
```
simu: IMO this test doesn't check the right thing (external users shouldn't need to care whether reclass-rs's `as_dict()` does the right thing, since they should never even notice that this method is called or care that it's necessary) -- if anything, we should verify that each backend's `kapitan/inventory/<backends>` implementation produces the right inventory shape in `render_targets()` rather than copying (one of) those implementations into the test, but it's not clear to me that we need a separate test for that (since multiple other tests will already fail if a backend produces a wrong inventory shape).

Reclass-rs itself already has a test which ensures that the `as_dict()` shape is correct (cf. https://github.com/projectsyn/reclass-rs/blob/34cbcae38ef6f5c10d3ee2b4ce4b2e81bedd87c4/tests/test_inventory.py#L102-L116) and fairly thorough api compat test coverage with kapicorp-reclass (cf. https://github.com/projectsyn/reclass-rs/blob/main/tests/test_with_kapicorp_reclass.py and https://github.com/projectsyn/reclass-rs/blob/main/tests/test_inventory_class_mapping.py)

simu (commented):
some initial feedback, didn't have time yet to carefully look at the full PR

tests/test_external_contracts.py
Moep90: Thank you, @simu, I removed this part.

PR #1519, opened 2026-05-29
Seed inventory cache once per worker via Pool initializer

sviscaino:
## Summary

Currently, the inventory cache is passed at the task-level, which incurs a high overhead since python needs to pickle the whole inventory once per task. Instead, we pass it to each worker process using `initializer`. I tested this locally and seen ~5x speedups in compile times on large inventories.

## Affected files or URLs

kapitan/targets.py

## Test plan

- [x] Tests added or updated
- [x] Documentation updated (if user-facing behavior changed)
- [x] `make lint` passes
- [x] Full test suite passes (`uv run pytest tests/ --no-cov`)

### DIFF
--- a/kapitan/targets.py
+++ b/kapitan/targets.py
@@ -28,6 +28,23 @@ from kapitan.resources import get_inventory
 logger = logging.getLogger(__name__)
 
 
+def _pool_init(globals_cached):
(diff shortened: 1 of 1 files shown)


ramaro:
@sviscaino thank you!

ramaro (approved):

PR #1520, opened 2026-06-08
Fix Kadet cache key collisions from input_params

timReynolds:
## Summary

- Include Kadet's effective `input_params` in the compilation cache key.
- Build those params from a local copy so `compile_file()` no longer mutates `config.input_params` when adding the default `compile_path`.
- Add a focused unit test that asserts `Kadet.compile_file()` passes the effective params into `inputs_hash()` and leaves the original config params unchanged.

## Bug

Kadet cache keys currently include the target inventory digest, target name, and Kadet input path, but not the per-input parameters passed to the Kadet component.

When `kapitan compile --cache` is used, two compile entries can call the same Kadet component with different `input_params`. Without those params in the key, both entries compute the same cache key, so the second call can load and write the first call's cached output. That makes parameterized Kadet components produce stale or incorrect manifests whenever the same component is reused with different params.

`compile_file()` also adds a default `compile_path` to the params passed into Kadet modules. This PR treats that effective params dict as part of the cache input, while avoiding mutation of the original `config.input_params` object.

## Root cause

The regression was introduced by `65e98654a62eefc4b5bef1c0b6ac7f23289aa0f6` (`Implement kadet input type cache (#1319)`). That commit added Kadet cache lookup and `inputs_hash()`, but the key only used:

- `inventory_digest(current_target.get())`
- `target_name`
- `Path(input_path)`

Kadet `input_params` already existed before that change, so enabling the new cache created a collision path for parameterized Kadet inputs.

## Tests

- `uv run pytest -q tests/test_kadet.py --no-cov`
- `uv run pytest -q tests/test_compile.py::CompileTestResourcesTestKadet --no-cov`
- `uv run ruff check kapitan/inputs/kadet.py tests/test_kadet.py`

### DIFF
--- a/kapitan/inputs/kadet.py
+++ b/kapitan/inputs/kadet.py
@@ -128,28 +128,29 @@ class Kadet(InputType):
         Write the kadet evaluated items as files to compile_path.  External variables are not used in Kadet.
         """
 
-        input_params = config.input_params
+        input_params = dict(config.input_params)
         target_name = self.target_name
         current_target.set(target_name)
         search_paths.set(self.search_paths)
         inputs_hash = None
         output_obj = None
 
+        # set compile_path allowing kadet functions to have context on where files
+        # are being compiled on the current kapitan run
+        # we only do this if user didn't pass its own value
+        input_params.setdefault("compile_path", compile_path)
+
         if cache_obj := self.cacheable():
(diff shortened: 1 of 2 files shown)


ramaro:
@timReynolds good find, thanks for the fix!

ramaro (approved):

PR #1521, opened 2026-06-08
Clarify OmegaConf literal escaping docs

timReynolds:
## Summary

Clarifies the OmegaConf inventory documentation around escaped interpolation syntax:

- documents that `\${...}` is deferred interpolation under the double-pass OmegaConf backend
- documents `${escape:...}` as the intended way to emit literal `${...}` strings in compiled output
- adds examples for Terraform references, shell variables, and quoted complex expressions
- updates the migration table from the stale `${tag:...}` guidance to `${escape:...}`

## Rationale

Based on the upstream 0.35 changes, this appears to be intentional behavior rather than a runtime regression to patch in the loader. PR #1371 introduced the double-pass resolution flow where `\${...}` is unescaped after the first pass and resolved in the second pass. PR #1445 then repurposed `${escape:...}` to emit literal `${...}` output protected from both resolution passes.

This PR therefore updates the documentation to match the current behavior instead of changing `OmegaConfInventory.load_file()` or altering resolution semantics.

## Validation

- `git diff --check`

Docs-only change; no runtime tests were run.

### DIFF
--- a/docs/pages/inventory/omegaconf.md
+++ b/docs/pages/inventory/omegaconf.md
@@ -34,6 +34,8 @@ The OmegaConf backend uses a **double-pass resolution** strategy:
 
 This enables **deferred evaluation** - interpolations that resolve based on their final context after merging.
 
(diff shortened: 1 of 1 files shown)


ramaro (approved):

ramaro:
@timReynolds thank you

PR #1522, opened 2026-06-08
Fix Kadet inventory_global default boxing

timReynolds:
## Summary

Fix Kadet `inventory_global()` default-box behavior by separating the cached Kadet inventory wrappers for lazy/default-box and non-lazy access.

## Why this changed behavior between 0.34 and 0.35

The shared wrapper behavior already existed in 0.34: `inventory()` delegated through `inventory_global(lazy)` and both helpers reused `cached.inventory_global_kadet`. That meant the first caller decided whether the cached wrapper used `default_box=True` or `False`.

In 0.35, Kadet input caching added a preflight inventory digest path before the Kadet module is executed. That path calls `inventory_digest()`, which calls `inventory()` with the default `lazy=False`. When compile cache is enabled, Kapitan itself can therefore initialize the shared global Kadet wrapper as non-lazy before user Kadet code later calls `inventory_global()`.

That makes `inventory_global()` unexpectedly lose default-box behavior in 0.35 for users who compile Kadet with cache enabled.

## What changed

- Cache Kadet global inventory wrappers by `default_box` mode instead of using a single shared wrapper.
- Make `inventory_global()` default to the default-box wrapper.
- Preserve `inventory()` default behavior as non-lazy unless called with `lazy=True`.
- Clear Kadet wrapper/input caches in `reset_cache()` so wrappers cannot outlive refreshed global inventory state.
- Add regression coverage for the 0.35 cache-preflight call order where `inventory_digest()` runs before `inventory_global()`.

## Validation

- `uv run pytest tests/test_kadet.py tests/test_cached.py --no-cov`
- `uv run ruff check kapitan/inputs/kadet.py kapitan/cached.py tests/test_kadet.py tests/test_cached.py`

### DIFF
--- a/kapitan/inputs/kadet.py
+++ b/kapitan/inputs/kadet.py
@@ -34,16 +34,21 @@ search_paths = contextvars.ContextVar("current search_paths in thread")
 current_target = contextvars.ContextVar("current target in thread")
 
 
-@cache
-def inventory_global(lazy=False):
+def _inventory_global(default_box=False):
     # At hoc inventory for kadet
-    if not cached.inventory_global_kadet:
-        cached.inventory_global_kadet = Dict(cached.global_inv, default_box=lazy)
-    return cached.inventory_global_kadet
+    if default_box not in cached.inventory_global_kadet:
+        cached.inventory_global_kadet[default_box] = Dict(
+            cached.global_inv, default_box=default_box
+        )
+    return cached.inventory_global_kadet[default_box]
+
+
+def inventory_global(lazy=True):
+    return _inventory_global(default_box=lazy)
 
 
 def inventory(lazy=False):
-    return inventory_global(lazy)[current_target.get()]
+    return _inventory_global(default_box=lazy)[current_target.get()]
 
 
 def topics(name=None, lazy=False):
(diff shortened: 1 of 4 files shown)


kapitan/inputs/kadet.py
```diff
@@ -34,16 +34,21 @@
...
-        cached.inventory_global_kadet = Dict(cached.global_inv, default_box=lazy)
-    return cached.inventory_global_kadet
+    if default_box not in cached.inventory_global_kadet:
+        cached.inventory_global_kadet[default_box] = Dict(
+            cached.global_inv, default_box=default_box
+        )
+    return cached.inventory_global_kadet[default_box]
+
+
+def inventory_global(lazy=True):
```
ramaro, line 46: @timReynolds are we changing it to default false because of the behaviour (badly) introduced by kadet caching? If so, wouldn't this break when not using cache?

timReynolds: Probably. I'm less sure about this change but when caching was introduced the behaviour changed as outlined which makes the upgrade path harder. Maybe I'd be best to define an example in an issue so someone with better understanding of the code can make a more considered change

PR #1523, opened 2026-06-08
Respect container CPU limits for compile parallelism

timReynolds:
## Summary

This change updates Kapitan's default parallelism calculation so it uses the number of CPUs available to the current process rather than the underlying machine CPU count.

That matters in virtualised environments such as Kubernetes, where the host may expose many CPUs but the container is only allowed to use a smaller quota. Previously Kapitan could pick too many workers from the underlying CPU count, oversubscribing the container and slowing build times.

The implementation intentionally avoids adding a runtime library for this small calculation, reducing dependency bloat. It uses the Python standard library plus Linux cgroup v1/v2 CPU quota files and process affinity when available.

## Validation

- `uv run pytest tests/test_utils.py --no-cov -q`
- `uv run pytest tests/test_targets.py --no-cov -q`
- `uv run --extra omegaconf pytest tests/test_omegaconf.py --no-cov -q`
- `uv run ruff check kapitan/utils.py kapitan/targets.py kapitan/inventory/backends/omegaconf/__init__.py kapitan/cli.py tests/test_utils.py`
- `git diff --check`

### DIFF
--- a/docs/pages/commands/kapitan_compile.md
+++ b/docs/pages/commands/kapitan_compile.md
@@ -219,7 +219,8 @@ library), enabled with `--yaml-use-rapidyaml`.
           --validate            validate compile output against schemas as specified
                                 in inventory
           --parallelism INT, -p INT
-                                Number of concurrent compile processes, default is 4
+                                Number of concurrent compile processes, default is
+                                min(len(targets), available CPU count)
           --indent INT, -i INT  Indentation spaces for YAML/JSON, default is 2
           --refs-path REFS_PATH
                                 set refs path, default is "./refs"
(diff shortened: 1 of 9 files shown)


ramaro:
@timReynolds thanks for this, I agree your approach is better. Please can you fix the conflicts? We can merge once ready.

timReynolds:
@ramaro done, this should be good to merge now and I'd expect people will see a performance boost when using Kapitan in a virtualised environment if the underlying machine has significantly more CPUs than allocated which was the case from testing

ramaro (approved):

PR #1524, opened 2026-06-08
 Improve OmegaConf inventory rendering performance

timReynolds:
## Summary

Depends on #1523. This branch is built on top of #1523's `codex/container-aware-parallelism` changes and keeps its container-aware worker count handling via `available_cpu_count()`.

The OmegaConf inventory renderer currently pays high multiprocessing overhead on large inventories with many shared classes:

- worker results are written through an `mp.Manager().dict()`, adding synchronized IPC and proxy serialization
- each work item passes the inventory instance through the task payload
- shared class trees are repeatedly parsed and merged for each target even when every target uses the same class stack

This change removes the manager proxy, returns rendered targets directly from workers with `imap_unordered()`, seeds one inventory instance per worker via a pool initializer, and caches class-level config trees through `get_class_config()`.

## Performance Context

I ran a local synthetic benchmark that models the workload this change is intended to help: 240 OmegaConf targets, all sharing one root class that expands into 60 class files with 120 keys each, using 8 workers.

Results:

| Path | Time |
| --- | ---: |
| Old Manager/per-target class merge path | 21.525s |
| New worker-return/class-cache path | 8.775s |

That is about **2.45x faster**, or **59.2% less wall time**, on that shared-class-stack workload.

The gain is workload-dependent. A weaker synthetic case where targets referenced many leaf classes directly showed only about 3% improvement, because there was much less reusable class-tree work to cache. The expected benefit is highest for large inventories where many targets share the same nested class hierarchy.

## Validation

Added tests covering the optimization contract:

- `render_targets()` does not use `mp.Manager()` and receives rendered targets through worker return values
- the worker pool is initialized with the inventory once per worker
- shared class config is loaded once while target-specific interpolation still resolves independently per target

Checks run:

- `uv run ruff check kapitan/inventory/backends/omegaconf/__init__.py tests/test_omegaconf.py`
- `uv run --extra omegaconf pytest tests/test_omegaconf.py -q --no-cov` -> 8 passed
- `uv run --extra omegaconf pytest tests/test_inventory_backend_examples.py -q --no-cov` -> 51 passed, 13 skipped because `reclass_rs` is unavailable locally

### DIFF
--- a/kapitan/inventory/backends/omegaconf/__init__.py
+++ b/kapitan/inventory/backends/omegaconf/__init__.py
@@ -69,6 +69,8 @@ class OmegaConfTarget(InventoryTarget):
 
 
 class OmegaConfInventory(Inventory):
+    _instance = None
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs, target_class=OmegaConfTarget)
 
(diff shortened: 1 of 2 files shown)


timReynolds:
@ramaro would be great if you could take a look at this, it significantly improves performance but should be checked by someone with a better understanding of the codebase than I

Issue #1510, opened 2026-05-26
[refactor] Replace global cached.py with explicit context objects


Moep90:
## Problem

`kapitan/cached.py` is a module-level bag of 20+ mutable globals that get used as hidden dependencies all over the codebase:

```python
inv: dict[str, Any] = {}
global_inv: dict[str, Any] = {}
gpg_obj: Any = None
awskms_obj: Any = None
azkms_obj: Any = None
ref_controller_obj: Any = None
revealer_obj: Any = None
args: Namespace = Namespace()
...
```

This causes a few problems:

1. Brittle multiprocessing. `compile_targets` pickles the whole cache via `cached.as_dict()` and ships it to every worker. That payload includes the full inventory, all backend clients, and the CLI args, so it's big.
2. Hidden coupling. `kapitan/resources.py:383-425` reads `cached.args.inventory_backend` and writes back to `cached.inv`. You can't call this code as a library without first stuffing CLI args into the global cache.
3. Untestable Jinja2 filters. `kapitan/jinja2_filters.py:96-99` reads `cached.args.reveal` and `cached.revealer_obj` directly, so the filters can't be unit-tested without mutating global state.
4. Untestable YAML presenters. `kapitan/utils.py:183` `null_presenter()` reads `cached.args.yaml_dump_null_as_empty` directly.
5. Unsafe multiprocessing rehydration. `kapitan/targets.py:312-313` calls `cached.from_dict(globals_cached)` inside workers, which leaks all the parent state into every subprocess.

## Proposed solution

Introduce a `CompileContext` (or `KapitanContext`) dataclass holding `inventory`, `ref_controller`, `revealer`, `args`, and the backend clients. Pass it explicitly down the call stack instead of reaching into `cached`.

### Concrete steps

1. Create `kapitan/context.py` with a `CompileContext` dataclass.
2. Replace `cached.gpg_obj`, `cached.awskms_obj`, etc. with memoized factory functions that take a config object, or use `@functools.lru_cache` on functions with explicit parameters.
3. Change `get_inventory()` to accept `backend_id`, `compose_target_name`, and `migrate` explicitly. The CLI layer reads `cached.args` and passes them; library code shouldn't have to.
4. Register Jinja2 filters as closures that capture context: `env.filters["reveal_maybe"] = make_reveal_maybe(reveal_flag, revealer_instance)`.
5. For multiprocessing, pass only what's needed (inventory path, target config) instead of the entire global cache dict.

## Expected impact

[...]

PR #1531, opened 2026-06-12
refactor(cached): drop orphaned KMS globals + trim worker payload (#1510 tail, PR-4+5)

Moep90:
## ⚠️ Stacked, merge after #1528, #1529, #1530

This is the dependent cleanup tail of the #1510 series. It's built on top of #1528 + #1529 + #1530, so this PR's diff currently includes those three. They'll drop out of the diff automatically as they merge, so merge this last. It combines the planned PR-4 (worker payload) and PR-5 (delete globals) into one PR, per maintainer preference.

## What (the tail-specific changes)

Now that `gkms`/`awskms`/`azkms` clients are memoized in their own modules via `functools.cache` (from #1528), the `cached.*_obj` mirrors are dead weight:

- remove the mirror writes from `gkms_obj`/`awskms_obj`/`azkms_obj` (return the memoized client directly)
- delete `cached.gkms_obj` / `cached.awskms_obj` / `cached.azkms_obj` and their entries in `reset_cache` / `as_dict` / `from_dict`
- update `test_cached.py` accordingly

Multiprocessing payload (the PR-4 part): the per-worker seed dict (`cached.as_dict`) no longer carries the three KMS client objects. Workers rebuild them lazily via `functools.cache` instead of getting them in the pickled payload. `reset_cache` still flushes those caches through the clearer registry. (Note: #1519 already made worker seeding once-per-worker, so this is a payload trim rather than a rework.)

## Deliberately retained (out of scope)

- `gpg_obj`, because its `*args`/`**kwargs` first-wins semantics need separate care.
- `cached.args` fallbacks (in `get_inventory`, `null_presenter`, `reveal_maybe`) and the `cached.inv` store. Removing these means wiring explicit params through `cli.py` / `targets.py` / `refs/cmd_parser.py`, which overlaps the in-flight OmegaConf perf PRs (#1524/#1525). Left as a further step.

## Note for #1527

#1527's `test_kms_factories.py` asserts `cached.gkms_obj is <client>` (the mirror this PR removes). When both land, those two `assertIs(cached.*_obj, ...)` lines should be dropped. The construct-once + memoize contract they check is preserved, only the global mirror is gone. Tracked as follow-up #1532.

## Verification

`tests/test_cached.py tests/test_refs.py tests/test_azure.py tests/test_targets.py tests/test_compile.py::CompileTestResourcesTestObjs` gives 84 passed; ruff lint + format clean.

### DIFF
--- a/kapitan/cached.py
+++ b/kapitan/cached.py
@@ -44,6 +44,16 @@ kapitan_input_kadet = None
 # bumps the same shared CacheMetrics across processes.
 input_cache_metrics: dict | None = None
 
+# Memoized secret-handler factories register their ``cache_clear`` here so
+# ``reset_cache`` can flush them without importing the optional cloud SDK
+# modules. Populated lazily as each refs.secrets.* module is imported.
+_handler_cache_clearers: list = []
+
+
+def register_handler_cache_clearer(clear_fn) -> None:
+    """Register a callable that clears a memoized secret-handler factory cache."""
+    _handler_cache_clearers.append(clear_fn)
+
 
 def reset_cache():
     global \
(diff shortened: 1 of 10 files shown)


Moep90:
Remaining cached.py teardown (cached.inv store, cached.args fallbacks, gpg_obj migration, serialization removal) is tracked in #1532 — to be done after this series merges.

PR #1551, opened 2026-06-13
fix: clean up temp dirs and close fd leaks in Helm, Kustomize, Cuelang

### LINKED ISSUE #1509: [bug] Temp directory and file descriptor leaks in Helm, Kustomize, Cuelang input compilers
## Problem

Three input compilers create temp directories/files and never clean them up. On large compilation runs this leaks disk space and file descriptors.

### Leaked temp directories

- `Helm.compile_file` (`kapitan/inputs/helm.py:62`) calls `tempfile.mkdtemp()` for rendered chart output, never deleted.
- `Kustomize.compile_file` (`kapitan/inputs/kustomize.py:98`) calls `tempfile.mkdtemp()` for working dir, never deleted.
- `Kustomize.compile_file` (`kapitan/inputs/kustomize.py:144`) calls `tempfile.mkdtemp()` for kustomize build output, never deleted.
- `Cuelang.compile_file` (`kapitan/inputs/cuelang.py:43`) calls `tempfile.mkdtemp()` for working dir, never deleted.

### Leaked file descriptors

- `write_helm_values_file` (`kapitan/inputs/helm.py:232`) calls `tempfile.mkstemp()`, which returns an open fd that gets discarded (`_`), so the fd leaks.
- `render_chart` (`kapitan/inputs/helm.py:194`) has the same `mkstemp` fd leak.

## Proposed solution

1. Wrap the `tempfile.mkdtemp()` calls in `tempfile.TemporaryDirectory()` context managers (or `try/finally`).

2. Close the `mkstemp` fd right after creation:
   ```python
   fd, helm_values_file = tempfile.mkstemp(...)
   os.close(fd)
   ```
   Or switch to `tempfile.NamedTemporaryFile` with `delete=False` and close it properly.

## Expected impact

Stops disk space and fd exhaustion on long-running or large compilation jobs.

kaizeenn:
## Summary

Fixes #1509.

Three input compilers created temp directories with `tempfile.mkdtemp()` and never cleaned them up. Two `tempfile.mkstemp()` calls discarded the returned file descriptor (`_`) without closing it. On large or long-running compilation jobs this causes disk-space exhaustion and file-descriptor leaks.

### `kapitan/inputs/helm.py`
- `compile_file`: replace `mkdtemp()` with `tempfile.TemporaryDirectory()` context manager — rendered-chart dir is now always deleted on exit, even if an exception is raised.
- `render_chart`: close the `mkstemp()` fd immediately with `os.close(fd)`, then `os.unlink()` the temp file after its contents are read.
- `write_helm_values_file`: close the `mkstemp()` fd immediately with `os.close(fd)` before opening the file for writing.

### `kapitan/inputs/kustomize.py`
- `compile_file`: replace two separate `mkdtemp()` calls with a single `with tempfile.TemporaryDirectory() as temp_dir, tempfile.TemporaryDirectory() as output_dir:` block; consolidate the previously split `try/except` blocks into one, re-raising `KustomizeTemplateError` directly so it is not double-wrapped.

### `kapitan/inputs/cuelang.py`
- `compile_file`: replace `mkdtemp()` with `tempfile.TemporaryDirectory()` context manager.

## Testing

All pre-existing test failures (helm/kustomize/cuelang suites require `helm`/`kustomize`/`cue` binaries not present in CI) are identical before and after this change — no regressions introduced.

### DIFF
--- a/kapitan/inputs/helm.py
+++ b/kapitan/inputs/helm.py
@@ -59,35 +59,35 @@ class Helm(InputType):
         if config.kube_version:
             helm_flags["--api-versions"] = config.kube_version
 
-        temp_dir = tempfile.mkdtemp()
-        # save the template output to temp dir first
-        _, error_message = render_chart(
-            chart_dir=input_path,
-            output_path=temp_dir,
-            helm_path=helm_path,
-            helm_params=helm_params,
-            helm_values_file=helm_values_file,
-            helm_values_files=helm_values_files,
-            helm_flags=helm_flags,
-        )
-        if error_message:
-            raise HelmTemplateError(error_message)
-        # Iterate over all files in the temporary directory
+        with tempfile.TemporaryDirectory() as temp_dir:
+            # save the template output to temp dir first
+            _, error_message = render_chart(
+                chart_dir=input_path,
+                output_path=temp_dir,
+                helm_path=helm_path,
+                helm_params=helm_params,
+                helm_values_file=helm_values_file,
+                helm_values_files=helm_values_files,
+                helm_flags=helm_flags,
+            )
+            if error_message:
+                raise HelmTemplateError(error_message)
+            # Iterate over all files in the temporary directory
 
-        walk_root_files = os.walk(temp_dir)
-        for current_dir, _, files in walk_root_files:
-            for file in files:  # go through all the template files
-                rel_dir = os.path.relpath(current_dir, temp_dir)
-                rel_file_name = os.path.join(rel_dir, file)
-                full_file_name = os.path.join(current_dir, file)
-                # Open each file and write its content to the compilation path
-                with open(full_file_name, encoding="utf-8") as f:
-                    file_path = os.path.join(compile_path, rel_file_name)
-                    os.makedirs(os.path.dirname(file_path), exist_ok=True)
-                    item_value = [
-                        doc for doc in yaml.safe_load_all(f) if doc is not None
-                    ]
-                    self.to_file(config, file_path, item_value)
+            walk_root_files = os.walk(temp_dir)
+            for current_dir, _, files in walk_root_files:
+                for file in files:  # go through all the template files
+                    rel_dir = os.path.relpath(current_dir, temp_dir)
+                    rel_file_name = os.path.join(rel_dir, file)
+                    full_file_name = os.path.join(current_dir, file)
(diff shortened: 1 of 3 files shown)


Moep90 (commented):
The `mkdtemp` → `TemporaryDirectory()` conversions and both fd closes look correct, and the kustomize `except KustomizeTemplateError: raise` ahead of the generic handler nicely avoids the double-wrap. One inline nit on exception-safety below; non-blocking.

Two related leaks I found while reviewing, both pre-existing and out of scope here, were opened as separate PRs rather than expanding this one:

- #1555 `write_helm_values_file()`'s temp file is never unlinked by either caller. The fd half is covered here; the file lifecycle isn't. Complementary to this PR, it does touch `compile_file`, so expect a small conflict with whichever of us merges second.
- #1554  unrelated to #1509: `cuelang.py` raises `KustomizeTemplateError` (copy-paste) when `CuelangTemplateError` already exists in `kapitan/errors.py`.

Nice cleanup overall.

kapitan/inputs/helm.py
```diff
@@ -193,11 +193,14 @@ def render_chart(
...
     if output_path == "-":
-        _, helm_output = tempfile.mkstemp(".helm_output.yml", text=True)
+        fd, helm_output = tempfile.mkstemp(".helm_output.yml", text=True)
+        os.close(fd)
         with open(helm_output, "w+") as f:
             error_message = helm_cli(helm_path, args, stdout=f)
             f.seek(0)
-            return (f.read(), error_message)
+            content = f.read()
+        os.unlink(helm_output)
```
Moep90, line 202: `os.unlink` here isn't exception-safe if `helm_cli` or `f.read()` above raises; this line is skipped and the temp file leaks (the same bug class this PR targets). Wrap in `try/finally`, or let a context manager delete on exit:

```python
with tempfile.NamedTemporaryFile(suffix=".helm_output.yml", mode="w+") as f:
    error_message = helm_cli(helm_path, args, stdout=f)
    f.seek(0)
    return (f.read(), error_message)
```

Low real-world risk since `helm_cli` returns the error rather than raising non-blocking.

ramaro:
@kaizeenn thanks for sending this in. However this approach seems to repeat the same formula in the mentioned input types. Perhaps we can abstract this in the base class or have common generic functions to avoid duplication?

PR #1553, opened 2026-06-15
fix: fix kapitan version in container image

### LINKED ISSUE #1552: [bug]: Container image 0.36.0 reporting kapitan version as 0.0.0
### Description of the bug

Container image 0.36.0 reporting kapitan version as 0.0.0:

```shell
docker runkapicorp/kapitan:0.36 --version                                                                                                                                                                                        
0.0.0
```

### Steps to reproduce

```bash
1. docker runkapicorp/kapitan:0.36 --version
```

### What happened?

Since the automated versioning support for the python package was introduced, the container image is reporting version 0.0.0

### Additional context

_No response_

### Version (kapitan)

master

### Version (Python)

3.13

### Inventory backend

None

- [ ] pyenv
- [ ] venv
- [ ] no environment

### OS information

_No response_

### Relevant log output

```bash

```

ramaro:
## Summary

fix kapitan version in container image

Fixes #1552

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,13 @@ COPY ./pyproject.toml ./pyproject.toml
 COPY ./uv.lock ./uv.lock
 COPY ./README.md ./README.md
 COPY ./kapitan/version.py ./kapitan/version.py
-# uv-dynamic-versioning resolves the package version from git tags at build
-# time; without .git it falls back to 0.0.0 (see pyproject.toml).
-COPY ./.git ./.git
+
+# uv-dynamic-versioning normally derives the version from git tags, which
+# would require shipping .git into the build context. Instead the caller
+# computes the version once outside Docker and passes it in via this
+# build-arg; the env var tells uv-dynamic-versioning to use it verbatim.
+ARG KAPITAN_VERSION
+ENV UV_DYNAMIC_VERSIONING_BYPASS=${KAPITAN_VERSION}
 
 RUN uv sync --locked --all-extras
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,9 @@ COPY ./pyproject.toml ./pyproject.toml
 COPY ./uv.lock ./uv.lock
 COPY ./README.md ./README.md
 COPY ./kapitan/version.py ./kapitan/version.py
+# uv-dynamic-versioning resolves the package version from git tags at build
+# time; without .git it falls back to 0.0.0 (see pyproject.toml).
+COPY ./.git ./.git
 
 RUN uv sync --locked --all-extras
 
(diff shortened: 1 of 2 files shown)


uberspot (approved):

Dockerfile
```diff
@@ -33,6 +33,9 @@ COPY ./pyproject.toml ./pyproject.toml
 COPY ./uv.lock ./uv.lock
 COPY ./README.md ./README.md
 COPY ./kapitan/version.py ./kapitan/version.py
+# uv-dynamic-versioning resolves the package version from git tags at build
+# time; without .git it falls back to 0.0.0 (see pyproject.toml).
+COPY ./.git ./.git
```
Moep90: Can we skip `.git` entirely? `uv-dynamic-versioning` has a bypass env var, and version-info already knows the real version. Just pass it as a build-arg:

```dockerfile
ARG KAPITAN_VERSION
ENV UV_DYNAMIC_VERSIONING_BYPASS=${KAPITAN_VERSION}
```

```yaml
build-args: |
  PYTHON_VERSION=${{ matrix.python-version }}
  KAPITAN_VERSION=${{ needs.<version-job>.outputs.version }}
```

No `.git`, no cache busting, no ownership mess. Just double-check the env var name against whatever `uv-dynamic-versioning` we've got pinned in `uv.lock`.

Side note if we do end up keeping the `.git` copy: it's sitting before `uv sync --locked`, which means every commit busts the dependency layer cache. Move it down next to `uv pip install .` instead.

Moep90 (changes requested):

Dockerfile
ramaro: yeah that is cleaner! Let me fix

ramaro: fixed, please re-review

Moep90:
## One thing I'd add before merging

The `Test Kapitan image` step runs `docker run -t --rm local-test --version` but never asserts on the output. That means it prints `0.0.0` and still passes green, which is exactly the failure mode in #1552. If the build-arg wiring ever breaks (job renamed, output goes empty, arg dropped), the bug comes back silently. A small guard turns should ensure that it stays fixed:

```bash
GOT=$(docker run --rm local-test --version | tr -d '[:space:]')
EXP="${{ needs.version-info.outputs.version }}"
[ "$GOT" = "$EXP" ] || { echo "version mismatch: got '$GOT' want '$EXP'"; exit 1; }
```

## Minor, take it or leave it

The `version-info` job has no `needs` or `if:` gate, so it runs on every event, including doc-only PRs where the build is skipped. It's only ~9s plus a depth-0 checkout, so the cost is trivial. Worth wiring into the `has-code-changes` pattern only if you want it consistent with the rest of the workflow.

Happy to approve once the version assertion is in.

ramaro:
thanks for the review @Moep90 @uberspot , I added the suggested fixes

Moep90 (approved):

PR #1554, opened 2026-06-15
fix(cuelang): raise CuelangTemplateError instead of KustomizeTemplateError

Moep90:
What

The Cuelang input compiler raised `KustomizeTemplateError` when a `cue export` call failed:

```python
from kapitan.errors import KustomizeTemplateError
...
if result.returncode != 0:
    err = f"Failed to run CUE export: {result.stderr}"
    raise KustomizeTemplateError(err)
```

`CuelangTemplateError` already exists in `kapitan/errors.py` but was never used — this was a copy-paste leftover from the Kustomize input that Cuelang was modelled on. I switched it to the correct exception type.

Why

Anything catching `CuelangTemplateError` (or filtering errors by input type) silently misses Cuelang failures today, and a Cuelang error surfacing as a "Kustomize" error is confusing to debug. Both subclass `KapitanError`, so broad handlers are unaffected — this only sharpens the specific type.

Approach

- `kapitan/inputs/cuelang.py`: import and raise `CuelangTemplateError`.
- No behaviour change beyond the exception class.

Verification

- Added `test_compile_file_raises_cuelang_error_on_failure`, which mocks `subprocess.run` to force a non-zero return code and asserts `CuelangTemplateError` is raised. It needs no `cue` binary, so it runs in CI.
- `uv run pytest tests/test_cuelang_input.py::CuelangInputTest::test_compile_file_raises_cuelang_error_on_failure` passes.
- `ruff check` / `ruff format --check` clean on both files.

Spotted while reviewing #1551.


Moep90:
Coordination note: adjacent to #1551 (fixes #1509) in the Cuelang compiler, but orthogonal — #1551 handles fd/temp-dir leaks, this only corrects the raised exception class (CuelangTemplateError vs KustomizeTemplateError). No conflict expected; flagging for review ordering.

PR #1555, opened 2026-06-15
fix(helm): delete temporary helm values file after rendering

Moep90:
What

`write_helm_values_file()` creates a temp file with `tempfile.mkstemp(".helm_values.yml")` and returns its path. Both callers — `Helm.compile_file` and `HelmChart.load_chart` — hand that path to `render_chart` as the `-f` values file but never delete it:

```python
helm_values_file = None
if config.helm_values:
    helm_values_file = write_helm_values_file(config.helm_values)
...
render_chart(..., helm_values_file=helm_values_file, ...)
# helm_values_file is never removed
```

So every helm input compile that sets `helm_values` leaves one `*.helm_values.yml` behind in the system temp directory.

Why

On large inventories or long-running / repeated compile jobs these files accumulate in `/tmp`, wasting disk and inodes until the next reboot or temp sweep. It is the temp-file half of the leak class tracked in #1509; that issue (and #1551) addresses the unclosed file descriptors and the leaked temp *directories*, but the values file itself is still left on disk.

Approach

- Wrap each `render_chart(...)` call in `try/finally` and `os.unlink(helm_values_file)` in the `finally`, in both `Helm.compile_file` and `HelmChart.load_chart`.
- The file is only read by helm during `render_chart`, so removing it right after that call is safe, and `finally` guarantees cleanup even when rendering raises.
- Deliberately small/surgical to stay independent of #1551 (which rewrites the surrounding temp-dir handling); the two changes are complementary.

Verification

- Added `HelmValuesFileCleanupTest` with two binary-free tests: they mock `render_chart`, capture the values-file path it receives, and assert the file no longer exists after a successful render and after a render that raises.
- Confirmed the tests fail against the unpatched source and pass with the fix.
- `ruff check` / `ruff format --check` clean on both files.

Related to #1509. Spotted while reviewing #1551.


Moep90:
Coordination note: overlaps #1551 (fixes #1509) in the Helm temp-handling area. This PR is complementary — it deletes the rendered *values* file, which #1551 does not touch. If #1551 merges first I'll rebase this on top; the two changes don't conflict in intent.

PR #1558, opened 2026-06-16
fix(kadet): exclude compile_path from cache key to restore cache hits

timReynolds:
PR #1520 correctly added input_params to the cache key to prevent collisions between inputs that differ only by input_params, but it also moved the compile_path setdefault *before* the hash computation. Since compile_path is a per-run tempfile.mkdtemp() path it changes on every invocation, making the cache key unstable and causing every lookup to miss.

Fix: copy config.input_params (preserving mutation-safety) and include it in the hash before setdefault injects compile_path. The setdefault call is moved back to after the cache lookup so the volatile tempdir never appears in the key. User-supplied input_params (without the auto-injected compile_path) continue to differentiate cache entries, so collision safety from #1520 is preserved.

Adds regression tests asserting compile_path is absent from the hashed dict and that two runs with different compile_paths produce identical hash arguments.

https://claude.ai/code/session_011AJrbdseq5WcYsdLswJoh2

## Summary

Fixes #

## Test plan

- [ ] Tests added or updated
- [ ] Documentation updated (if user-facing behavior changed)
- [ ] `make lint` passes
- [ ] `make format` passes (or no formatting changes needed)
- [ ] Full test suite passes (`uv run pytest tests/ --no-cov`)

### DIFF
--- a/kapitan/inputs/kadet.py
+++ b/kapitan/inputs/kadet.py
@@ -135,12 +135,11 @@ class Kadet(InputType):
         inputs_hash = None
         output_obj = None
 
-        # set compile_path allowing kadet functions to have context on where files
-        # are being compiled on the current kapitan run
-        # we only do this if user didn't pass its own value
-        input_params.setdefault("compile_path", compile_path)
-
         if cache_obj := self.cacheable():
+            # Hash input_params before setdefault injects compile_path below, so the
+            # volatile per-run tempdir never appears in the key. Any compile_path the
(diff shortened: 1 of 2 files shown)


ramaro:
@timReynolds thanks for the fix! I have also experienced the mentioned regression. Please can you ensure the ruff formatting passes? Then this is good to go.

timReynolds:
@ramaro done, this should now be good to merge

ramaro (approved):

PR #1559, opened 2026-06-16
test: bound vault test-server startup so an unreachable VAULT_ADDR can't hang the suite

Moep90:
## What

The vault test-server helper (`tests/vault_server.py`) polled the server in a `while not initialized:` loop with **no timeout** and `except: continue`. If the vault it talks to is unreachable, that loop never exits.

When `VAULT_ADDR` points at a vault that doesn't resolve/respond (common locally — a stale env var, VPN off, no Docker), the whole test suite **hangs forever** instead of failing. It bites in `test_cli.py`, `test_vault.py` and `test_vault_transit.py`, whose `setUpClass` builds the shared vault server.

## Why

I hit this running the full suite locally: it stalled on `test_cli.py::CliFuncsTest::test_cli_inventory` for 20+ minutes. The process wasn't deadlocked — it was retrying a health check against an unreachable `VAULT_ADDR` in an unbounded loop, logging `status is {}` / `NameResolutionError` endlessly. CI never sees it because vault is reachable there.

## Approach

- `vault_server.py`: bound **both** startup loops (container-running wait and the health-check/init wait) with a deadline (`KAPITAN_VAULT_SETUP_TIMEOUT`, default `60`s). On timeout raise `VaultServerError` instead of looping. Cache construction failures so each dependent test class skips instantly rather than re-waiting the full timeout.
- `test_cli` / `test_vault` / `test_vault_transit`: catch `VaultServerError` in `setUpClass` and `SkipTest` — the same graceful-skip behaviour the reclass-rs tests already use when their backend is unavailable. Teardown guarded for the skipped case.
- New `tests/test_vault_server.py`: regression test asserting that an unreachable `VAULT_ADDR` makes `VaultServer()` **raise quickly** instead of hanging.

## Verification

- Regression test: red→green (infinite hang → raises in ~4s).
- Full non-slow suite on this branch with an unreachable `VAULT_ADDR`: **522 passed, 36 skipped in ~33s** (previously hung indefinitely). The 35 vault tests skip with a clear reason.
- `ruff check` / `ruff format` clean; pre-commit hooks pass.

**CI is unaffected**: vault is reachable there, so the loops exit immediately and nothing skips — behaviour is unchanged on CI.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -20,7 +20,7 @@ from kapitan.cli import build_parser
 from kapitan.cli import main as kapitan
 from kapitan.errors import KapitanError
 from kapitan.refs.secrets.vaultkv import VaultSecret
-from tests.vault_server import get_shared_vault_server
+from tests.vault_server import VaultServerError, get_shared_vault_server
 
 
 REFS_PATH = tempfile.mkdtemp()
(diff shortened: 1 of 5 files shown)


Moep90:
Relates to #1512 — this is the test-isolation/hang half of that test-infra issue (bounding the vault test-server startup).

PR #1564, opened 2026-06-18
feat: require consume declaration and mix into kadet cache key

### LINKED ISSUE #1567: [feat]: set explicit topic consumers for effective caching
## Problem

Topics aggregate parameters across targets, which creates a cross-target dependency the input cache can't see: changes to a producer target don't move the consumer's cache key, so kadet returns stale output.

The dependencies need to be made explicit.

Consumers must declare `kapitan.topics.<name>.consume: true` for every topic they read.

- `topics(name)` raises a `CompileError` when called from a target that hasn't declared that topic.
- `topics()` (no name) requires every topic in the inventory to be declared on the calling target.
- Outside a compile context (e.g. `kapitan inventory --topics`), the check is skipped.

The kadet input cache mixes a digest of each declared topic's aggregated view into the consumer's cache key, so producer-side changes invalidate the consumer's cache automatically. Targets that consume no topics keep their existing cache keys.

### Use case

Ensure cache is invalidated when consumed topics change.

### Contribution

- [x] I would be willing to contribute this feature

### Additional context

_No response_

ramaro:
## Summary

Fixes #1567 

## Test plan

- [x] Tests added or updated
- [x] Documentation updated (if user-facing behavior changed)
- [x] `make lint` passes
- [x] Full test suite passes (`uv run pytest tests/ --no-cov`)

### DIFF
--- a/kapitan/inputs/jinja2.py
+++ b/kapitan/inputs/jinja2.py
@@ -11,7 +11,7 @@ import os
 from kapitan import cached
 from kapitan.inputs.base import CompiledFile, InputType
 from kapitan.inventory.model.input_types import KapitanInputTypeJinja2Config
-from kapitan.topics import topics
+from kapitan.topics import current_target, topics
 from kapitan.utils import render_jinja2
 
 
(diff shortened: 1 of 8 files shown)


ademariag (approved):

Moep90 (changes requested):
See inline comments.

kapitan/topics.py
```diff
@@ -13,31 +13,124 @@
...
+            if name not in declared:
+                raise CompileError(_hint(target, [name]))
+        else:
+            undeclared = set(all_topics) - declared
+            if undeclared:
+                raise CompileError(_hint(target, undeclared))
+
     if not name:
         return all_topics
     return all_topics.get(name, dict(_EMPTY_TOPIC))
```
Moep90, line 93: `dict(_EMPTY_TOPIC)` only copies the top level, so the returned `["parameters"]` is still the same object as the module global. A caller that mutates `result["parameters"]["targets"]` would pollute it for everyone. This predates the PR and it is safe here since the digest path only reads, but `copy.deepcopy` would be safer down the line.

kapitan/inputs/jinja2.py
```diff
@@ -42,6 +42,9 @@ def compile_file(
 
         reveal = self.args.reveal
         target_name = self.target_name
+        # Publish the calling target so topics() can enforce consume declarations
+        # when called from a template.
+        current_target.set(target_name)
```
Moep90: `current_target` gets set here but never reset. Pool workers are reused across targets, and each input sets it before evaluating, so the target being compiled is always correct. The value sticks around afterward though. If something other than an input calls `topics()` later in the same worker, it picks up a stale target. A `try/finally: current_target.reset(token)` would close that off, and it is what the tests already do in tearDown.

kapitan/inputs/jsonnet.py
```diff
@@ -72,6 +73,10 @@ def compile_file(
 
         use_go = self.args.use_go_jsonnet
         ext_vars = {"target": self.target_name}
+        # Publish the calling target so the ``topics`` native callback (see
+        # kapitan.resources.resource_callbacks) can enforce consume declarations
+        # when called from jsonnet code.
+        current_target.set(self.target_name)
```
Moep90: `current_target` gets set here but never reset. Pool workers are reused across targets, and each input sets it before evaluating, so the target being compiled is always correct. The value sticks around afterward though. If something other than an input calls `topics()` later in the same worker, it picks up a stale target. A `try/finally: current_target.reset(token)` would close that off, and it is what the tests already do in tearDown.

kapitan/inventory/inventory.py
```diff
@@ -102,17 +102,56 @@ def topics(self) -> dict:
...
+        target = self.targets.get(target_name)
+        if target is None:
+            return set()
+        target_topics = getattr(target.parameters.kapitan, "topics", None) or {}
+        declared: set[str] = set()
+        for topic_name, topic_values in target_topics.items():
+            consume = getattr(topic_values, "consume", None)
+            if consume is None and isinstance(topic_values, dict):
+                consume = topic_values.get("consume")
+            if consume is True:
```
Moep90, line 151: The strict `consume is True` check makes sense, since it rejects the string `"true"`. One thing though: the new tests use a MockInv rather than the real backends. Was omegaconf tested end to end? I am wondering whether omegaconf hands back a real Python `bool` here, or whether it could coerce the value and make this check silently fail for omegaconf consumers. Worth confirming, since inventory changes should hold across all backends.

kapitan/topics.py
ramaro: I don't fully agree here - simply because this structure is pretty much readonly in this context. Guarding it would be unnecessary and would take extra time

kapitan/inputs/jinja2.py
ramaro: fixed, thanks!

kapitan/inputs/jsonnet.py
ramaro: fixed, thanks!

kapitan/inventory/inventory.py
ramaro: this should not be a problem, since it is specific to topics and also a battle tested use case - type coertion in omegaconf should not be a problem for bools that are properly. set. i I tested on my real use case end to end and this

Moep90: Good enough for me — verified the `consume is True` check on the current branch and your e2e test covers the real path. One non-blocking follow-up: `InventoryTopicsTest` only runs the reclass backend, so the omegaconf coercion case I flagged isn't guarded by CI. Worth parametrizing `consumed_topics` across all three backends like `InventoryTargetTestBase` does (`tests/test_inventory.py:56-92`). Not a blocker.

kapitan/topics.py
Moep90: Fair,  withdrawing the deepcopy suggestion.

Moep90 (approved):
Reset fixes on jinja2/jsonnet verified on `aeb12715`, thanks. Approving. Left one optional follow-up on backend test coverage for `consumed_topics`.

kapitan/inventory/inventory.py
ramaro: thanks, I added the fix

PR #1568, opened 2026-06-19
docs: auto-generate CLI and .kapitan flag reference from argparse

Moep90:
## What

Flag documentation that generates itself from `build_parser()` at docs-build time, so it can't drift from the actual CLI.

- **`docs/gen_flags.py`** — introspects the argparse parser, returns Markdown tables: global flags, a per-command flag table, and the per-section `.kapitan` dotfile tables.
- **`docs/hooks.py`** — an mkdocs `on_page_markdown` hook that replaces marker comments with the generated tables:
  - `<!-- kapitan-flags:command:compile -->` → that command's flags
  - `<!-- kapitan-flags:global -->` → global flags
  - `<!-- kapitan-flags:dotfile -->` → all `.kapitan` sections
- Per-command pages embed their own table; new pages for `eval`/`refs`/`init`; the `.kapitan` page is split into one table per section.
- **Makefile**: `docs_build` (strict, timeout-guarded), `docs_preview` (`mkdocs serve` for local edits), and `docs_serve` relabelled (mike serves *published* versions).
- **CI** (`documentation.yml`) runs `make docs_build` + `make docs_check`.
- **`tests/test_docs_flags.py`** guards the generator (empty tables, an undocumented new subcommand, broken columns) plus a timeout-guarded real build.

## Why

The hand-written `kapitan compile --help` dump in the docs had already gone stale — it listed `--inventory-backend {reclass}` while the parser offers `reclass-rs` and `omegaconf` too. Generating from the parser removes that whole class of drift.

A hook is used instead of `markdown-exec` on purpose: markdown-exec re-renders its output a second time, which made `pymdownx.inlinehilite` backtrack catastrophically on the many inline-code cells and **hang the build forever**. The hook injects Markdown that flows through the normal pipeline exactly once. The timeout-guarded build test exists to catch any such hang as a fast failure rather than a stuck CI job.

## Verification

- `make docs_build` — strict build green in ~1.4s.
- `make docs_check` — 68 passed, including the real-build smoke test.
- Confirmed live data renders (`reclass-rs`/`omegaconf`), long flags like `--inventory-pool-cache` no longer wrap mid-token (`docs/stylesheets/extra.css`), and the `.kapitan` page is findable via "dot-kapitan"/"dot kapitan" search.

## Out of scope

Pre-existing `versions.json` 404s (mike, local-serve only) and httplib2/pyparsing `DeprecationWarning`s (transitive deps) — both cosmetic, neither fails the build.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -46,10 +46,11 @@ jobs:
           uv run mike --version
           uv run mkdocs --version
 
-      - run: uv run mkdocs build
+      - name: Build docs (strict, timeout-guarded)
+        run: make docs_build
 
-      - name: Verify rendering
-        run: uv run pytest tests/test_docs_rendering.py -v --no-cov
+      - name: Verify rendering and flag-table generation
+        run: make docs_check
 
       # On master pushes (and manual dispatch): refresh the rolling "dev" docs.
       - name: Deploy dev docs
--- a/Makefile
+++ b/Makefile
@@ -175,21 +175,21 @@ clean:
 .PHONY: docs_build
 docs_build:
 	@echo "===== Building Documentation ====="
-	timeout 300 $(UV_RUN) mkdocs build --strict
+	timeout 300 uv run mkdocs build --strict
 
 # Preview docs with live reload of LOCAL changes (unlike docs_serve / mike serve,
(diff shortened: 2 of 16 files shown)


Moep90:
image

ademariag (approved):

PR #1571, opened 2026-06-19
test(vault): fix flaky test_vault_reveal permission-denied under parallel runs

Moep90:
## What

Fixes the intermittent `test_vault_reveal` failure that has been showing up across multiple PRs:

```
hvac.exceptions.Forbidden: 1 error occurred:
    * permission denied
, on post http://localhost:8200/v1/secret/data/foo
```

## Why

The KV (`VaultServer`) and transit (`VaultTransitServer`) test servers both write the **same** process-global `VAULT_TOKEN` (plus the userpass/approle vars) in `set_vault_attributes`, each scoped to its own policy. The transit policy only grants `transit/*`, not `secret/*`.

We run `pytest -n auto` with `pytest-random-order` installed, so a single xdist worker can set up both servers in random order. When the transit server initialises after the KV server, its transit-only token overwrites the shared `VAULT_TOKEN`. Any later KV test in that worker (`test_vault_reveal` writing to `secret/data/foo`) then gets `permission denied`. Because it depends on order, it surfaces on different PRs at random rather than reliably.

## Approach

- Store each server's issued credentials on the instance instead of only in the environment.
- Add `VaultServer.export_env()` that re-exports that server's credentials.
- Call `self.server.export_env()` from `setUp` in each vault test class (`VaultSecretTest`, `VaultTransitTest`, `CliFuncsTest`) so the active token always matches the server under test, regardless of what another worker did.

## Verification

Reproduced the failure by running the transit suite before the KV suite in a single process (the order random-order can pick), which previously failed with the `Forbidden` error. With this change that same order passes.


ramaro (approved):

ramaro:
@Moep90 thanks for the fix

PR #1574, opened 2026-06-19
fix(dev): uv Makefile commands should run through mise

jjmaestro:
Use the shared UV_RUN wrapper for the remaining Makefile commands that invoke uv.

These came from 69810d3, 93547f2 and 08be97a. One was missed when #1411 merged cleanly, but the other two are coming fro missed reviews.


ramaro (approved):

ramaro:
@jjmaestro thanks for the fix!

PR #1575, opened 2026-06-19
Add flake.nix and direnv dev to bootstrap the Kapitan dev environment

jjmaestro:
> [!NOTE]
> Stacked on top of #1574 

Provide a small nix develop environment, plus direnv, with make and mise so contributors can enter the project toolchain without installing those bootstrap tools globally.

Also, install Go through the project toolchain so uv can build Python packages that require a Go compiler.


jjmaestro:
@ademariag I've added a flake.nix that just installs the bare minimum to then use the `make` targets with mise, etc.

Moep90:
**Correction / withdrawn:** I was wrong below. `gojsonnet` (`pyproject.toml`) builds go-jsonnet from source via cgo, and `make install` pulls it in through `uv sync --locked --all-extras --dev`, so the `go` pin is a real dev requirement, not dead weight. Thanks @jjmaestro. LGTM once the title/description mention the go pin and the `uv run` to `$(UV_RUN)` cleanup.

<details><summary>Original comment (incorrect)</summary>

> **`go` toolchain added but nothing uses it**
>
> `mise.toml`, `mise.lock`, and the `go version` check in `install_tools` now pin `go = "1.26.4"`, but I can't find anything that actually uses it:
> - No `.go` files, and no `go build` / `go run` / `go install` anywhere
> - No Makefile target touches go apart from the new `go version` check
> - cue, kustomize, and helm all install as prebuilt binaries through the aqua/ubi backends, so none of them need a local go toolchain
>
> So every contributor running `make install_tools` now pulls a full go toolchain (150MB or so per platform) that nothing currently calls.

</details>

jjmaestro:
image

It's right there in the commit description that adds the dependency. If you don't have Go, you won't build `gojsonnet` and the tests fail.

I guess you & other devs, and myself included when I added mise, leak dependencies into the setup because it's not hermetic. I just found this because I've reinstalled my laptop and was re-running things from scratch. 

> Could you drop the go pin, or note in the description why it's needed (some build step coming in a later PR)? It also feels like scope creep against the title. This PR is "add flake.nix and direnv", but the go pin and the uv run to $(UV_RUN) Makefile cleanup are separate changes the description doesn't mention.

Ah, sorry, I thought it would be clear that, a flake.nix aiming to setup the dev environment encompasses actually fixing the dev environment by adding the missing dependency. I'll address this changing the content of the description and title!

jjmaestro:
> LGTM once the title/description mention the go pin and the uv run to $(UV_RUN) cleanup.

I've added `go` to the description, but the `uv` run is coming from the PR being stacked on top of #1574.

ramaro (approved):

Issue #1370, opened 2025-12-30
[bug]:  Numeric strings with leading zeros converted to scientific notation in Helm output


sergeykad:
### Description of the bug

When a parameter value looks like a number with a leading zero (e.g., "03190301"), Kapitan outputs it in scientific notation (3.190301e+06) in the compiled Helm manifests.

The value passes through three YAML parsers:
1. reclass-rs reads the inventory and outputs the value unquoted
2. Python YAML writes the temp helm_values file, keeping it unquoted
3. Go YAML (Helm) interprets the unquoted 03190301 as integer 3190301

Helm then renders large integers in scientific notation, losing the leading zero.

### Steps to reproduce

1. Add a parameter with a leading-zero numeric string to inventory:
  parameters:
    my_value: "03190301"
2. Reference it in Helm values:
  helm_values:
    MY_ENV_VAR: ${my_value}
3. Compile the target:
  kapitan compile -t <target>
4. Check the output manifest - the value appears as 3.190301e+06 instead of 03190301

### What happened?

Expected: The compiled manifest contains value: "03190301" (string preserved exactly as defined).

Actual: The compiled manifest contains value: "3.190301e+06" (converted to scientific notation, leading zero lost).

### Additional context

Environment:
- Kapitan: latest (Docker image kapicorp/kapitan:latest)
- Inventory backend: reclass-rs
- Input type: Helm

### Version (kapitan)

v0.31.0 (stable)

### Version (Python)

3.11

### Python virtual environment

- [ ] pyenv
- [ ] venv
- [ ] no environment

### OS information

Ubuntu 22.04

### Relevant log output

```bash

```

PR #1596, opened 2026-07-09
fix(yaml): quote leading-zero numeric strings in compiled output (#1595)

### LINKED ISSUE #1595: [bug]: Numeric-string quoting (#1372 fix) not applied to compiled manifest output — breaks Deployment `env[].value`
### Description of the bug

Follow-up to #1370 / #1372. That fix stopped leading-zero numeric strings like `"03190301"` from becoming scientific notation, by quoting them in the helm values temp file (`write_helm_values_file()`, via the new `_helm_str_representer`). That path is now correct.

The same root cause survives on a second path the fix never touched: the compiled manifest output. `_helm_str_representer` guards only the helm values file (Helm's input); the serializer that writes the final `compiled/<target>/.../*.yaml` manifests does not, so it re-emits the value unquoted:

```
inventory:  "03190301"  (string)
  → write_helm_values_file()          # #1372 quotes it → '03190301'   ✓
  → helm template                     # emits  value: "03190301"       ✓
  → Kapitan parses helm stdout        # Python str "03190301"          ✓
  → Kapitan writes compiled manifest  # plain dump, NO representer      ✗
                                      # → value: 03190301  (UNQUOTED)
```

In a ConfigMap `data` field this is harmless: the schema coerces any scalar to a string, so unquoted `03190301` still loads as `"03190301"`. #1372's tests assert against exactly this case (`test_numeric_string_values_preserved`), so they pass while the bug survives.

In a Deployment `env[].value` (typed `string`, no coercion) the unquoted value is a YAML/JSON integer, and the API server rejects the manifest:

```
Error from server (BadRequest): error when creating "deploy.yaml":
Deployment in version "v1" cannot be handled as a Deployment:
json: cannot unmarshal number into Go struct field EnvVar.spec.template.spec.containers.env.value of type string
```

A target that routes such a value into a container env var compiles cleanly but produces a manifest that `kubectl apply` and Argo CD sync refuse.

### Steps to reproduce

1. Inventory parameter, correctly quoted per #1370:
   ```yaml
   parameters:
     sender_id: "03190301"
   ```
2. A chart that renders it into a Deployment env var — `value: "{{ .Values.sender_id }}"` — with `helm_values: { sender_id: ${sender_id} }`.
3. `kapitan compile -t <target>`
4. Inspect the compiled Deployment: the value is unquoted — `value: 03190301`, not `value: "03190301"`.
5. `kubectl apply --dry-run=server -f <compiled-deployment>.yaml` fails. A fresh create reports `cannot unmarshal number into Go struct field EnvVar.spec.template.spec.containers.env.value of type string`; patching an existing object reports `cannot convert int64 to string`. Both come from parsing the unquoted value as an integer, which also drops the leading zero: `03190301` → `3190301`.

### What happened?

Expected: `value: "03190301"` — the string preserved on the output path, the same guarantee #1372 gives the input path.

Actual: `value: 03190301` — unquoted. It loads only where the schema coerces scalars (ConfigMap data); the API server rejects it wherever the field is typed `string` (EnvVar.value and any other `string`-typed field).

### Suggested fix

Apply the `_helm_str_representer` rule, or an equivalent, to the manifest-output serializer, not only to `write_helm_values_file()`. #1372's representer is the right pattern; it just needs to guard the output emitter too, so numeric-looking strings stay quoted end to end.

A test beyond #1372's ConfigMap case would also help: assert the value survives in a `string`-typed field such as Deployment `env[].value`, where scalar coercion no longer hides the defect.

### Additional context

Root cause and the input-side fix are documented in #1370 and #1372.

Reproduced on 0.36.2 (Docker `kapicorp/kapitan:0.36.2`), reclass-rs backend, Helm input. The value is no longer scientific notation (`3.190301e+06`) — #1372 fixed that — but the unquoted `03190301` now fails apply on `string`-typed fields.

### Version (kapitan)

0.36.2

### Inventory backend

reclass-rs

### Input type

helm

sergeykad:
## Summary

Quote leading-zero numeric strings (e.g. `"03190301"`) in the compiled
manifest serializer so they survive as strings end to end.

Fixes #1595

## Motivation

PyYAML's YAML 1.1 resolver treats `"03190301"` as a string and emits it
unquoted, but the Go/YAML 1.2 parsers in Helm and the Kubernetes API server
read it as an integer and drop the leading zero. #1370/#1372 fixed this only
for helm's *input* values file; the serializer that writes the final
`compiled/.../*.yaml` manifests still re-emitted the value unquoted. It loads
where the schema coerces scalars (ConfigMap `data`) but is rejected on
`string`-typed fields like Deployment `env[].value` (`cannot unmarshal number
into ... of type string`).

## Affected files or URLs

- `kapitan/utils.py` — `is_leading_zero_int_string()` predicate; applied in `PrettyDumper` (PyYAML output path)
- `kapitan/yaml_ryml.py` — same guard on the rapidyaml output path (value + key)
- `kapitan/inputs/helm.py` — `_helm_str_representer` reuses the shared predicate
- `tests/` — predicate unit tests, ryml raw-text test, helm end-to-end Deployment `env` regression test + fixture

## Test plan

- [x] Tests added or updated
- [x] Documentation updated (if user-facing behavior changed) — N/A, correctness fix, no new flag/config
- [x] `make lint` passes
- [~] Full test suite passes (`uv run pytest tests/ --no-cov`) — all serializer/helm/compile/util tests pass (678 passed); the only local failures are Vault/cue/kustomize tests that need external tooling not installed in my sandbox, unrelated to this change. CI has that tooling.

## Risk assessment

Low. The output serializer is shared across all input types, but the change
only adds quoting for leading-zero digit strings — a class PyYAML previously
emitted plain. Other numeric strings are unaffected (already quoted by PyYAML's
emitter / rapidyaml's resolver mirror), so no existing golden output changes;
all `test_compile.py` goldens (reclass, reclass-rs, omegaconf) pass unchanged.
Quoting is lossless. Revert is a single-commit revert.

## Follow-up work intentionally left out

None.

### DIFF
--- a/kapitan/inputs/helm.py
+++ b/kapitan/inputs/helm.py
@@ -20,7 +20,7 @@ from kapitan.inputs.base import InputType
 from kapitan.inputs.cache import InputCache, walk_and_hash
 from kapitan.inputs.kadet import BaseModel, BaseObj
 from kapitan.inventory.model.input_types import KapitanInputTypeHelmConfig
-from kapitan.utils import is_leading_zero_int_string
+from kapitan.utils import leading_zero_str_representer
 
 
 logger = logging.getLogger(__name__)
@@ -214,17 +214,6 @@ def render_chart(
         return ("", error_message)
 
 
-def _helm_str_representer(dumper, data):
-    """Quote leading-zero digit strings so Helm's Go YAML parser keeps them as
-    strings. ``is_leading_zero_int_string`` guards the compiled-manifest
-    serializer too, so the value stays quoted end to end.
-
-    See https://github.com/kapicorp/kapitan/issues/1370 and #1595.
-    """
-    style = "'" if is_leading_zero_int_string(data) else None
-    return dumper.represent_scalar("tag:yaml.org,2002:str", data, style=style)
-
-
 def write_helm_values_file(helm_values: dict):
     """Dump helm values into a temporary YAML file.
 
@@ -237,7 +226,9 @@ def write_helm_values_file(helm_values: dict):
     _, helm_values_file = tempfile.mkstemp(".helm_values.yml", text=True)
     with open(helm_values_file, "w") as fp:
         dumper = yaml.SafeDumper
-        dumper.add_representer(str, _helm_str_representer)
+        # Quote leading-zero digit strings so Helm's Go YAML parser keeps them
+        # as strings (#1370/#1595); shared with the compiled-output serializer.
+        dumper.add_representer(str, leading_zero_str_representer)
         yaml.dump(helm_values, fp, Dumper=dumper)
 
     return helm_values_file
--- a/kapitan/inputs/helm.py
+++ b/kapitan/inputs/helm.py
@@ -20,6 +20,7 @@ from kapitan.inputs.base import InputType
 from kapitan.inputs.cache import InputCache, walk_and_hash
 from kapitan.inputs.kadet import BaseModel, BaseObj
 from kapitan.inventory.model.input_types import KapitanInputTypeHelmConfig
+from kapitan.utils import is_leading_zero_int_string
 
 
 logger = logging.getLogger(__name__)
@@ -214,15 +215,13 @@ def render_chart(
 
 
 def _helm_str_representer(dumper, data):
-    """Quote strings that could be misinterpreted by Helm's Go YAML parser.
+    """Quote leading-zero digit strings so Helm's Go YAML parser keeps them as
+    strings. ``is_leading_zero_int_string`` guards the compiled-manifest
+    serializer too, so the value stays quoted end to end.
 
-    Fixes https://github.com/kapicorp/kapitan/issues/1370 where strings like
-    "03190301" are converted to scientific notation (3.190301e+06).
+    See https://github.com/kapicorp/kapitan/issues/1370 and #1595.
     """
-    style = None
-    # Quote digit-only strings starting with 0 or large numeric strings
-    if data and len(data) > 1 and data.isdigit() and (data[0] == "0" or len(data) > 6):
-        style = "'"
+    style = "'" if is_leading_zero_int_string(data) else None
     return dumper.represent_scalar("tag:yaml.org,2002:str", data, style=style)
 
 
--- a/kapitan/utils.py
+++ b/kapitan/utils.py
@@ -181,6 +181,19 @@ def prune_empty(d):
 _SUPPORTED_MULTILINE_STYLES = {"literal": "|", "folded": ">", "double-quotes": '"'}
 
 
+def is_leading_zero_int_string(data: str) -> bool:
+    """True if ``data`` is a leading-zero digit string like ``"03190301"``.
+
+    PyYAML emits these unquoted (YAML 1.1 rejects them as octal), but Helm and
+    Kubernetes (Go/YAML 1.2) read them as ints and drop the leading zero, so
+    callers force-quote them. Other int-looking strings are already quoted by
+    PyYAML's emitter and by ``_str_is_ambiguous``.
+
+    See https://github.com/kapicorp/kapitan/issues/1595 (follows #1370/#1372).
+    """
+    return len(data) > 1 and data[0] == "0" and data.isdigit()
+
+
 class PrettyDumper(yaml.SafeDumper):
     """
     Increases indent of nested lists.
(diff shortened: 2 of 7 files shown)


kapitan/utils.py
```diff
@@ -210,18 +223,14 @@ def get_dumper_for_style(cls, style_selection="double-quotes"):
...
-            def _str_presenter(dumper, data, _style=style_char):
-                if "\n" in data:
-                    return dumper.represent_scalar(
-                        "tag:yaml.org,2002:str", data, style=_style
-                    )
-                return dumper.represent_scalar("tag:yaml.org,2002:str", data)
+        def _str_presenter(dumper, data, _style=style_char):
+            if _style is not None and "\n" in data:
+                style = _style
+            elif is_leading_zero_int_string(data):
```
Moep90: While testing this I found one path where the fix doesn't hold yet: the presenter only exists on the subclasses that `get_dumper_for_style()` builds, but `Revealer._reveal_file()` in `kapitan/refs/base.py` dumps with the bare `PrettyDumper` class, which has no str representer at all. Revealing a compiled manifest strips the quoting again:

```python
>>> from kapitan.utils import PrettyDumper
>>> yaml.dump(yaml.safe_load("value: '03190301'"), Dumper=PrettyDumper)
'value: 03190301\n'
```

So `kapitan compile` followed by `kapitan refs reveal -f compiled/.../deployment.yaml | kubectl apply -f -` still hits the #1595 error, and reveal is the usual route to the cluster for manifests that contain refs. `kapitan inventory` (`resources.py`) uses the bare class too, though that one only prints to stdout.

Two easy options: route those call sites through `get_dumper_for_style()`, or add the representer to the base `PrettyDumper` so every subclass inherits it. Keeping this PR as is and tracking it in a follow-up issue works too, whichever you prefer.

kapitan/inputs/helm.py
```diff
@@ -214,15 +215,13 @@ def render_chart(
...
 
-    Fixes https://github.com/kapicorp/kapitan/issues/1370 where strings like
-    "03190301" are converted to scientific notation (3.190301e+06).
+    See https://github.com/kapicorp/kapitan/issues/1370 and #1595.
     """
-    style = None
-    # Quote digit-only strings starting with 0 or large numeric strings
-    if data and len(data) > 1 and data.isdigit() and (data[0] == "0" or len(data) > 6):
-        style = "'"
+    style = "'" if is_leading_zero_int_string(data) else None
```
Moep90: Dropping the `len(data) > 6` branch is safe as far as I can tell: PyYAML's emitter already quotes int-resolvable strings like `"1234567"` on its own (checked locally, it comes out as `'1234567'`). Could you add one value without a leading zero, say `"1234567"`, to `test_write_helm_values_file_preserves_numeric_strings`? All its current values start with `0`, so nothing pins the case the removed branch used to cover if the dumper setup ever changes.

tests/test_helm_input.py
```diff
@@ -318,6 +318,41 @@ def test_numeric_string_values_preserved(self):
...
+    def test_numeric_string_quoted_in_compiled_output(self):
+        """Numeric strings must stay quoted in the compiled manifest (#1595).
+
+        The #1372 fix quoted only helm's values file, not the output serializer,
+        so it re-emitted "03190301" unquoted. A Deployment ``env[].value`` is
+        string-typed with no coercion, so an unquoted value fails ``kubectl
+        apply``. Assert on raw text: ``yaml.safe_load`` reads "03190301" as a
+        string whether or not it is quoted, so it cannot detect the defect.
+        """
+        temp = tempfile.mkdtemp()
```
Moep90: Small one: the repo's testing conventions ask for `tempfile.TemporaryDirectory()` as a context manager instead of bare `mkdtemp()`, so the directory gets cleaned up after the test. The older tests in this file predate that rule, so matching them is understandable, but since this test is new it's a cheap switch.

Moep90 (commented):
Thanks for the thorough fix. The test hygiene here is well above average: raw-text assertions where `safe_load` can't see the defect, and the `"08"`/`"09"` ryml test that isolates the new predicate from `_str_is_ambiguous` is a nice touch.

I ran everything locally on the branch: the new predicate/helm/ryml tests pass (including `test_yaml_use_rapidyaml.py` with the `rapidyaml` extra installed, which a plain `uv sync` skips), and the `tests/test_compile.py` goldens pass unchanged on all three inventory backends with `--all-extras`.

One real gap around `refs reveal` re-emitting the value unquoted, details inline.

sergeykad:
All three addressed in `434136dc`.

While fixing the reveal/inventory gap I consolidated the quoting rule — there were three copies of the leading-zero representer (per-style closure, helm, and the new base one). They're now one shared `leading_zero_str_representer` in `utils.py` that the base `PrettyDumper`, its `get_dumper_for_style` subclasses, and helm all use; the per-style presenter keeps its multiline case and delegates the rest. The ryml emitter stays separate (different library).

`refs reveal` now emits `'03190301'` quoted; goldens and helm/utils/ryml tests pass.

kapitan/utils.py
sergeykad: Fixed in `434136dc`. Registered a shared `leading_zero_str_representer` on the base `PrettyDumper` so `_reveal_file` and inventory inherit it, and routed the per-style closure and helm through it instead of adding a third copy. Verified `refs reveal` keeps `'03190301'` quoted.

kapitan/inputs/helm.py
sergeykad: Done in `434136dc` — added `"1234567"` to the test (as a `subTest` over both values).

tests/test_helm_input.py
sergeykad: Done in `434136dc` — switched to `with tempfile.TemporaryDirectory()`.

